PR #84, opened 2019-04-11
Addresses #82

### LINKED ISSUE #82: Generalized HVAC models
https://trello.com/c/DmfG2UXi/87-generalized-hvac-models

TODO:
- [x] Single-speed
- [x] Two-speed
- [x] Variable-speed
- [x] Remove all debug statements
- [x] Test against Jon's values
- [x] Jon to spot check var speed cooling mode curves

joseph-robertson:
Closes #82


tests/hpxml_translator_test.rb
```diff
@@ -105,6 +105,84 @@ def test_invalid
...
+      assert_in_epsilon(expected_cop, actual_cop, 0.01)
+    end
+
+    # two-speed air conditioner
+    cOOL_EIR_FT_SPEC = [[-3.877526888, 0.164566276, -0.001272755, -0.019956043, 0.000256512, -0.000133539],
+                        [-1.990708931, 0.093969249, -0.00073335, -0.009062553, 0.000165099, -0.0000997]]
+    cOOL_CAP_FT_SPEC = [[3.940185508, -0.104723455, 0.001019298, 0.006471171, -0.00000953, -0.000161658],
+                        [3.109456535, -0.085520461, 0.000863238, 0.00863049, -0.0000210, -0.000140186]]
+    capacity_ratios = [0.72, 1.0]
+    fan_speed_ratios = [0.86, 1.0]
```
shorowit: Can you wrap all of these assumed values into hvac.rb methods that are referenced here and in hvac.rb? That way we only have the values defined once and they can't become out of sync.

joseph-robertson:
@shorowit, I suppose you can start reviewing this while we wait for Jon to check variable speed curve coefficients.

shorowit:
I'm looking over the simulation results compared to master branch and a couple things pop out:
- Variable-speed ACs, ASHPs, and MSHPs are getting different cooling capacities (heating capacities and other system types are unchanged). Is that expected?
- MSHPs are showing the largest HVAC energy changes. I didn't think they were modified as part of this work though? Maybe the SI->IP stuff affected them?

shorowit:
Thanks! The MSHP CI results now perfectly match `master`.

measure.rb
```diff
@@ -2258,10 +2258,12 @@ def self.add_cooling_system(runner, model, building)
 
         if num_speeds == "1-Speed"
 
-          eers = [0.82 * seer + 0.64]
           shrs = [0.73]
+          capacity_ratios = [1.0]
+          fan_speed_ratios = [1.0]
```
shorowit: I'd rather go the opposite approach. For all ACs/ASHPs, can you make these assignments inside the apply_foo methods and remove as arguments?

resources/constants.rb
```diff
@@ -5,6 +5,22 @@ def self.AssumedInsideTemp
...
 
+  def self.C_d_clg_low_seer
+    return 0.20
+  end
+
+  def self.C_d_clg_high_seer
+    return 0.07
+  end
+
+  def self.C_d(var_speed = false)
```
shorowit: Can you merge all three methods into a single C_d with the appropriate arguments?

resources/hvac.rb
```diff
@@ -6,7 +6,64 @@
...
+      return [[0.718605468, 0.410099989, -0.128705457]]
+    elsif num_speeds == 2
+      return [[0.65673024, 0.516470835, -0.172887149],
+              [0.690334551, 0.464383753, -0.154507638]]
+    elsif num_speeds == 4
+      return [[1, 0, 0]] * 4
+    end
+  end
+
+  def self.cOOL_EIR_FFLOW_SPEC_AC(num_speeds = 1)
```
shorowit: Please move all methods that aren't apply_foo below the apply_foo methods.

resources/hvac.rb
```diff
@@ -689,7 +778,7 @@ def self.apply_central_ashp_2speed(model, runner, seer, hspf, eers, cops, shrs,
       htg_coil.setMaximumOutdoorDryBulbTemperatureforCrankcaseHeaterOperation(UnitConversions.convert(crankcase_temp, "F", "C"))
       htg_coil.setDefrostEnergyInputRatioFunctionofTemperatureCurve(defrost_eir_curve)
       htg_coil.setMaximumOutdoorDryBulbTemperatureforDefrostOperation(UnitConversions.convert(40.0, "F", "C"))
-      htg_coil.setDefrostStrategy("ReverseCryle")
+      htg_coil.setDefrostStrategy("ReverseCycle")
```
shorowit: Nice catch.

resources/hvac.rb
```diff
@@ -2849,12 +2928,816 @@ def self.get_gshp_PumpKW_Adjust(gpm_btuh)
...
+    eir_35_H = eir_47_H * calc_biquad(coeff_eir[1], 70.0, 35.0)
+    eir_17_H = eir_47_H * calc_biquad(coeff_eir[1], 70.0, 17.0)
+
+    eir_47_L = calc_EIR_from_COP(cops[0], fan_power_rated)
+    eir_62_L = eir_47_L * calc_biquad(coeff_eir[0], 70.0, 62.0)
+    eir_35_L = eir_47_L * calc_biquad(coeff_eir[0], 70.0, 35.0)
+    eir_17_L = eir_47_L * calc_biquad(coeff_eir[0], 70.0, 17.0)
+
+    q_H47 = 1.0
+    q_H35 = 0.7519 # Hard code Q_35 from BEopt1 #TODO: UPDATE!!
```
shorowit: What is this comment about?

resources/hvac.rb
```diff
@@ -2849,12 +2928,816 @@ def self.get_gshp_PumpKW_Adjust(gpm_btuh)
...
+    q_H17 = q_H47 * calc_biquad(coeff_q[1], 70.0, 17.0)
+
+    q_L47 = q_H47 * capacity_ratios[0]
+    q_L62 = q_L47 * calc_biquad(coeff_q[0], 70.0, 62.0)
+    q_L35 = 0.7519 * capacity_ratios[0] # Hard code Q_35 from BEopt1  # TODO: UPDATE!!!
```
shorowit: And this one?

resources/hvac.rb
```diff
@@ -1069,11 +1138,19 @@ def self.apply_mshp(model, runner, seer, hspf, shr,
     d = oat_slope
     e = 0
     f = 0
+
+    a = a - 160.0 / 9.0 * (b + d) + 25600.0 / 81.0 * (c + e + f)
```
shorowit: Why did you copy this code instead of using the convert_curve_biquadratic method?

tests/hpxml_translator_test.rb
shorowit: Not sure why you resolved this comment, it still hasn't been addressed.

resources/hvac.rb
joseph-robertson: Not sure. They are in Jon's notebook.

shorowit: Can you ask Jon?

resources/hvac.rb
joseph-robertson: Good question.

tests/hpxml_translator_test.rb
joseph-robertson: Not sure what is left to be addressed...? You're talking about wrapping `capacity_ratios` and `fan_speed_ratios`, and not just the curves?

shorowit: My thinking was that we shouldn't be setting these values both here and somewhere else, since they can become out of sync. But now that I think about it, I guess I don't know whether that's a problem. For example, if we decided to change the capacity ratios we use in the model, maybe the tests could/would stay unchanged? If that's the case, then my comment doesn't apply.

Issue #92, opened 2019-05-08
Add neighbors for shading


shorowit:
Add an HPXML test file (via the Rakefile) with neighbors specified. HPXML doesn't really have a good place for it, so I suggest using an extension in the `Site` element like so: 

```
        <Site>
          <extension>
            <Neighbors>
              <NeighborBuilding>
                <Azimuth>45</Azimuth> # neighbor is to the northeast
                <Distance>10</Distance> # ft away
              </NeighborBuilding>
              <NeighborBuilding>
                <Azimuth>225</Azimuth> # neighbor is to the southwest
                <Distance>15</Distance> # ft away
            </NeighborBuilding>
          </Neighbors>
        </Site>
```

Note that we don't know the front of the home, so neighbors are defined in HPXML based on azimuth instead of FBLR. 

Because we have exploded surfaces rather than full 3d geometry in our model, instead of copying all the surfaces from the main building, let's simply add 1 shading surface for each azimuth at the specified distance from that azimuth's surfaces. Each shading surface will be the height of the building, and the width can be set to [max_azimuth_length](https://github.com/NREL/OpenStudio-HPXML/blob/master/measure.rb#L522).

So a building that looks like this without neighbors:
![image](https://user-images.githubusercontent.com/5861765/57412728-f890d400-71ae-11e9-9f35-6a95a6f82729.png)

might look like this with neighbors:
![image](https://user-images.githubusercontent.com/5861765/57412767-28d87280-71af-11e9-995b-a21c5412c9e1.png)

We should also add tests that verify the number of shading surfaces (and their azimuth, if possible).

Needed for https://github.com/NREL/OpenStudio-HEScore/issues/52

joseph-robertson:
What if there exists a neighbor whose azimuth doesn't match any other surface or subsurface azimuth? What should the area of the neighbor shading surface be in this case?

@shorowit

shorowit:
The shading surface in any direction has identical area, since it's just dependent on the building height and max_azimuth_length. Or am I missing a nuance here?

joseph-robertson:
In your second screenshot, which set of surfaces gets the neighbor shading surface, e.g., at 10 ft and at 15 ft?

PR #93, opened 2019-05-09
Addresses #92

joseph-robertson:


Rakefile
```diff
@@ -569,6 +576,18 @@ def get_hpxml_file_site_values(hpxml_file, site_values)
...
 end
 
+def get_hpxml_file_site_neighbor_values(hpxml_file, site_neighbors_values)
+  if ['base-site-neighbors.xml'].include? hpxml_file
+    site_neighbors_values << { :azimuth => 0,
+                               :distance => 10 }
+    site_neighbors_values << { :azimuth => 180,
+                               :distance => 15 }
+  elsif ['invalid_files/bad-site-neighbor-azimuth.xml'].include? hpxml_file
+    site_neighbors_values[0][:azimuth] = 145
```
shorowit: 👍

measure.rb
```diff
@@ -521,12 +521,37 @@ def self.explode_surfaces(runner, model)
...
+    model.getShadingSurfaceGroups.each do |shading_surface_group|
+      next if shading_surface_group.name.to_s != Constants.ObjectNameNeighbors
+
+      shading_surface_group.shadingSurfaces.each do |shading_surface|
+        azimuth = shading_surface.additionalProperties.getFeatureAsInteger("Azimuth").get
+        azimuth_rad = UnitConversions.convert(azimuth, "deg", "rad")
+        distance = shading_surface.additionalProperties.getFeatureAsDouble("Distance").get
+
+        unless azimuth_lengths.keys.include? azimuth
+          runner.registerError("A neighbor building surface has an azimuth '#{azimuth}' not equal to the amizuth of any wall (#{azimuth_lengths.keys * ", "}).")
```
shorowit: Let's say "neighbor building has an azimuth" instead of "neighbor building surface has an azimuth".

measure.rb
```diff
@@ -1249,6 +1274,38 @@ def self.add_walls(runner, model, building, spaces, subsurface_areas)
...
+    building_construction_values = HPXML.get_building_construction_values(building_construction: building.elements["BuildingDetails/BuildingSummary/BuildingConstruction"])
+
+    shading_surfaces = []
+    building.elements.each("BuildingDetails/BuildingSummary/Site/extension/Neighbors/NeighborBuilding") do |neighbor_building|
+      neighbor_building_values = HPXML.get_neighbor_building_values(neighbor_building: neighbor_building)
+      azimuth = neighbor_building_values[:azimuth]
+      distance = neighbor_building_values[:distance]
+
+      wall_height = 8.0 * building_construction_values[:number_of_conditioned_floors_above_grade]
+      z_origin = foundation_top
```
shorowit: I suggest setting z_origin to be 0 (so that the shading surface starts at grade even for pier & beam foundations). And wall_height is currently ignoring attic surfaces -- can we use the max y-value of any model surface instead?

With these changes, you should be able to remove lines 1277-1278.

resources/hpxml.rb
```diff
@@ -90,6 +90,29 @@ def self.get_site_values(site:)
...
   end
 
+  def self.add_site_neighbor(hpxml:,
+                             azimuth:,
+                             distance:,
+                             **remainder)
+    neighbors = XMLHelper.create_elements_as_needed(hpxml, ["Building", "BuildingDetails", "BuildingSummary", "Site", "extension", "Neighbors"])
+    neighbor_building = XMLHelper.add_element(neighbors, "NeighborBuilding")
+    XMLHelper.add_element(neighbor_building, "Azimuth", to_integer(azimuth)) unless azimuth.nil?
+    XMLHelper.add_element(neighbor_building, "Distance", to_float(distance)) unless distance.nil?
```
shorowit: Remove the unless foo.nil? on these two lines. Both values are required, so if nil is provided for one of them, we'd like it cause an error here (instead of creating an HPXML file that is missing a required element).

resources/hpxml.rb
```diff
@@ -90,6 +90,29 @@ def self.get_site_values(site:)
              :disable_natural_ventilation => to_bool(XMLHelper.get_value(site, "extension/DisableNaturalVentilation")) }
   end
 
+  def self.add_site_neighbor(hpxml:,
+                             azimuth:,
+                             distance:,
+                             **remainder)
+    neighbors = XMLHelper.create_elements_as_needed(hpxml, ["Building", "BuildingDetails", "BuildingSummary", "Site", "extension", "Neighbors"])
+    neighbor_building = XMLHelper.add_element(neighbors, "NeighborBuilding")
+    XMLHelper.add_element(neighbor_building, "Azimuth", to_integer(azimuth)) unless azimuth.nil?
```
shorowit: Yes, azimuths are integer. 👍

Issue #94, opened 2019-05-11
Add Schedule Type Limits


shorowit:
Add Schedule Type Limits to Schedule class (and any other schedules we generate), so as to prevent a large number of warnings in the EnergyPlus err file. 

Example err file: 
[eplusout.err.txt](https://github.com/NREL/OpenStudio-HPXML/files/3169201/eplusout.err.txt)

PR #95, opened 2019-05-13
Addresses #94

joseph-robertson:
#94


resources/hotwater_appliances.rb
```diff
@@ -179,7 +179,7 @@ def self.apply(model, runner, weather, living_space,
       dist_pump_obj_name = Constants.ObjectNameHotWaterRecircPump
       dist_pump_weekday_sch = "0.010, 0.006, 0.004, 0.002, 0.004, 0.006, 0.016, 0.032, 0.048, 0.068, 0.078, 0.081, 0.074, 0.067, 0.057, 0.061, 0.055, 0.054, 0.051, 0.051, 0.052, 0.054, 0.044, 0.024"
       dist_pump_monthly_sch = "1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0"
-      dist_pump_schedule = MonthWeekdayWeekendSchedule.new(model, runner, dist_pump_obj_name, dist_pump_weekday_sch, dist_pump_weekday_sch, dist_pump_monthly_sch, 1.0, 1.0)
+      dist_pump_schedule = MonthWeekdayWeekendSchedule.new(model, runner, dist_pump_obj_name, dist_pump_weekday_sch, dist_pump_weekday_sch, dist_pump_monthly_sch, 1.0, 1.0, true, true, 0, 1, "Continuous")
```
shorowit: There are other schedules in this file, why are they not updated?

resources/schedules.rb
```diff
@@ -363,6 +383,16 @@ def createSchedule()
     schedule = OpenStudio::Model::ScheduleRuleset.new(@model)
     schedule.setName(@sch_name)
 
+    if not @lower_limit_value.nil? or not @upper_limit_value.nil? or not @numeric_type.nil? or not @unit_type.nil?
```
shorowit: Make this code a method so you don't have identical code two places....

shorowit: Was the HotWaterSchedule class intentionally not touched?

resources/hotwater_appliances.rb
joseph-robertson: I didn't see any warnings about them in the err file.

resources/schedules.rb
joseph-robertson: Oversight. I will address.

joseph-robertson:
I think this should be ready for another review.

resources/airflow.rb
```diff
@@ -969,6 +969,13 @@ def self.create_nat_vent_objects(model, runner, obj_name_natvent, building, nat_
     avail_sch = OpenStudio::Model::ScheduleRuleset.new(model)
     avail_sch.setName(obj_name_natvent + " avail schedule")
 
+    schedule_type_limits = OpenStudio::Model::ScheduleTypeLimits.new(model)
+    schedule_type_limits.setName(avail_sch.name.to_s)
+    schedule_type_limits.setLowerLimitValue(0)
+    schedule_type_limits.setUpperLimitValue(1)
+    schedule_type_limits.setNumericType("Discrete")
+    avail_sch.setScheduleTypeLimits(schedule_type_limits)
```
shorowit: Can you use the Schedule.createScheduleTypeLimits method here?

resources/hotwater_appliances.rb
```diff
@@ -28,7 +28,7 @@ def self.apply(model, runner, weather, living_space,
 
     # Schedules init
     timestep_minutes = (60.0 / model.getTimestep.numberOfTimestepsPerHour).to_i
-    start_date = model.getYearDescription.makeDate(1, 1)
+    start_date = OpenStudio::Date.new(OpenStudio::MonthOfYear.new(1), 1, model.getYearDescription.assumedYear)
```
shorowit: Does this get rid of the OS warning? Yay!

resources/schedules.rb
```diff
@@ -25,6 +26,7 @@ def initialize(model, runner, sch_name, weekday_month_by_hour_values, weekend_mo
     end
     if create_sch_object
       @schedule = createSchedule()
+      Schedule.createScheduleTypeLimits(@model, @schedule, lower_limit_value, upper_limit_value, numeric_type, unit_type)
```
shorowit: Much better. 👍

resources/schedules.rb
```diff
@@ -133,13 +135,8 @@ def createSchedule()
...
         end
-        wkdy_rule.setApplySunday(true)
-        wkdy_rule.setApplyMonday(true)
-        wkdy_rule.setApplyTuesday(true)
-        wkdy_rule.setApplyWednesday(true)
-        wkdy_rule.setApplyThursday(true)
-        wkdy_rule.setApplyFriday(true)
-        wkdy_rule.setApplySaturday(true)
+        Schedule.set_weekday_rule(wkdy_rule)
+        Schedule.set_weekend_rule(wkdy_rule)
```
shorowit: Good change.

tests/hpxml_translator_test.rb
```diff
@@ -298,6 +298,14 @@ def _test_simulation(args, this_dir, rundir, expect_error, expect_error_msgs)
   end
 
   def _verify_simulation_outputs(rundir, hpxml_path, results)
+    # Check that eplusout.err has no lines that include "Blank Schedule Type Limits Name input"
+    File.readlines(File.join(rundir, "eplusout.err")).each do |err_line|
+      next if err_line.include? 'Schedule:Constant="ALWAYS ON CONTINUOUS", Blank Schedule Type Limits Name input' # FIXME
+      next if err_line.include? 'Schedule:Constant="ALWAYS OFF DISCRETE", Blank Schedule Type Limits Name input' # FIXME
```
shorowit: What's the deal with these two? They are provided by OS and don't have schedule type limits provided?

resources/hotwater_appliances.rb
joseph-robertson: Yes.

resources/airflow.rb
joseph-robertson: Yes.

tests/hpxml_translator_test.rb
joseph-robertson: These two schedules exist in the idf file (without schedule type limits assigned) but do not exist in the osm. Perhaps OS forward translates these schedules regardless of their existence in the osm?

joseph-robertson: Or perhaps we have a reporting measure that's injecting it?

shorowit: Probably forward translator, yeah. You can go ahead and remove the FIXME comments.

joseph-robertson: Although if it is the forward translator, I still wonder what it's forward translating these schedules based on? Because I only see this for a small subset of the test xml files...

shorowit:
Sorry, I just looked at the implementation in more detail and noticed that you're creating many identical objects but with different names. It works, but isn't very clean. For example:
```
ScheduleTypeLimits,
  res mech vent range exhaust schedule,   !- Name
  0,                                      !- Lower Limit Value {BasedOnField A3}
  1,                                      !- Upper Limit Value {BasedOnField A3}
  Discrete;                               !- Numeric Type

ScheduleTypeLimits,
  res natural vent avail schedule,        !- Name
  0,                                      !- Lower Limit Value {BasedOnField A3}
  1,                                      !- Upper Limit Value {BasedOnField A3}
  Discrete;                               !- Numeric Type
```

I'm thinking of a simpler implementation where:
- We define constants for the unique types of objects (e.g., Constants.ScheduleTypeLimitsFraction, Constants.ScheduleTypeLimitsOnOff, Constants.ScheduleTypeLimitsWhatever)
- We pass those constants into the schedule classes instead of the 4 separate arguments
- We use the specified constant to reference (and create the first time) the ScheduleTypeLimits object.

This would result in the above two objects simply becoming one:
```
ScheduleTypeLimits,
  OnOff(OrWhateverItIs),                  !- Name
  0,                                      !- Lower Limit Value {BasedOnField A3}
  1,                                      !- Upper Limit Value {BasedOnField A3}
  Discrete;                               !- Numeric Type
```
and it would be referenced by multiple objects.

PR #101, opened 2019-05-20
Generalized HPWH model

jmaguire1:
Moving most of the esoteric HPWH inputs to waterheater.rb, hardcoding to use the 1st gen AO Smith parameters, and adding a relationship between the COP and UEF


jmaguire1:
This branch uses simulation data of the UEF test procedure to come up with a relationship between UEF and COP. All of the esoteric HPWH inputs are also moved into waterheater.rb, so the model now only takes EF (and the EF_adj) as inputs.

PR #112, opened 2019-06-12
Desuperheater OS Workflow

yzhou601:
Add desuperheater approach


Rakefile
```diff
@@ -53,6 +53,8 @@ def create_hpxmls
     'invalid_files/unattached-window.xml' => 'base.xml',
     'invalid_files/water-heater-location.xml' => 'base.xml',
     'invalid_files/water-heater-location-other.xml' => 'base.xml',
+    'invalid_files/desuperheater_1_speed_invalid_relatedHVAC.xml' => 'base-hvac-central-ac-only-1-speed.xml',
+    'invalid_files/desuperheater_1_speed_repeated_relatedHVAC.xml' => 'base-hvac-central-ac-only-1-speed.xml',
```
shorowit: Great to see these 👍. But can you use dashes instead of underscores for consistency with all the other files?

measure.rb
```diff
@@ -266,6 +266,8 @@ def self.create(hpxml_doc, runner, model, weather, map_tsv_dir)
     @hvac_map = {} # mapping between HPXML HVAC systems and model objects
     @dhw_map = {}  # mapping between HPXML Water Heating systems and model objects
 
+    @related_hvac_list = [] # list of hvac systems refered in water heating system "RelatedHvac" element
```
shorowit: Since this variable is only used in the `add_hot_water_and_appliances`, can we define it there instead of making it a global?

measure.rb
```diff
@@ -1872,10 +1877,18 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
           capacity_kbtuh = water_heating_system_values[:heating_capacity] / 1000.0
           oncycle_power = 0.0
           offcycle_power = 0.0
+          if has_desuperheater
+            if not @related_hvac_list.include? relatedhvac
+              @related_hvac_list << relatedhvac
+              coil_object = get_desuperheatercoil(@hvac_map, relatedhvac, sys_id)
+            else
+              fail "RelatedHVACSystem '#{relatedhvac}' for water heating system '#{sys_id}' is already attached to another water heating system."
+          end
```
shorowit: Misaligned "end", looks like the rake task needs to be run.

measure.rb
```diff
@@ -1872,10 +1877,18 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
+          end
           success = Waterheater.apply_tank(model, runner, space, to_beopt_fuel(fuel),
                                            capacity_kbtuh, tank_vol, ef, re, setpoint_temp,
                                            oncycle_power, offcycle_power, ec_adj,
-                                           @nbeds, @dhw_map, sys_id)
+                                           @nbeds, @dhw_map, sys_id, has_desuperheater, coil_object)
```
shorowit: We'll want to allow desuperheater for tankless water heater and HPWH too, right? I guess HPWH can't be implemented right now because it requires connecting to a stratified tank (?), but we should be able to add support for tankless? If so, let's add a test HPXML for it.

measure.rb
```diff
@@ -1932,6 +1945,30 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
+    coil_found = false
+    # Supported cooling coil options
+    clg_coil_supported = [OpenStudio::Model::CoilCoolingDXSingleSpeed, OpenStudio::Model::CoilCoolingDXMultiSpeed, OpenStudio::Model::CoilCoolingWaterToAirHeatPumpEquationFit]
+    if hvac_map.keys.include? relatedhvac
+      target_hvac = hvac_map[relatedhvac]
+      target_hvac.each do |comp|
+        clg_coil_supported.each do |coiltype|
+          if comp.is_a? coiltype
+            return comp
+            coil_found = true
```
shorowit: Something is weird here. You're returning from the method right above assigning a value (which means the `coil_found = true` will never execute).

measure.rb
```diff
@@ -1932,6 +1945,30 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
     return true
   end
 
+  def self.get_desuperheatercoil(hvac_map, relatedhvac, wh_id)
+    # search for the related cooling coil object for desuperheater
+
+    coil_found = false
+    # Supported cooling coil options
+    clg_coil_supported = [OpenStudio::Model::CoilCoolingDXSingleSpeed, OpenStudio::Model::CoilCoolingDXMultiSpeed, OpenStudio::Model::CoilCoolingWaterToAirHeatPumpEquationFit]
```
shorowit: Nice!

measure.rb
```diff
@@ -1932,6 +1945,30 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
+      target_hvac = hvac_map[relatedhvac]
+      target_hvac.each do |comp|
+        clg_coil_supported.each do |coiltype|
+          if comp.is_a? coiltype
+            return comp
+            coil_found = true
+           end
+        end
+      end
+      if coil_found == false
```
shorowit:
```suggestion
      if not coil_found
```

resources/EPvalidator.rb
```diff
@@ -437,6 +437,11 @@ def self.run_validator(hpxml_doc)
         "TankVolume" => one,
       },
 
+      ## [Have desuperheater]
+      "/HPXML/Building/BuildingDetails/Systems/WaterHeating/WaterHeatingSystem[HasDesuperheater='true']" => {
+        "RelatedHVACSystem" => one,
+      },
```
shorowit: This is good, indeed the "RelatedHVACSystem" requirement should be conditional based on the HasDesuperheater value. But you'll also want to add a requirement to the WaterHeatingSystem like:
`"HasDesuperheater" => zero_or_one, # See [Desuperheater]`
And please rename "[Have desuperheater]" to just "[Desuperheater]".

resources/hpxml.rb
```diff
@@ -1232,7 +1237,9 @@ def self.get_water_heating_system_values(water_heating_system:)
              :heating_capacity => to_float_or_nil(XMLHelper.get_value(water_heating_system, "HeatingCapacity")),
              :energy_factor => to_float_or_nil(XMLHelper.get_value(water_heating_system, "EnergyFactor")),
              :uniform_energy_factor => to_float_or_nil(XMLHelper.get_value(water_heating_system, "UniformEnergyFactor")),
-             :recovery_efficiency => to_float_or_nil(XMLHelper.get_value(water_heating_system, "RecoveryEfficiency")) }
+             :recovery_efficiency => to_float_or_nil(XMLHelper.get_value(water_heating_system, "RecoveryEfficiency")),
+             :has_desuperheater => to_bool_or_nil(XMLHelper.get_value(water_heating_system, "HasDesuperheater")),
+             :related_hvac => HPXML.get_idref(water_heating_system, "RelatedHVACSystem") }
```
shorowit: 👍

resources/waterheater.rb
```diff
@@ -75,6 +75,9 @@ def self.apply_tank(model, runner, space, fuel_type, cap, vol, ef,
       new_heater.addToNode(storage_tank.supplyOutletModelObject.get.to_Node.get)
     end
 
+    if has_desuperhater
+      add_desuperheater(model, t_set, new_heater, related_clg_coil, Constants.WaterHeaterTypeTank)
```
shorowit: Great use of a method. Should more easily allow accommodating tankless & HPWH.

resources/waterheater.rb
```diff
@@ -9,7 +9,7 @@
 
 class Waterheater
   def self.apply_tank(model, runner, space, fuel_type, cap, vol, ef,
-                      re, t_set, oncycle_p, offcycle_p, ec_adj, nbeds, dhw_map, sys_id)
+                      re, t_set, oncycle_p, offcycle_p, ec_adj, nbeds, dhw_map, sys_id, has_desuperhater, related_clg_coil)
```
shorowit: "desuperhater" -> "desuperheater"

I hope this doesn't mean you're subconsciously hating desuperheaters! 🤣

measure.rb
shorowit: If indeed the intent is to return with the coil as soon as you find it, you can probably remove the `coil_found` variable altogether. If you get to the code after the loop, by definition it will mean that the coil wasn't found.

resources/waterheater.rb
yzhou601: Hahahahahaha, I've already fixed some places with this misspelling, one more left out!

resources/EPvalidator.rb
yzhou601: Okay, I added that. Did this requirement really functional, or it's just for pointing to the right location where desuperheater requirements lie.

measure.rb
```diff
@@ -1888,10 +1900,19 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
           offcycle_power = 0.0
+
+          if has_desuperheater
+            if not related_hvac_list.include? relatedhvac
+              related_hvac_list << relatedhvac
+              coil_object = get_desuperheatercoil(@hvac_map, relatedhvac, sys_id)
+            else
+              fail "RelatedHVACSystem '#{relatedhvac}' for water heating system '#{sys_id}' is already attached to another water heating system."
+            end
+          end
```
shorowit: Please try to avoid duplicate code two places :) Also, because we will ultimately allow desuperheater with any water heating system, it seems like the code can simply be inserted right after has_desuperheater and relatedhvac are retrieved.

measure.rb
```diff
@@ -1888,10 +1900,19 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
           success = Waterheater.apply_tankless(model, runner, space, to_beopt_fuel(fuel),
                                                capacity_kbtuh, ef, cycling_derate,
                                                setpoint_temp, oncycle_power, offcycle_power, ec_adj,
-                                               @nbeds, @dhw_map, sys_id)
+                                               @nbeds, @dhw_map, sys_id, has_desuperheater, coil_object)
```
shorowit: In all the water heater methods, how about just passing the desuperheater coil_object?

resources/waterheater.rb
```diff
@@ -9,7 +9,7 @@
 
 class Waterheater
   def self.apply_tank(model, runner, space, fuel_type, cap, vol, ef,
-                      re, t_set, oncycle_p, offcycle_p, ec_adj, nbeds, dhw_map, sys_id)
+                      re, t_set, oncycle_p, offcycle_p, ec_adj, nbeds, dhw_map, sys_id, has_desuperheater, related_clg_coil)
```
shorowit: If we just have a single argument, I suggest calling it desuperheater_clg_coil instead of related_clg_coil.

resources/waterheater.rb
```diff
@@ -75,11 +75,14 @@ def self.apply_tank(model, runner, space, fuel_type, cap, vol, ef,
       new_heater.addToNode(storage_tank.supplyOutletModelObject.get.to_Node.get)
     end
 
+    if has_desuperheater
```
shorowit: If we just have a single argument, you can check `if not desuperheater_clg_coil.nil?` instead.

resources/EPvalidator.rb
shorowit: Good question. You're right that the requirement isn't really functional, since `zero_or_one` implies that the element is optional. It's mostly as you said, to point to the right location. When software developers are writing HPXML files that get used by this OS measure, we point them to this EPvalidator.rb file so they can understand what's required. If they are working on a WaterHeatingSystem element, they will now more easily see that they can specify a desuperheater. Previously, it would be easy for someone to not notice the desuperheater section.

measure.rb
```diff
@@ -1932,6 +1953,26 @@ def self.add_hot_water_and_appliances(runner, model, building, weather, spaces)
...
   end
 
+  def self.get_desuperheatercoil(hvac_map, relatedhvac, wh_id)
+    # search for the related cooling coil object for desuperheater
+
+    # Supported cooling coil options
+    clg_coil_supported = [OpenStudio::Model::CoilCoolingDXSingleSpeed, OpenStudio::Model::CoilCoolingDXMultiSpeed, OpenStudio::Model::CoilCoolingWaterToAirHeatPumpEquationFit]
+    if hvac_map.keys.include? relatedhvac
+      target_hvac = hvac_map[relatedhvac]
+      target_hvac.each do |comp|
```
shorowit: Seems like there's no point to create the `target_hvac` variable. Just go with `hvac_map[relatedhvac].each do |comp|`.

measure.rb
yzhou601: I agree that we could shift this part out of the water heater type conditions. But for E+ capability, the desuperheater is not allowed together with heat pump water heater (because they both require connecting with tank source sides which is only one space available currently) I am wondering if any check need to be added for this concern.

shorowit: So, this makes me think that we should add test files for desuperheater with:
1. Multispeed cooling coil
2. GSHP cooling coil
3. HPWH
Let's add them as *.xml.skip, so that they don't run on the CI, since they will fail. But you should be able to at least demonstrate that the first two run locally with your E+ PR, right? And we'll continue to investigate the HPWH question.

yzhou601: Yes, I could definitely add the .skip and start with testing first two.

measure.rb
yzhou601: Do you mean adding a new "if not desuperheater_clg_coil.nil?" check in add_desuperheater function to avoid error for those systems without desuperheater?

yzhou601:
@shorowit We might need to make minor changes for the [desuperheater element name change](https://github.com/hpxmlwg/hpxml/pull/184) ?

shorowit:
@yzhou601 Yes we'll have to make that change at some point, but I don't think it needs to be in your branch. We can make the change down the road when we pull in the latest HPXML v3 schema.

yzhou601:
image

Pulled results from CI and plotted the diffs between desuperheater & rake-base model without desuperheater. Looked good, four desuperheater options followed similar trends of saving water heating energy while kept anything else not related to water heating untouched. EC adjustment decreased 0.1 GJ( from -0.7GJ to -0.8GJ) for desuperheater models except for gshp (both -0.7GJ w/ or w/o desuperheater), not familiar with EC adjustment workflow, is it expected? @jmaguire1 @shorowit 
GSHP had a very small impact (0.01GJ) on cooling load report, maybe related to the changes made to reflect impact on hvac plants, maybe this is minor enough to neglect?

PR #116, opened 2019-06-14
Fixing SSB combined draw profile. Results now match within 0.01%

jmaguire1:
Fixed some of the scaling factors associated with the showers, sinks, and baths. Results are now consistent with the disaggregated draw profiles


jmaguire1:
@shorowit I also compared hourly for all possible combinations of bedrooms and they all looked good, it's not unique to 3 bedrooms.

PR #120, opened 2019-06-19
FoundationWall insulation flexibility

shorowit:
Allows:
1. Insulation on a foundation wall that doesn't start at the top of the wall. 
2. Separate interior vs exterior insulation specifications.

For walls described with an insulation layer, both interior and exterior layers now must be described. For example, an uninsulated wall that used to be described as
```
          <FoundationWall>
            <SystemIdentifier id='FoundationWall'/>
            ...
            <DistanceToBottomOfInsulation>0</DistanceToBottomOfInsulation>
            <Insulation>
              <SystemIdentifier id='FoundationWallInsulation'/>
              <Layer>
                <InstallationType>continuous</InstallationType>
                <NominalRValue>0</NominalRValue>
              </Layer>
            </Insulation>
          </FoundationWall>
```
is now described as
```
          <FoundationWall>
            <SystemIdentifier id='FoundationWall'/>
            ...
            <Insulation>
              <SystemIdentifier id='FoundationWallInsulation'/>
              <Layer>
                <InstallationType>continuous - exterior</InstallationType>
                <NominalRValue>0</NominalRValue>
                <extension>
                  <DistanceToTopOfInsulation>0</DistanceToTopOfInsulation>
                  <DistanceToBottomOfInsulation>0</DistanceToBottomOfInsulation>
                </extension>
              </Layer>
              <Layer>
                <InstallationType>continuous - interior</InstallationType>
                <NominalRValue>0</NominalRValue>
                <extension>
                  <DistanceToTopOfInsulation>0</DistanceToTopOfInsulation>
                  <DistanceToBottomOfInsulation>0</DistanceToBottomOfInsulation>
                </extension>
              </Layer>
            </Insulation>
          </FoundationWall>
```

For walls described with an assembly R-value, there is no change required. A wall can continue to be described like so:
```
          <FoundationWall>
            <SystemIdentifier id='FoundationWall'/>
            ...
            <Insulation>
              <SystemIdentifier id='FoundationWallInsulation'/>
              <AssemblyEffectiveRValue>10.69</AssemblyEffectiveRValue>
            </Insulation>
          </FoundationWall>
```


yzhou601:
@shorowit I have a question in proceeding exterior/interior insulation implementation:

It looks like the space of interior insulation in Kiva has already been taken with logic [in these lines](https://github.com/NREL/OpenStudio-HPXML/blob/master/resources/constructions.rb#L1463-L1468). I am not sure if we want to: 1. only use user-defined interior insulation layers in the same way as exterior or 2. address both user-defined + current logic. Can you clarify more about this interior wall r value? Is this for some structures (frame, cavity, gap etc.) we didn't address in the foundation wall construction?

yzhou601:
After changing exterior insulation to be no offset (starting from the top of foundation walls) for base test files, CI only shows diffs in autosize test files, which means using CustomBlock is giving consistent results with previous approach. And switching to using layer extension elements instead of `DistanceToTopOfInsuation` + `DistanceToBottomOfInsuation` shows no CI diff (before and after 7d60b7b).

shorowit:
> @shorowit I have a question in proceeding exterior/interior insulation implementation:
> 
> It looks like the space of interior insulation in Kiva has already been taken with logic [in these lines](https://github.com/NREL/OpenStudio-HPXML/blob/master/resources/constructions.rb#L1463-L1468). I am not sure if we want to: 1. only use user-defined interior insulation layers in the same way as exterior or 2. address both user-defined + current logic. Can you clarify more about this interior wall r value? Is this for some structures (frame, cavity, gap etc.) we didn't address in the foundation wall construction?

When insulating the interior of basement walls, there are two approaches. One is to simply add rigid insulation to the interior of the concrete wall. However, you wouldn't want to look at wall insulation if you actually want to live/play in your basement, so in this case it's common to create an insulated wood frame wall on the interior of the concrete wall and finish it with drywall.

In BEopt, we had inputs to describe either situation. But via HPXML, we just ask for the insulation R-value (whether describing just a layer of rigid insulation or an interior framed wall). So we're [not actually using that code](https://codecov.io/gh/NREL/OpenStudio-HPXML/src/master/resources/constructions.rb#L1464...1468). I suggest you remove the `calc_interior_wall_r_value` method and hook up to the interior vertical insulation inputs instead. If we ever use additional FoundationWall inputs in HPXML (e.g., framing factor, cavity insulation R-value, etc.), we can bring the code back.

shorowit:
And you can also remove many of the arguments for the `apply_foundation_wall` method that are only used for that purpose -- wall_cavity_r, wall_install_grade, etc.

measure.rb
```diff
@@ -1649,41 +1649,42 @@ def self.add_foundation_wall(runner, model, spaces, fnd_wall_values, slab_frac,
     else
       drywall_thick_in = 0.0
     end
```
yzhou601: @shorowit Just want to confirm, we still want this drywall assumption (excluded from interior/exterior insulation) based on thermal boundary, right?

shorowit: Yes, and we should use it as part of the surface's construction (not the Kiva insulation blocks).

measure.rb
```diff
@@ -4585,13 +4590,9 @@ def self.check_surface_assembly_rvalue(runner, surfaces, film_r, assembly_r, mat
...
-        if foundation.interiorVerticalInsulationMaterial.is_initialized
-          int_mat = foundation.interiorVerticalInsulationMaterial.get.to_StandardOpaqueMaterial.get
-          constr_r += UnitConversions.convert(int_mat.thickness, "m", "ft") / UnitConversions.convert(int_mat.thermalConductivity, "W/(m*K)", "Btu/(hr*ft*R)")
-        end
-        if foundation.exteriorVerticalInsulationMaterial.is_initialized
-          ext_mat = foundation.exteriorVerticalInsulationMaterial.get.to_StandardOpaqueMaterial.get
-          constr_r += UnitConversions.convert(ext_mat.thickness, "m", "ft") / UnitConversions.convert(ext_mat.thermalConductivity, "W/(m*K)", "Btu/(hr*ft*R)")
+        foundation.customBlocks.each do |custom_block|
+          ins_mat = custom_block.material.to_StandardOpaqueMaterial.get
+          constr_r += UnitConversions.convert(ins_mat.thickness, "m", "ft") / UnitConversions.convert(ins_mat.thermalConductivity, "W/(m*K)", "Btu/(hr*ft*R)")
```
yzhou601: I removed the codes of getting interior/exterior vertical insulation r values here because foundation walls have switched to using custom blocks. Sometimes, slab will use interior/exterior vertical insulation too, but it never calls this function. Is it ok to remove this or is it better to keep it for future use of interior/exterior vertical insulation inputs?

resources/constructions.rb
```diff
@@ -1459,25 +1395,37 @@ def self.create_kiva_slab_foundation(model, int_horiz_r, int_horiz_width, int_ve
...
-    if int_vert_r > 0 and int_vert_depth > 0
-      int_vert_mat = create_insulation_material(model, "FoundationIntVertIns", int_vert_r)
-      foundation.setInteriorVerticalInsulationMaterial(int_vert_mat)
-      foundation.setInteriorVerticalInsulationDepth(UnitConversions.convert(int_vert_depth, "ft", "m"))
+    if not int_vert_r.nil?
+      if int_vert_r > 0 and int_vert_depth > 0
+        int_vert_mat = create_insulation_material(model, "FoundationIntVertIns", int_vert_r)
+        foundation.addCustomBlock(int_vert_mat,
+                                  UnitConversions.convert(int_vert_depth, "ft", "m"),
+                                  -int_vert_mat.thickness,
```
yzhou601: Interior layer x position starts at negative of its thickness. Need to verify. It looks like kiva 0.5.0 (standalone program) is not supporting inputs similar to custom block.

resources/hpxml.rb
```diff
@@ -2055,22 +2089,38 @@ def self.add_layer_insulation(parent:,
...
       XMLHelper.add_element(layer, "InstallationType", "cavity")
       XMLHelper.add_element(layer, "NominalRValue", Float(cavity_nominal_r_value)) unless cavity_nominal_r_value.nil?
+      layers[:exterior] = layer
     end
     unless continuous_nominal_r_value.nil?
       layer = XMLHelper.add_element(insulation, "Layer")
       XMLHelper.add_element(layer, "InstallationType", "continuous")
       XMLHelper.add_element(layer, "NominalRValue", Float(continuous_nominal_r_value)) unless continuous_nominal_r_value.nil?
+    else
+      unless ext_continuous_nominal_r_value.nil?
```
yzhou601: Add ext/int continuous path for foundation wall to keep other function calls untouched, add layer elements as return values, to proceed with adding extension.

measure.rb
```diff
@@ -1648,39 +1650,42 @@ def self.add_foundation_wall(runner, model, spaces, fnd_wall_values, slab_frac,
...
       drywall_thick_in = 0.0
     end
-    filled_cavity = true
     concrete_thick_in = fnd_wall_values[:thickness]
-    cavity_r = 0.0
-    cavity_depth_in = 0.0
-    install_grade = 1
-    framing_factor = 0.0
     assembly_r = fnd_wall_values[:insulation_assembly_r_value]
     if not assembly_r.nil?
```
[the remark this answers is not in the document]
yzhou601: Need a check that insulation layer height is not greater than foundation wall height?

tests/hpxml_translator_test.rb
```diff
@@ -42,7 +42,7 @@ def test_simulations
 
     xmls = []
     test_dirs.each do |test_dir|
-      Dir["#{test_dir}/base*.xml"].sort.each do |xml|
+      Dir["#{test_dir}/base-foundation*.xml"].sort.each do |xml|
```
yzhou601: Is this intended?

shorowit: Oops. 🤦‍♂ No, I was just testing those files to make sure I handled the merge correctly. Good catch, please fix!

Rakefile
```diff
@@ -1158,18 +1159,24 @@ def get_hpxml_file_foundation_walls_values(hpxml_file, foundation_walls_values)
...
                                  :depth_below_grade => 7,
-                                 :insulation_distance_to_bottom => 8,
-                                 :insulation_r_value => 8.9 }]
+                                 :exterior_layer_distance_to_top => 1,
+                                 :exterior_layer_height => 7,
+                                 :exterior_layer_r_value => 8.9 }]
+  elsif ['base-foundation-conditioned-basement-wall-interior-insulation.xml'].include? hpxml_file
+    foundation_walls_values[0][:interior_layer_distance_to_top] = 0
+    foundation_walls_values[0][:interior_layer_height] = 8
+    foundation_walls_values[0][:interior_layer_r_value] = 10
```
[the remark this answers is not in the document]
yzhou601: Reverted back to top + bottom inputs with original names instead of top + height

yzhou601:
image
@shorowit I changed insulations to be thicker to better inspect its location (changed interior x position to be corresponding (- its thickness)). Kiva inputs are only slightly different from E+, assuming I converted this small portion of inputs correctly, the profile looks great.

yzhou601:
image
@shorowit I changed insulations to be thicker to better inspect its location (changed interior x position to be corresponding (- its thickness)). Kiva inputs are only slightly different from E+, assuming I converted this small portion of inputs correctly, the profile looks great.

measure.rb
```diff
@@ -1486,7 +1486,9 @@ def self.add_foundation_walls_slabs(runner, model, building, spaces)
         assembly_r = fnd_wall_values[:insulation_assembly_r_value]
         if assembly_r.nil?
           concrete_thick_in = fnd_wall_values[:thickness]
-          assembly_r = fnd_wall_values[:insulation_r_value] + Material.Concrete(concrete_thick_in).rvalue + Material.GypsumWall(drywall_thick_in).rvalue + film_r
+          int_r = fnd_wall_values[:interior_layer_r_value].nil? ? 0.0 : fnd_wall_values[:interior_layer_r_value]
+          ext_r = fnd_wall_values[:exterior_layer_r_value].nil? ? 0.0 : fnd_wall_values[:exterior_layer_r_value]
```
shorowit: Seems like we can simplify these lines to, e.g., `int_r = fnd_wall_values[:interior_layer_r_value]` because the layer R-values are now required when assembly R-value is not provided.

measure.rb
```diff
@@ -1539,40 +1541,44 @@ def self.add_foundation_wall(runner, model, spaces, fnd_wall_values, slab_frac,
...
       end
     else
-      rigid_height = fnd_wall_values[:insulation_distance_to_bottom]
-      rigid_r = fnd_wall_values[:insulation_r_value]
+      if not fnd_wall_values[:distance_to_exterior_insulation_top].nil?
+        ext_rigid_offset = fnd_wall_values[:distance_to_exterior_insulation_top]
+        ext_rigid_height = fnd_wall_values[:distance_to_exterior_insulation_bottom] - ext_rigid_offset
+        ext_rigid_r = fnd_wall_values[:exterior_layer_r_value]
+      end
+      if not fnd_wall_values[:distance_to_interior_insulation_top].nil?
```
shorowit: Don't need the nil checks here and above because the fields are now required.

measure.rb
shorowit: Always good to remove unused code -- it's available in git if we ever need it again in the future. Good change.

resources/EPvalidator.rb
```diff
@@ -169,16 +169,23 @@ def self.run_validator(hpxml_doc)
...
+        "Insulation/Layer[InstallationType='continuous - interior'] | Insulation/AssemblyEffectiveRValue" => one,
+        "Insulation/Layer[InstallationType='continuous - exterior'] | Insulation/AssemblyEffectiveRValue" => one,
       },
 
       ## [VentedCrawlspace]
       "/HPXML/Building/BuildingDetails/Enclosure/FoundationWalls/FoundationWall[InteriorAdjacentTo='crawlspace - vented']" => {
         "../../Foundations/Foundation[FoundationType/Crawlspace[Vented='true']]/VentilationRate[UnitofMeasure='SLA']/Value" => zero_or_one,
       },
 
+      ## [Insulation/Layer]
```
shorowit: How about "FoundationWallInsLayer" instead of "Insulation/Layer"? It helps to denote that it's specific to foundation walls.

resources/EPvalidator.rb
```diff
@@ -169,16 +169,23 @@ def self.run_validator(hpxml_doc)
         "Thickness" => one,
         "DepthBelowGrade" => one,
         "Insulation/SystemIdentifier" => one, # Required by HPXML schema
-        # Either specify insulation layer R-value and insulation height OR assembly R-value:
-        "[DistanceToBottomOfInsulation | Insulation/AssemblyEffectiveRValue]" => one,
-        "[Insulation/Layer[InstallationType='continuous']/NominalRValue | Insulation/AssemblyEffectiveRValue]" => one,
+        # Insulation: either specify interior and exterior layers OR assembly R-value:
+        "Insulation/Layer[InstallationType='continuous - interior'] | Insulation/AssemblyEffectiveRValue" => one,
+        "Insulation/Layer[InstallationType='continuous - exterior'] | Insulation/AssemblyEffectiveRValue" => one,
```
shorowit: These two lines need comments "See [FoundationWallInsLayer]"

shorowit: Nice elegant xpath expression! 😁

resources/constructions.rb
```diff
@@ -820,21 +820,15 @@ def self.apply_floor(model, surfaces, constr_name,
...
-                                 wall_rigid_r, wall_drywall_thick_in, wall_concrete_thick_in,
+                                 ext_rigid_ins_offset, int_rigid_ins_offset, ext_rigid_ins_height,
+                                 int_rigid_ins_height, ext_rigid_r, int_rigid_r, wall_drywall_thick_in, wall_concrete_thick_in,
                                  wall_height, wall_height_above_grade)
 
-    # Calculate interior wall R-value
-    int_wall_rvalue = calc_interior_wall_r_value(wall_cavity_depth_in, wall_cavity_r,
-                                                 wall_filled_cavity, wall_framing_factor,
-                                                 wall_install_grade, wall_rigid_r,
-                                                 wall_drywall_thick_in)
```
shorowit: Good cleanup of removing method and unused arguments. 👍

resources/constructions.rb
```diff
@@ -1380,25 +1319,37 @@ def self.create_kiva_slab_foundation(model, int_horiz_r, int_horiz_width, int_ve
...
     # Create the Foundation:Kiva object for crawl/basement foundations
     foundation = OpenStudio::Model::FoundationKiva.new(model)
 
     # Interior vertical insulation
-    if int_vert_r > 0 and int_vert_depth > 0
-      int_vert_mat = create_insulation_material(model, "FoundationIntVertIns", int_vert_r)
-      foundation.setInteriorVerticalInsulationMaterial(int_vert_mat)
-      foundation.setInteriorVerticalInsulationDepth(UnitConversions.convert(int_vert_depth, "ft", "m"))
+    if not int_vert_r.nil?
+      if int_vert_r > 0 and int_vert_depth > 0
```
shorowit: Just combine into a single if statement with 3 clauses, no need for nested if statements.

resources/constructions.rb
```diff
@@ -1380,25 +1319,37 @@ def self.create_kiva_slab_foundation(model, int_horiz_r, int_horiz_width, int_ve
...
+        int_vert_mat = create_insulation_material(model, "FoundationIntVertIns", int_vert_r)
+        foundation.addCustomBlock(int_vert_mat,
+                                  UnitConversions.convert(int_vert_depth, "ft", "m"),
+                                  -int_vert_mat.thickness,
+                                  UnitConversions.convert(int_vert_offset, "ft", "m"))
+      end
+    end
+
+    if wall_drywall_thick_in < 0.0
+      wall_drywall_thick_in = 0.0
```
shorowit: Why is this check needed? Seems like it can't ever be negative.

resources/constructions.rb
```diff
@@ -1380,25 +1319,37 @@ def self.create_kiva_slab_foundation(model, int_horiz_r, int_horiz_width, int_ve
...
     end
 
     # Exterior vertical insulation
-    if ext_vert_r > 0 and ext_vert_depth > 0
-      ext_vert_mat = create_insulation_material(model, "FoundationExtVertIns", ext_vert_r)
-      foundation.setExteriorVerticalInsulationMaterial(ext_vert_mat)
-      foundation.setExteriorVerticalInsulationDepth(UnitConversions.convert(ext_vert_depth, "ft", "m"))
+    if not ext_vert_r.nil?
+      if ext_vert_r > 0 and ext_vert_depth > 0
```
shorowit: Combine into a single if statement.

measure.rb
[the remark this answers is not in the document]
shorowit: Not for now.

shorowit:
> Options of solutions are:
> 
>     1. keep as it is
> 
>     2. add `Depth Below Slab` to exterior insulation depth
> 
>     3. reduce wall height (if height inputs already included Depth Below Slab)?

Looks good to me. Let's go with 1. keep as it is.

shorowit:
Thank you @yzhou601!

PR #122, opened 2019-06-21
Allow water heaters to be located outside

shorowit:
The `WaterHeatingSystem/Location` can now be set to "other exterior".

Updated the code for tank/tankless and HPWH. Adds three test files (conventional storage, heat pump water heater, and tankless).

Only tested to make sure the test files run, model results still need to be reviewed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -76,8 +76,11 @@ def create_hpxmls
     'base-dhw-recirc-temperature.xml' => 'base.xml',
     'base-dhw-recirc-timer.xml' => 'base.xml',
     'base-dhw-tank-gas.xml' => 'base.xml',
+    'base-dhw-tank-gas-outside.xml' => 'base-dhw-tank-gas.xml',
     'base-dhw-tank-heat-pump.xml' => 'base.xml',
+    'base-dhw-tank-heat-pump-outside.xml' => 'base-dhw-tank-heat-pump.xml',
     'base-dhw-tankless-electric.xml' => 'base.xml',
+    'base-dhw-tankless-electric-outside.xml' => 'base-dhw-tankless-electric.xml',
     'base-dhw-tankless-gas.xml' => 'base.xml',
     'base-dhw-tankless-oil.xml' => 'base.xml',
     'base-dhw-tankless-propane.xml' => 'base.xml',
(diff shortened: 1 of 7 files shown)


shorowit:
@yzhou601 is going to simplify my implementation for mixed tank water heaters and do some testing. (FYI @jmaguire1)

yzhou601:
image

@shorowit @jmaguire1 , the same energy results using these two approaches

shorowit:
Awesome, thanks @yzhou601! I'm embarrassed about my implementation. 😞

shorowit:
Ready again for your review, @jmaguire1.

yzhou601:
@shorowit Never mind, it happens especially when trapped somewhere in the logic. 
I wish I could notice it earlier. But anyway, it now looks good!

jmaguire1:
Just pushed a fix for HPWHs being outside, we weren't overwriting the schedule correctly so it was using the Schedule:Constant values that are supposed to be overwritten. Also checked that E+ won't error out if the tank temperature goes below 0 C, but maybe we ought to do something in that situation (at least a warning)

jmaguire1:
@shorowit It looks like this works correctly for tank/tankless water heaters, if the HPWH change makes sense then I think this is ok to merge

jmaguire1:
@shorowit @nmerket (cc: @yzhou601): I was thinking more about the possibility of the water in the tank freezing over the weekend. I got it to happen by setting the element capacity to 0 and locating the water heater outside in someplace cold, but I could see this actually happen for certain HPWHs that only have an upper backup heating element if the water heater is outside on a very cold day (which would disable the heat pump). We handle the possibility of boiling in the tank by modeling a temperature and pressure relief valve, but it looks like EnergyPlus does nothing in the case of freezing. Should EnergyPlus throw a warning/error if the water in the tank freezes? Since this would be a catastrophic failure for the tank, my first thought is E+ ought to error, but we might not want to do something so drastic.

nmerket:
I can say for a certainty that the tank models don't check for freezing. The tank model is also used for chilled water tanks where that might be a real concern. 

It may be something good to bring up with the E+ team. Let's submit a issue on the NREL/EnergyPlus repo and see what they have to say about it. @jmaguire1 would you like to do that or should I?

jmaguire1:
Added an issue to E+ for the potential freezing issue with both @nmerket and @shorowit tagged.

PR #126, opened 2019-06-26
Tank jacket insulation

yzhou601:
Tank jacket insulation option added

image

water heating system energy results
--
3 bedroom water use
--

#### Electric water heater
**Baseline: 14.2**

|JacketRValue | Side | Top+side | diff | perc diff|Side Saving |Side+top Saving|
|:---|---|---|---|---|---|---:|
|1 | 14.13 | 14.12 | 0.01 | 0.07%|0.49%|0.56%|
|3 | 14.01 | 13.99 | 0.02 | 0.14%|1.34%|1.48%|
|6 | 13.89 | 13.86 | 0.03 | 0.22%|2.18%|2.39%|
|10 | 13.78 | 13.74 | 0.04 | 0.29%|2.96%|3.24%|
|16 | 13.67 | 13.63 | 0.04 | 0.29%|3.73%|4.01%|

#### Gas water heater
**Baseline: 24.25**

|JacketRValue | Side | Top+side | diff | perc diff|Side Saving |Side+top Saving|
|:---|---|---|---|---|---|---:|
|1 | 24 | 23.98 | 0.02 | 0.08%|1.03%|1.11%|
|3 | 23.67 | 23.62 | 0.05 | 0.21%|2.39%|2.60%|
|6 | 23.37 | 23.3 | 0.07 | 0.30%|3.63%|3.92%|
|10 | 23.16 | 23.06 | 0.1 | 0.43%|4.49%|4.91%|
|16 | 22.98 | 22.86 | 0.12 | 0.52%|5.24%|5.73%|

1 bedroom water use
--

#### Electric water heater
**Baseline: 8.68**

|JacketRValue | Side | Top+side | diff | perc diff|Side Saving |Side+top Saving|
|:---|---|---|---|---|---|---:|
|1 | 8.6 | 8.6 |0 | 0.00%|0.92%|0.92%|
|3 | 8.48 | 8.47 | 0.01 | 0.12%|2.3%|2.42%|
|6 | 8.36 | 8.34 | 0.02 | 0.24%|3.69%|3.92%|
|10 | 8.25 | 8.22 | 0.03 | 0.36%|4.95%|5.30%|
|16 | 8.15 | 8.11 | 0.04 | 0.49%|6.11%|6.57%|

#### Gas water heater
**Baseline: 19.70**

|JacketRValue | Side | Top+side | diff | perc diff|Side Saving |Side+top Saving|
|:---|---|---|---|---|---|---:|
|1 | 19.44 | | | |1.32%| |
|3 | 19.08 | | | |3.15%| |
|6 | 18.78 | | | |4.67%| |
|10 |18.56 | | | |5.79%| |
|16 | 18.38 | | | |6.70%| |

Just ignore the Top+side combined with gas water heater if it's not realistic. @shorowit @jmaguire1


resources/EPvalidator.rb
```diff
@@ -410,6 +410,8 @@ def self.run_validator(hpxml_doc)
         "[WaterHeaterType='storage water heater' or WaterHeaterType='instantaneous water heater' or WaterHeaterType='heat pump water heater' or WaterHeaterType='space-heating boiler with storage tank' or WaterHeaterType='space-heating boiler with tankless coil']" => one, # See [WHType=Tank] or [WHType=Tankless] or [WHType=HeatPump] or [WHType=Indirect] or [WHType=CombiTankless]
         "[Location='living space' or Location='basement - unconditioned' or Location='basement - conditioned' or Location='attic - unvented' or Location='attic - vented' or Location='garage' or Location='crawlspace - unvented' or Location='crawlspace - vented']" => one,
         "FractionDHWLoadServed" => one,
+        "[EnergyFactor | UniformEnergyFactor]" => one,
```
shorowit: Was this an accident?

resources/EPvalidator.rb
```diff
@@ -410,6 +410,8 @@ def self.run_validator(hpxml_doc)
...
+        "WaterHeaterInsulation/Jacket/JacketRValue" => zero_or_one, # Capable to model tank wrap insulation
```
shorowit: We will need to decide which water heating system types to allow this for. For example, we probably shouldn't offer it as an input for tankless or tankless coil, so the above specification should probably become conditional for certain water heater types. Ideally we'd allow it for conventional storage, HPWHs, and indirect water heaters.

resources/EPvalidator.rb
yzhou601: Oops! Must be the conflicts I solved today, yeah for sure it's an error!

resources/EPvalidator.rb
yzhou601: Do we want to add error checks for those combinations not allowed (tankless + jacket)? Since it seems even if we remove this line to the certain water heater type blocks, it didn't give proper errors but simply ignored this input, correct?

shorowit: That's correct, it would be ignored. If someone accidentally provided a jacket insulation R-value for a tankless system, it shouldn't affect results, so ignoring it seems fine to me. (Or maybe they'd want to specify a jacket R-value of zero, to simplify their HPXML export code across different system types?)

yzhou601: Emm I am not sure if I followed. I didn't read this input in the tankless tank types in measure.rb for apply_tankless function and inside the `` calc_tank_ua`` function, it also bypassed the tankless ua recalculation for jacket. So it would ignore the input for those systems even if the EPvalicator checked that (zero_or_one might be fine? ) But removing this line to be inside of different types would be more clear to see the available combinations.

yzhou601:
image

Results look reasonable after testing different cases.

Rakefile
```diff
@@ -2202,6 +2205,12 @@ def get_hpxml_file_water_heating_system_values(hpxml_file, water_heating_systems
   elsif ['base-dhw-uef.xml'].include? hpxml_file
     water_heating_systems_values[0][:energy_factor] = nil
     water_heating_systems_values[0][:uniform_energy_factor] = 0.93
+  elsif ['base-dhw-jacket-electric.xml'].include? hpxml_file
+    water_heating_systems_values[0][:jacket_r_value] = 10.0
+  elsif ['base-dhw-jacket-indirect.xml'].include? hpxml_file
+    water_heating_systems_values[0][:jacket_r_value] = 10.0
+  elsif ['base-dhw-jacket-gas.xml'].include? hpxml_file
+    water_heating_systems_values[0][:jacket_r_value] = 10.0
```
shorowit: Since all three elsif statements make the same change, you can combine with using:
```
elsif ['base-dhw-jacket-electric.xml',
       'base-dhw-jacket-indirect.xml',
       'base-dhw-jacket-gas.xml'].include? hpxml_file
```

shorowit:
@yzhou601 Can you pull in `master` before we review? There are a couple conflicts right now.

yzhou601:
> @yzhou601 Can you pull in `master` before we review? There are a couple conflicts right now.

Sorry my bad, should be fine now.

Rakefile
```diff
@@ -93,6 +93,10 @@ def create_hpxmls
     'base-dhw-tank-oil.xml' => 'base.xml',
     'base-dhw-tank-propane.xml' => 'base.xml',
     'base-dhw-uef.xml' => 'base.xml',
+    'base-dhw-jacket-electric.xml' => 'base.xml',
+    'base-dhw-jacket-gas.xml' => 'base-dhw-tank-gas.xml',
+    'base-dhw-jacket-indirect.xml' => 'base-dhw-indirect.xml',
+    'base-dhw-jacket-hpwh.xml' => 'base-dhw-tank-heat-pump.xml',
```
shorowit: Nice set of test files.

resources/hpxml.rb
```diff
@@ -1221,8 +1222,15 @@ def self.add_water_heating_system(hpxml:,
     XMLHelper.add_element(water_heating_system, "EnergyFactor", Float(energy_factor)) unless energy_factor.nil?
     XMLHelper.add_element(water_heating_system, "UniformEnergyFactor", Float(uniform_energy_factor)) unless uniform_energy_factor.nil?
     XMLHelper.add_element(water_heating_system, "RecoveryEfficiency", Float(recovery_efficiency)) unless recovery_efficiency.nil?
-    related_hvac_el = XMLHelper.add_element(water_heating_system, "RelatedHVACSystem") unless related_hvac.nil?
-    XMLHelper.add_attribute(related_hvac_el, "idref", related_hvac) unless related_hvac.nil?
+    unless jacket_r_value.nil?
+      water_heater_insulation = XMLHelper.add_element(water_heating_system, "WaterHeaterInsulation")
+      jaket = XMLHelper.add_element(water_heater_insulation, "Jacket")
```
shorowit: jaket -> jacket

resources/waterheater.rb
```diff
@@ -59,7 +59,7 @@ def self.apply_tankless(model, runner, space, fuel_type, cap, ef,
     new_manager = create_new_schedule_manager(t_set, model, Constants.WaterHeaterTypeTankless)
     new_manager.addToNode(loop.supplyOutletNode)
 
-    new_heater = create_new_heater(Constants.ObjectNameWaterHeater, cap, fuel_type, 1, ef, 0, t_set, space, oncycle_p, offcycle_p, ec_adj, Constants.WaterHeaterTypeTankless, cd, nbeds, model, runner)
+    new_heater = create_new_heater(Constants.ObjectNameWaterHeater, cap, fuel_type, 1, ef, 0, 0, t_set, space, oncycle_p, offcycle_p, ec_adj, Constants.WaterHeaterTypeTankless, cd, nbeds, model, runner)
```
shorowit: How about passing nil instead of 0, since it doesn't apply and shouldn't be used in the calculation?

resources/hpxml.rb
yzhou601: Again! 🤦‍♀😂

resources/waterheater.rb
```diff
@@ -126,6 +126,17 @@ def self.apply_heatpump(model, runner, space, weather, t_set, vol, ef,
...
     a_tank = 2 * pi * r_tank * (r_tank + h_tank)
+
+    # water heater wrap calculation based on:
+    # Modeling Water Heat Wraps in BEopt DRAFT Technical Note
+    # Authors:  Ben Polly and Jay Burch (NREL)
+    if not jacket_r.nil?
+      a_side = 2 * pi * UnitConversions.convert(r_tank, "m", "ft") * UnitConversions.convert(h_tank, "m", "ft") # sqft
+      skin_insulation_t = 2.0 # inch
+      skin_insulation_R = 5.0 # R5
+      u_pre_skin = 1 / (skin_insulation_t * skin_insulation_R + 1.0 / 1.3 + 1.0 / 52.8) # Btu/hr-ft^2-F = (1 / hout + kins / tins + t / hin)^-1
```
jmaguire1: Should this be tied to the U value that we're already using? We already have a UA calculated earlier on in the HPWH method, I think we ought to be using that value here for "u_pre_skin"

yzhou601: Is that only accounts for skin losses? Since the reason we cannot use UA calculated previously is that it not only accounts for skin but also shorts and flues. We kind of want to separate UA into UAskin, UAshorts, and UAflue and only apply jacket to the skin portion. (The ua in hpwh is hardcoded so I am not sure what it accounts for). But that ua value was used in the next line to be adjusted by subtracting something calculated as reflecting the jacket impact.

jmaguire1: OK, that's what I thought when I mentioned I might be misinterpreting things, the UA calculated previously is based on data from real HPWHs that were lab tested so it includes shorts (no flue in a HPWH). I think now that I understand this better I see the logic in what you're doing.

PR #130, opened 2019-07-09
Allow combi boiler water heaters to be located outside

shorowit:


jmaguire1:
@shorowit Is the intention for this one that only the tank is outside, or the boiler is as well? If the boiler is this could potentially raise additional concerns about freezing (although I doubt the boiler would ever be off if it's below 0 outside unless there's some sort of sequential load with multiple HVAC systems where the boiler is lowest priority)

shorowit:
Just the water heater tank.

jmaguire1:
Got it, thanks! That makes more sense

PR #142, opened 2019-07-22
Use hpxml values to set zone volumes

joseph-robertson:


resources/geometry.rb
```diff
@@ -19,6 +19,69 @@ def self.make_polygon(*pts)
...
+    end
+    building.elements.each("BuildingDetails/Enclosure/RimJoists/RimJoist") do |rim_joist|
+      rim_joist_values = HPXML.get_rim_joist_values(rim_joist: rim_joist)
+      thermal_zones << rim_joist_values[:interior_adjacent_to]
+    end
+    building.elements.each("BuildingDetails/Enclosure/FrameFloors/FrameFloor") do |framefloor|
+      framefloor_values = HPXML.get_framefloor_values(framefloor: framefloor)
+      thermal_zones << framefloor_values[:interior_adjacent_to]
+    end
+    building.elements.each("BuildingDetails/Enclosure/Slabs/S;an") do |slab|
```
shorowit: Typo

resources/geometry.rb
```diff
@@ -19,6 +19,69 @@ def self.make_polygon(*pts)
...
+      slab_values = HPXML.get_slab_values(slab: slab)
+      thermal_zones << slab_values[:interior_adjacent_to]
+    end
+    return thermal_zones.uniq
+  end
+  
+  def self.get_model_thermal_zone(model:,
```
shorowit: I don't understand what this method is doing. How is it different than just returning every thermal zone of the model? Why would some zones be excluded?

resources/geometry.rb
```diff
@@ -19,6 +19,69 @@ def self.make_polygon(*pts)
     return p
   end
 
+  def self.get_thermal_zones(building:)
+    thermal_zones = []
+    building.elements.each("BuildingDetails/Enclosure/Walls/Wall") do |wall|
+      wall_values = HPXML.get_wall_values(wall: wall)
+      thermal_zones << wall_values[:interior_adjacent_to]
```
shorowit: The code needs to check `exterior_adjacent_to` values for all surfaces as well. There is no convention in HPXML, so a wall between garage and living space can either be interior=living space and exterior=garage OR interior=garage and exterior=living space.

resources/geometry.rb
```diff
@@ -61,6 +124,28 @@ def self.get_height_of_spaces(spaces)
...
+  def self.get_height_of_thermal_zone(building:,
+                                      thermal_zone:)
+    heights = []
+    building.elements.each("BuildingDetails/Enclosure/FoundationWalls/FoundationWall") do |foundation_wall|
+      foundation_wall_values = HPXML.get_foundation_wall_values(foundation_wall: foundation_wall)
+      next if foundation_wall_values[:interior_adjacent_to] != thermal_zone
+
+      height = foundation_wall_values[:height]
+      if building.elements["BuildingDetails/Enclosure/RimJoists/RimJoist"]
+        height += 1.0
```
shorowit: Need to find out a way to avoid making similar assumptions different places in the code.

resources/geometry.rb
```diff
@@ -61,6 +124,28 @@ def self.get_height_of_spaces(spaces)
...
+      end
+      heights << height
+    end
+    building.elements.each("BuildingDetails/Enclosure/Walls/Wall") do |wall|
+      wall_values = HPXML.get_wall_values(wall: wall)
+      next if wall_values[:interior_adjacent_to] != thermal_zone
+
+      heights << 8.0
```
shorowit: Same comment as above.

shorowit: Seems like this needs to incorporate the number of above-grade stories for 2+ story homes?

resources/geometry.rb
joseph-robertson: You pass in a string (e.g., "living space"), and the corresponding thermal zone from the model is returned. In this case, we need it since we are manually setting the volume on this model object.

resources/geometry.rb
joseph-robertson: Good point. Although, I'll need to exclude "outside" and "garage" (I think).

resources/geometry.rb
joseph-robertson: `Constants.RimJoistHeight`?

resources/geometry.rb
joseph-robertson: `Constants.WallHeight`?

resources/geometry.rb
shorowit: Oh, right. 🤦‍♂

resources/geometry.rb
shorowit: Garage should not be excluded, but "ground" should. Maybe that's what you meant?

resources/geometry.rb
joseph-robertson: Is "living space" the only thermal zone that can have more than one story?

resources/geometry.rb
joseph-robertson: Good catch.

resources/geometry.rb
joseph-robertson: Yes, "ground" instead of "garage" is what I meant.

resources/geometry.rb
shorowit: Yes.

PR #144, opened 2019-07-26
Evaporative cooler

shorowit:


yzhou601:
@shorowit I tried following this [standard approach](https://github.com/NREL/openstudio-standards/blob/49626ee957db63129bb74cfe08b48abd571de759/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L3764) starting from osm and idf  generated in this PR. Fan seemed to be cycling on/off and reporting energy use reasonably when some changes are made to air loop mass flow rate and some objects are autosized (many sizing objects are added in this case). 

There're some issues using this approach:
1. The dummy unitary system sizing is locally ignored with some dummy method (simply quote search unitary system function), but need to improve it to be smart enough to filter dummy unitary systems if following this approach.
2. It seems hard to skip a E+ sizing run following this os standard approach, but I notice there's a separate sizing approach for our workflow, so I would think it's unexpected to run sizing from multiple sources. The attempt to remove all "autosize" failed when I hardcode the ``Cooling Supply Air Flow Rate`` of dummy unitary system, E+ disappointingly crashed (zero or non-zero values both caused crash). It's just too smart to  be dummy 💢 .
3. It seems for ``AirLoopHVAC`` object, the``DesignSupplyAirFlowRate`` was override somewhere (maybe in hvac sizing?) which caused I cannot set in apply_evaporative_cooler. But how to size this system should be investigated as next step if we still want to use this approach.

yzhou601:
Evaporative cooler model tested in newly built E+ 9.2 develop branch, pump and fan are cycling petty well. To move forward to system sizing:
1. The only sizing property is the max mass/volume flow rate of air loop/evap cooler/unitary system/oa because capacity is not deterministic (dependent on outdoor air condition)
2. The only energy consumption is from fan and pump. 
*Fan energy: `DesignVolumeFlowRate * PressureRise  * 3600 * RTF / Efficiency`;*
*Evap energy: `PumpPower * RTF * 3600`*

I think these make evap cooler different from other cooling systems in terms of hvac sizing, I don't know if we could still call `hvac_sizing` or hardcode values. @shorowit Any suggestion?

yzhou601:
TODO:
1. Evap cooler recirculating water pump power assumption. This will fix evap cooler fan power disaggregation program which is currently looking for empty evap cooler electricity use. 
2. Evap cooler ducted system without return ducts will require some changes in existing duct program, need to discuss details how to modify it.
3. Manual S- Air balance for conditioned space: A large amount of supply air flows into the room. There must be a method for getting this air out of the conditioned space (i.e., exhuast Cfm equal to , or approximately equal to supply Cfm). Do we need to consider this? Add a test file with evap cooler + exhaust fan?

resources/hvac.rb
```diff
@@ -1900,7 +1996,12 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_objects, clg_objects)
 
     clg_object_sensors = []
     clg_objects.each do |clg_object|
-      clg_object_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Cooling Coil Electric Energy")
+      if clg_object.is_a? OpenStudio::Model::EvaporativeCoolerDirectResearchSpecial
+        var = "Evaporative Cooler Electric Energy"
```
yzhou601: Need assumption for recirculating water pump power, or this variable will always report 0, and fan power will not be correctly disaggregated.

resources/hvac.rb
```diff
@@ -4112,13 +4221,28 @@ def self.get_unitary_system_from_air_loop_hvac(air_loop)
...
   def self.get_unitary_system_air_loops(model, runner, thermal_zone)
     # Returns the unitary system(s), cooling coil(s), heating coil(s), and air loops(s) if available
     unitary_system_air_loops = []
     thermal_zone.airLoopHVACs.each do |air_loop|
       system = get_unitary_system_from_air_loop_hvac(air_loop)
       next if system.nil?
 
+      # skip evap cooler dummy unitary system
+      if system.additionalProperties.getFeatureAsString(Constants.SizingInfoHVACCoolType).is_initialized
+        next if system.additionalProperties.getFeatureAsString(Constants.SizingInfoHVACCoolType).get == Constants.ObjectNameEvaporativeCooler
```
yzhou601: This filtered out evap cooler dummy unitary system, which blocked sizing this system in hvac_sizing.rb.

resources/hvac_sizing.rb
```diff
@@ -146,6 +146,7 @@ def self.process_site_calcs_and_design_temps(runner, weather)
 
     @cool_design_temps = {}
     @heat_design_temps = {}
+    @wetbulb_outdoor_cooling = weather.design.CoolingWetbulb
```
yzhou601: Need to double check if this is the proper wet bulb temperature for calculation

resources/hvac_sizing.rb
```diff
@@ -1653,6 +1664,11 @@ def self.process_cooling_equipment_adjustments(runner, hvac_final_values, weathe
 
       end
 
+    elsif hvac.has_type(Constants.ObjectNameEvaporativeCooler)
+      hvac_final_values.Cool_Capacity = hvac_final_values.Cool_Load_Tot
+      hvac_final_values.Cool_Capacity_Sens = hvac_final_values.Cool_Load_Sens
+      hvac_final_values.Cool_Airflow = calc_airflow_rate(hvac_final_values.Cool_Load_Sens, (@cool_setpoint - hvac.LeavingAirTemp))
```
yzhou601: for hot humid climate, maybe `@cool_setpoint - hvac.LeavingAirTemp` will be negative. Might need check how to deal with this situation.

resources/hvac_sizing.rb
```diff
@@ -3451,6 +3472,34 @@ def self.set_object_values(runner, model, hvac, hvac_final_values)
...
+        # Air Loop
+        vfr = UnitConversions.convert(hvac_final_values.Cool_Airflow, "cfm", "m^3/s")
+        object.setPrimaryAirDesignFlowRate(vfr)
+        air_loop = object.airLoopHVAC.get
+        air_loop.setDesignSupplyAirFlowRate(vfr)
+        unitary_sys = HVAC.get_unitary_system_from_air_loop_hvac(air_loop)
+        fan = unitary_sys.supplyFan.get.to_FanOnOff.get
+        fan.setMaximumFlowRate(vfr)
+        # Fan pressure rise calculation (based on design cfm)
+        fan_power = [2.79 * (hvac_final_values.Cool_Airflow / @cfa)**(-0.29), 0.6].min
```
yzhou601: fan power assumption by Leo's suggestion.

tests/hpxml_translator_test.rb
```diff
@@ -1111,6 +1111,7 @@ def _test_multiple_hvac(xmls, hvac_multiple_dir, hvac_base_dir, all_results)
     puts "Multiple HVAC test results:"
     xmls.sort.each do |xml|
       next if not xml.include? hvac_multiple_dir
+      next if xml.include? "evap-cooler" # skip for evap cooler, neither energy nor capacity is linearly scaled for evap cooler
```
yzhou601: Though good to check multiple and partial systems working as expected (with scaled airflow sizes, able to meet setpoint, etc.), there's no linearly scaled meter to check its output relationship.

resources/hvac_sizing.rb
yzhou601: Use Leo's rule of thumb sizing method for those situations

resources/hvac.rb
yzhou601: Used evap cooler water consumption to disaggregate.

yzhou601:
Update:
1. I checked wet bulb temperature we used: mean coincident wet bulb temperature at 1%, in both manual S and [EVAPORATIVE COOLING DESIGN GUIDELINES MANUAL](http://www.emnrd.state.nm.us/ECMD/Multimedia/documents/EvapCoolingDesignManual.pdf), I got some clues this is the correct value to use.
2. As far as Leo's suggestion [here](https://github.com/NREL/OpenStudio-HEScore/issues/50#issuecomment-545171938), it seemed the fan efficacy assumption we implemented already included pump power, and it's hard to split this number into two parts. Except for results being reported in two places( Evaporative Cooler Electric Energy vs. Fan Electric Energy), the different inputs to apply evap cooler power can also make a difference on final energy use, eg. fan might cycle differently with evap cooler, because evap cooler part load performance considers both partial airflow and wetness control while fan only looked at air flow ratio. For now, I kept fan power(that is, defining pressure rise) as where we input power, might need further discussion.
3. Some results got from simulations ran recently:


  | Airflow | Unmet load | Fan pressure rise | Fan energy
-- | -- | -- | -- | --
Evap cooler only | 0.470623 | 8.0709E+04 | 598.58494 | 3.6090E+08
Evap cooler + furnace | 0.470623 | 1.5032E+05 | 598.58494 | 3.7678E+08
Evap cooler multiple | 0.156717444534639 * 3 | 3.2837E+05 | 823.411550771869 * 3 | 3.8400E+08
Evap cooler partial | 0.1567174 | 3.0187E+08 | 823.41155 | 3.3220E+08
Evap cooler ducted | 0.682517 | 1.7682E+06 | 537.41289 | 5.3973E+08

It's encouraging that the unmet load looks not significant, which means the sizing seemed to be working well for Denver. I checked the hourly cooling energy of ideal loads air system for ducted system that accounted for the highest unmet load:

MAX | 1291386
-- | --
SUM | 1768221

So there's just one hour accounted 70% of unmet load, and that hour the evap cooler is fully on, so I guess this is not a big issue, there're some other conditions evap cooler not able to meet full cooling demand, but solved most. I also noticed there being specific hours when a small load is ignored by evap cooler but captured by ideal loads air system(about 1000-10000J/h scale), might be brought by imperfect control

yzhou601:
Evap cooler was delivering higher cooling energy so failed the load test. From my inspection, it looked like fan was fully on for many hours when there's need of cooling. It cooled zone falling into deadband but below cooling setpoint, the overcooling thus made a larger cooling load reported by Meter:EnergyTransfer. Investigating better approach to control fan cycling to meeting load.

resources/hvac.rb
```diff
@@ -369,6 +369,9 @@ def self.apply_evaporative_cooler(model, runner, frac_cool_load_served,
     evap_cooler = OpenStudio::Model::EvaporativeCoolerDirectResearchSpecial.new(model, model.alwaysOnDiscreteSchedule)
     evap_cooler.setName(obj_name)
     evap_cooler.setCoolerEffectiveness(cooler_effectiveness)
+    evap_cooler.setEvaporativeOperationMinimumDrybulbTemperature(0) # relax limitation to open evap cooler for any potential cooling
+    evap_cooler.setEvaporativeOperationMaximumLimitWetbulbTemperature(50) # relax limitation to open evap cooler for any potential cooling
+    evap_cooler.setEvaporativeOperationMaximumLimitDrybulbTemperature(50) # relax limitation to open evap cooler for any potential cooling
```
yzhou601: These three limitations caused evap cooler closed for some available hours. This turned out to be the main reason causing aggregated fan energy inconsistent with metered fan energy.

yzhou601:
It's confusing that **IdealLoadsAirSystem cooling energy + Evap Terminal cooling energy != EnergyTransfer:Cooling**, especially for partial hvac here for evap cooler. The evap cooler zone air terminal cooling energy is about 1/3 of total load except for some hours unable to meet 1/3, so I think it's operating normally. It looked like the issue was in backup ideal system. I checked other partial hvac test results on CI, unmet loads were also over 66.667%. Might need further investigation.

measure.rb
```diff
@@ -3388,7 +3398,7 @@ def self.create_ducts(air_distribution, model, spaces)
   end
 
   def self.add_hvac_sizing(runner, model, weather)
-    success = HVACSizing.apply(model, runner, weather, @cfa, @infilvolume, @nbeds, @min_neighbor_distance, false, @living_space)
+    success = HVACSizing.apply(model, runner, weather, @cfa, @infilvolume, @nbeds, @min_neighbor_distance, false, @living_space, @living_zone)
```
shorowit: This doesn't really need to be passed, since you can always do `@living_space.thermalZone.get` inside HVACSizing.apply().

Rakefile
```diff
@@ -186,9 +185,9 @@ def create_hpxmls
     'base-hvac-ducts-multiple.xml' => 'base.xml',
     'base-hvac-ducts-outside.xml' => 'base.xml',
     'base-hvac-elec-resistance-only.xml' => 'base.xml',
-    'base-hvac-evap-cooler-furnace-gas.xml' => 'base.xml',
-    'base-hvac-evap-cooler-only.xml' => 'base.xml',
-    'base-hvac-evap-cooler-only-ducted.xml' => 'base.xml',
+    'base-hvac-evap-cooler-furnace-gas.xml' => 'base-location-dallas-tx.xml',
```
shorowit: Changed the location to have a higher cooling load.

Rakefile
```diff
@@ -265,12 +264,14 @@ def create_hpxmls
     'cfis/base-hvac-central-ac-only-var-speed-cfis.xml' => 'base-hvac-central-ac-only-var-speed.xml',
     'cfis/base-hvac-dse-cfis.xml' => 'base-hvac-dse.xml',
     'cfis/base-hvac-ducts-in-conditioned-space-cfis.xml' => 'base-hvac-ducts-in-conditioned-space.xml',
+    'cfis/base-hvac-evap-cooler-only-ducted-cfis.xml' => 'base-hvac-evap-cooler-only-ducted.xml',
```
shorowit: Added CFIS test files so that we cover all ducted systems.

measure.rb
```diff
@@ -3368,6 +3368,8 @@ def self.create_ducts(air_distribution, model, spaces)
         duct_leakage_cfm = duct_leakage_value
       elsif duct_leakage_units == 'Percent'
         duct_leakage_frac = duct_leakage_value
+      else
+        fail "#{duct_side.capitalize} ducts exist but leakage was not specified for distribution system '#{dist_id}'."
```
shorowit: Moved error-checking from EPvalidator.rb to here.

measure.rb
```diff
@@ -3398,7 +3402,7 @@ def self.create_ducts(air_distribution, model, spaces)
   end
 
   def self.add_hvac_sizing(runner, model, weather)
-    success = HVACSizing.apply(model, runner, weather, @cfa, @infilvolume, @nbeds, @min_neighbor_distance, false, @living_space, @living_zone)
+    success = HVACSizing.apply(model, runner, weather, @cfa, @infilvolume, @nbeds, @min_neighbor_distance, false, @living_space)
```
shorowit: No need to pass the zone, it can easily be retrieved from the space.

resources/EPvalidator.rb
```diff
@@ -307,15 +307,15 @@ def self.run_validator(hpxml_doc)
       "/HPXML/Building/BuildingDetails/Systems/HVAC/HVACPlant/CoolingSystem[CoolingSystemType='central air conditioner']" => {
         "../../HVACDistribution[DistributionSystemType/AirDistribution | DistributionSystemType[Other='DSE']]" => one_or_more, # See [HVACDistribution]
         "DistributionSystem" => one,
-        "AnnualCoolingEfficiency[Units='SEER']/Value" => one,
         "CoolingCapacity" => one, # Use -1 for autosizing
+        "AnnualCoolingEfficiency[Units='SEER']/Value" => one,
```
shorowit: Since order matters in HPXML elements, I like to try to preserve the order in this file. It doesn't have any impact on functionality, but helps any software developers referencing this file when producing HPXML files.

resources/EPvalidator.rb
```diff
@@ -401,9 +401,9 @@ def self.run_validator(hpxml_doc)
       ## [HVACDistType=Air]
       "/HPXML/Building/BuildingDetails/Systems/HVAC/HVACDistribution/DistributionSystemType/AirDistribution" => {
         "DuctLeakageMeasurement[DuctType='supply']/DuctLeakage[Units='CFM25' or Units='Percent'][TotalOrToOutside='to outside']/Value" => one,
-        "DuctLeakageMeasurement[DuctType='return']/DuctLeakage[Units='CFM25' or Units='Percent'][TotalOrToOutside='to outside']/Value" => zero_or_one, # See [HVACDuctLeakageMeasurement]
-        "Ducts[DuctType='supply']" => one_or_more, # See [HVACDuct]
-        "Ducts[DuctType='return']" => zero_or_more, # return ducts are optional for eg. evap cooler, See [HVACDuct]
+        "DuctLeakageMeasurement[DuctType='return']/DuctLeakage[Units='CFM25' or Units='Percent'][TotalOrToOutside='to outside']/Value" => zero_or_one,
+        "Ducts[DuctType='supply']" => zero_or_more, # See [HVACDuct]
+        "Ducts[DuctType='return']" => zero_or_more, # See [HVACDuct]
```
shorowit: What you had was correctly implemented, but fairly complex. I decided to simplify it -- partially by taking the opportunity to offer more flexibility of duct inputs in the measure, and partially by throwing an error message in measure.rb instead for the one situation where the inputs are incomplete (which I would expect to be rarely hit, so I'm comfortable not calling it out in EPvalidator.rb).

resources/airflow.rb
```diff
@@ -334,11 +334,13 @@ def self.create_air_loop_objects(model, runner, air_loops, mech_vent, building)
...
         fan_rtf_sensor.setName("#{fan_rtf_var.name} s")
         fan_rtf_sensor.setKeyName(supply_fan.name.to_s)
-      else
+      elsif supply_fan.to_FanVariableVolume.is_initialized
         fan_mfr_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Fan Air Mass Flow Rate")
         fan_mfr_sensor.setName("#{supply_fan.name.to_s} air MFR")
         fan_mfr_sensor.setKeyName("#{supply_fan.name.to_s}")
         fan_rtf_sensor = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, "#{fan_rtf_var.name}_s")
+      else
+        fail "Unexpected fan: #{supply_fan.name}"
```
shorowit: Seems safer.

resources/constants.rb
```diff
@@ -37,10 +37,6 @@ def self.small
     return 1e-9
   end
 
-  def self.AssumedEvapCoolerEffectiveness
-    return 0.72
```
shorowit: Hard-coded this in hvac.rb, and then updated hvac_sizing.rb to reference the object property rather than having both files reference the constant. That way they are better guaranteed to stay in sync.

resources/constants.rb
```diff
@@ -143,7 +139,7 @@ def self.DuctSideSupply
     return 'supply'
   end
 
-  def self.DuctedInfoMiniSplitHeatPumpOrEvapCooler
+  def self.OptionallyDuctedSystemIsDucted
```
shorowit: If we kept getting more optionally ducted systems, we'd end up with DuctedInfoMiniSplitHeatPumpOrEvapCoolerOrAnotherSystemOrYetAnotherSystem 😆. I figure this is better.

resources/hvac_sizing.rb
```diff
@@ -2328,17 +2328,17 @@ def self.get_hvacs(runner, model)
...
-    equips = {}
+    equips = []
 
     HVAC.existing_equipment(model, runner, @cond_zone).each do |equip|
-      next if equips.keys.include? equip
+      next if equips.include? equip
       next if equip.is_a? OpenStudio::Model::ZoneHVACIdealLoadsAirSystem
 
-      equips[equip] = @cond_zone
+      equips << equip
```
shorowit: Found one more example of code that handled multiple conditioned zones. Thanks for all that cleanup you did on that!

resources/EPvalidator.rb
yzhou601: Looks neat!

resources/constants.rb
yzhou601: Oh I did this again, thanks for improving it.

shorowit:
Thanks for all your hard work here, @yzhou601!! I will bring into ERI, can you bring into HEScore?

yzhou601:
@shorowit No problem!

PR #149, opened 2019-08-06
REUL simplified approach for ERI

yzhou601:
Use uniform load output variable for all hvac systems.

### DIFF
--- a/measure.rb
+++ b/measure.rb
@@ -2976,9 +2976,8 @@ class OSModel
   def self.add_building_output_variables(runner, model, map_tsv_dir)
     hvac_output_vars = [OutputVars.SpaceHeatingElectricity,
                         OutputVars.SpaceHeatingFuel,
-                        OutputVars.SpaceHeatingLoad,
-                        OutputVars.SpaceCoolingElectricity,
-                        OutputVars.SpaceCoolingLoad]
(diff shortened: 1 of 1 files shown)


measure.rb
```diff
@@ -3028,6 +3029,43 @@ def self.add_output_variables(model, vars, objects)
...
+      # ems output variables
+      ['clg', 'htg'].each do |load_type|
+        @hvac_map["BLDGLOAD:#{living_zone.name}_#{load_type}_load"] = []
+        ems_output_load = OpenStudio::Model::EnergyManagementSystemOutputVariable.new(model, "#{living_zone.name}_#{load_type}_load")
+        ems_output_load.setName("#{living_zone.name} #{load_type} load")
+        ems_output_load.setTypeOfDataInVariable("Summed")
+        ems_output_load.setUpdateFrequency("ZoneTimestep")
+        ems_output_load.setEMSProgramOrSubroutineName(load_program)
+        ems_output_load.setUnits("J")
+        @hvac_map["BLDGLOAD:#{living_zone.name}_#{load_type}_load"] << ems_output_load
```
yzhou601: @shorowit Since loads are parsed by building unit not system unit, which is different from other mapping data here, I'm not sure if this is the best place to store EMS load output names.

PR #150, opened 2019-08-07
Autosized baseboard bugfix

shorowit:
Fixes heating capacity/energy for autosized electric baseboard runs. Updated tests to check that autosized hvac systems have E+ capacities > 0 (previous tests only checked hardsized hvac systems).

@joseph-robertson


shorowit:
CI results confirm that only base-hvac-elec-resistance-only-autosize.xml is affected.

PR #151, opened 2019-08-07
Duct Loads

shorowit:
Fixes latent loads so that they are actually introduced as latent.

### DIFF
--- a/resources/airflow.rb
+++ b/resources/airflow.rb
@@ -1055,7 +1055,7 @@ class Airflow
     end
     return true if all_ducts_conditioned
 
-    def self.create_duct_actuator(model, name, space, is_outside = false)
+    def self.create_duct_actuator_and_equipment(model, name, space, is_latent, is_outside = false)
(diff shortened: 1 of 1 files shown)


yzhou601:
@shorowit Hi, Scott. For DSE calculation, sorry I'm a little bit puzzled by the point why we want to split the sensible and latent cooling when interpreting it as additional load(currently all added as sensible). If we only want a better REUL, why not postprocess like: sum up these additional loads and then subtract them from `predicted sensible load to setpoint heat transfer rate`? It would be quite close. But if we are doing this to improve DSE approach, I might need to dig more into duct program understanding the difference and similarity between these two.

jmaguire1:
@yzhou601 The latent loads are changes in humidity, which won't need to be met by certain pieces of equipment (for example a furnace), but are critical for others (for example a dehumidifier). So tracking the latent loads separate from the sensible will be really important for getting the right HVAC energy consumption for a lot of cases.

yzhou601:
@jmaguire1 Thanks for explanation! Got it I will look into it.

PR #152, opened 2019-08-12
Update HVAC sizing

shorowit:
Updates for internal gains and infiltration.

### DIFF
--- a/measure.rb
+++ b/measure.rb
@@ -249,6 +249,7 @@ class OSModel
       @gfa += slab_values[:area]
     end
     @cvolume = construction_values[:conditioned_building_volume]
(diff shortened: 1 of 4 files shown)


yzhou601:
@shorowit Hi, Scott. For DSE calculation, sorry I'm a little bit puzzled by the point why we want to split the sensible and latent cooling when interpreting it as additional load(currently all added as sensible). If we only want a better reference home load, why not simply sum up these additional loads and then subtract them from `predicted sensible load to setpoint heat transfer rate`? It would be quite close. But if we are doing this to improve DSE approach, I might need to dig more into duct program understanding the difference and similarity between these two.

PR #154, opened 2019-08-14
Mixed water use

shorowit:
Simplified approach for handling mixed water use.

Results in smaller IDF files (no large annual schedule) and faster workflow runtime. All RESNET hot water tests still pass. CI results look good.

### DIFF
--- a/resources/hotwater_appliances.rb
+++ b/resources/hotwater_appliances.rb
@@ -23,6 +23,8 @@ class HotWaterAndAppliances
     start_date = OpenStudio::Date.new(OpenStudio::MonthOfYear.new(1), 1, model.getYearDescription.assumedYear)
     timestep_day = OpenStudio::Time.new(1, 0)
 
+    t_mix = 105.0 # F, Temperature of mixed water at fixtures
+
     # Map plant loops to sys_ids
     dhw_loops = {}
     dhw_map.each do |sys_id, dhw_objects|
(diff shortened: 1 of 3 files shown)


resources/hotwater_appliances.rb
```diff
@@ -23,6 +23,8 @@ def self.apply(model, runner, weather, living_space,
     start_date = OpenStudio::Date.new(OpenStudio::MonthOfYear.new(1), 1, model.getYearDescription.assumedYear)
     timestep_day = OpenStudio::Time.new(1, 0)
 
+    t_mix = 105.0 # F, Temperature of mixed water at fixtures
```
jmaguire1: Will we always want t_mix = 105 F? In the HSP we specified it as 110 F, there's some contention around whether 105 or 110 is more generally used. Not particularly concerned which we assume, but with consistency with what's in BuildStock for some future point.

shorowit: My thought is to move all projects (including ResStock) over to the same assumption (ANSI/RESNET Standard, 105 F). If there is evidence to suggest that a different value should be used, we would ideally propose that change for the Standard so that all projects continue using the same assumption.

jmaguire1: @shorowit I think 105 is a fine assumption, it's what we used to have in the HSP when I started here and then eventually got changed. I don't recall there really being much data either way in terms of what outlet temperature we ought to assume.

PR #155, opened 2019-08-15
Remove DSE

shorowit:
Removes DSE approach and test files. Effect of DSE will need to be post-processed using E+ output.


resources/EPvalidator.rb
```diff
@@ -244,7 +244,7 @@ def self.run_validator(hpxml_doc)
 
       ## [HeatingType=Furnace]
       "/HPXML/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem[HeatingSystemType/Furnace]" => {
-        "../../HVACDistribution[DistributionSystemType/AirDistribution | DistributionSystemType[Other='DSE']]" => one_or_more, # See [HVACDistribution]
```
yzhou601: I guess these changes would make those files (mostly reference home hpxmls) with DSE approach not passing EPvalidator check.

resources/EPvalidator.rb
```diff
@@ -364,11 +364,6 @@ def self.run_validator(hpxml_doc)
         "Ducts[DuctType='return']" => one_or_more, # See [HVACDuct]
       },
 
-      ## [HVACDistType=DSE]
-      "/HPXML/Building/BuildingDetails/Systems/HVAC/HVACDistribution[DistributionSystemType[Other='DSE']]" => {
-        "[AnnualHeatingDistributionSystemEfficiency | AnnualCoolingDistributionSystemEfficiency]" => one_or_more,
```
yzhou601: We might still want EPvalidator to be open for DSE option, so maybe still need this?

shorowit:
Hmm, we might need to reduce hard sized hvac capacities if DSE is provided?

yzhou601:
> Hmm, we might need to reduce hard sized hvac capacities if DSE is provided?

I think by definition "the ratio between the energy consumption by the equipment if the distribution system had no losses (gains) to the outdoors or effect on the equipment or building loads, and the energy consumed by **the same equipment** connected to the distribution system under test." maybe not?

jmaguire1:
@yzhou601 I think changing the size doesn't mean that it's different equipment. If you were to perfectly size your equipment for a DSE of 1.0, it would be undersized if your DSE was 0.8. So long as it's still the same type of equipment with the same efficiency I think changing the size would still qualify as meeting the wording of the definition

yzhou601:
> @yzhou601 I think changing the size doesn't mean that it's different equipment. If you were to perfectly size your equipment for a DSE of 1.0, it would be undersized if your DSE was 0.8. So long as it's still the same type of equipment with the same efficiency I think changing the size would still qualify as meeting the wording of the definition

I understand why we might want to change capacity, I mean maybe by the definition it didn't need to think too much about sizing issue. So, maybe as long as our model is not undersized, multiply/divide this ratio has included all the impact DSE might bring. I am not familiar with those factors in its calculation so I am not 100% certain about what they are justifying.

shorowit:
Having given it some thought and after reaching out to a software developer, I recommend we assume that HVAC capacities do not incorporate any kind of DSE effect. I believe that most users would expect this. If it becomes a point of confusion in the future, we can add it to the documentation.

PR #159, opened 2019-08-23
E+ HVAC error bugfix

shorowit:
Fixes an E+ error if you have a file with, e.g., multiple furnaces and an AC, all assigned to a DSE distribution system type. Discovered by @yzhou601.


shorowit:
File now runs. CI results for other files are unchanged.

PR #179, opened 2019-09-19
Update ERV implementation with EMS

yzhou601:
Use EMS program to implement ERV/HRV instead of using ZoneHVAC:EnergyRecoveryVentilator for better consistency.

### DIFF
--- a/resources/airflow.rb
+++ b/resources/airflow.rb
@@ -67,6 +67,9 @@ class Airflow
     pbar_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Site Outdoor Air Barometric Pressure")
     pbar_sensor.setName("out pb s")
 
+    wout_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Site Outdoor Air Humidity Ratio")
+    wout_sensor.setName("out wt s")
+
     vwind_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Site Wind Speed")
     vwind_sensor.setName("site vw s")
 
@@ -104,7 +107,7 @@ class Airflow
       return false if cfis_program.nil?
     end
 
-    nv_program = process_nat_vent(model, runner, nat_vent, tin_sensor, tout_sensor, pbar_sensor, vwind_sensor, wind_speed, infil, building, weather)
+    nv_program = process_nat_vent(model, runner, nat_vent, tin_sensor, tout_sensor, pbar_sensor, vwind_sensor, wind_speed, infil, building, weather, wout_sensor)
     return false if nv_program.nil?
 
     duct_programs = {}
@@ -117,7 +120,7 @@ class Airflow
       return false if not success
     end
 
-    infil_program = create_infil_mech_vent_objects(model, runner, building, infil, mech_vent, wind_speed, tin_sensor, tout_sensor, vwind_sensor, duct_lks)
+    infil_program = create_infil_mech_vent_objects(model, runner, building, infil, mech_vent, wind_speed, tin_sensor, tout_sensor, vwind_sensor, duct_lks, wout_sensor, pbar_sensor)
     return false if infil_program.nil?
 
     create_ems_program_managers(model, infil_program, nv_program, cfis_program, duct_programs)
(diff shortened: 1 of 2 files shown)


yzhou601:
@shorowit @jonwinkler I compared `base-mechvent-erv.xml` run in this erv-ems branch with that run in erv-object master branch:

  | ERV-EMS branch | ERV-Object branch
-- | -- | --
Heating | 16.19 | 16.58
Cooling | 1.47 | 1.46

yzhou601:
Step2 update:

  | ERV-EMS 0 effectiveness | Balanced ventilation system with   Infiltration actuator
-- | -- | --
Heating | 31.22 | 31.17
Cooling | 1.42 | 1.42

It shows good consistency with approach used for balanced vent w/o heat recovery.

resources/airflow.rb
```diff
@@ -67,6 +67,9 @@ def self.apply(model, runner, infil, mech_vent, nat_vent, duct_systems,
     pbar_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Site Outdoor Air Barometric Pressure")
     pbar_sensor.setName("out pb s")
 
+    wout_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Site Outdoor Air Humidity Ratio")
+    wout_sensor.setName("out wt s")
```
shorowit: Nice. You found that we already have a sensor for this and made it more global in scope to prevent duplicate sensors.

resources/airflow.rb
```diff
@@ -1027,6 +1027,25 @@ def self.create_return_air_duct_zone(model, air_loop_name, adiabatic_const)
     return ra_duct_zone
   end
 
+  def self.create_sens_lant_load_actuator_and_equipment(model, name, space, is_latent, is_outside = false)
```
shorowit: "lant" should be "lat"?

shorowit: You are reusing existing code from the duct model to create the actuated objects. Excellent! 👍

resources/airflow.rb
```diff
@@ -1722,6 +1667,60 @@ def self.create_infil_mech_vent_objects(model, runner, building, infil, mech_ven
       infil_program.addLine("Set Qn = #{building.living.ACH * UnitConversions.convert(building.infilvolume, "ft^3", "m^3") / UnitConversions.convert(1.0, "hr", "s")}")
     end
 
+    if mech_vent.type == Constants.VentTypeBalanced and mech_vent.sensible_effectiveness > 0 and mech_vent.whole_house_cfm > 0
+      # ERV/HRV; balanced systems without energy recovery are modeled via infiltration actuated EMS
```
shorowit: This comment needs to be updated since *all* mech vent systems are now modeled via EMS. It seems we could refer to "EMS airflow model" and "EMS load model"?

shorowit: Also, I suggest adding a comment describing why we are not using the E+ ERV object.

resources/airflow.rb
```diff
@@ -1722,6 +1667,60 @@ def self.create_infil_mech_vent_objects(model, runner, building, infil, mech_ven
...
+
+      balanced_flow_rate = [UnitConversions.convert(mech_vent.whole_house_cfm, "cfm", "m^3/s"), 0.0000001].max
+
+      # Effectiveness
+      sensible_effectiveness = mech_vent.sensible_effectiveness
+      latent_effectiveness = mech_vent.latent_effectiveness
```
shorowit: These variables are fine but seem a bit pointless.

resources/airflow.rb
```diff
@@ -1722,6 +1667,60 @@ def self.create_infil_mech_vent_objects(model, runner, building, infil, mech_ven
       infil_program.addLine("Set Qn = #{building.living.ACH * UnitConversions.convert(building.infilvolume, "ft^3", "m^3") / UnitConversions.convert(1.0, "hr", "s")}")
     end
 
+    if mech_vent.type == Constants.VentTypeBalanced and mech_vent.sensible_effectiveness > 0 and mech_vent.whole_house_cfm > 0
+      # ERV/HRV; balanced systems without energy recovery are modeled via infiltration actuated EMS
+
+      balanced_flow_rate = [UnitConversions.convert(mech_vent.whole_house_cfm, "cfm", "m^3/s"), 0.0000001].max
```
shorowit: The if statement above prevents an airflow of zero, so is this really needed?

shorowit:
CI results show that only files with ERV/HRV are changed relative to the `master` branch. The changes are limited to heating/cooling load (as expected) and heating energy use (~0.4 GJ). Looks good.

resources/airflow.rb
yzhou601: The If is differentiate ERV from balanced system without heat recovery, do you want it to follow the same logic as ERV instead of actuating `Infiltration` object?

shorowit: Sorry, I am referring to the `[foo, , 0.0000001].max` part of the code here.

yzhou601:
> I see some tabs instead of spaces 😱. The rake task will obviously fix this, but I highly suggest changing your text editor's settings to insert spaces when tab is pressed.

Oops sorry, I Googled it and set Notepad++, should work in the future!

PR #184, opened 2019-09-26
Zero out basement interior solar absorptance

yzhou601:
Zero out basement interior solar absorptance and fix some remaining issues of switching to single conditioned zones.


yzhou601:
Instead of checking @ cfa and @ cfa_ag every time in every piece of surface, decided to move forward to an approach of creating a global list of conditioned basement surfaces and dealing with solar absorptance together, could also be used for zeroing out view factor. The way of checking if add to list is in `set_surface_interior` function checking its `interior_adjacent_to`. This is the same logic as previous separate zone approach to put surfaces together into conditioned basement.

resources/constructions.rb
```diff
@@ -1004,36 +1004,39 @@ def self.apply_skylight(runner, model, subsurfaces, constr_name, weather,
   end
 
   def self.apply_partition_walls(runner, model, surfaces, constr_name,
-                                 drywall_thick_in, frac_of_ffa)
+                                 drywall_thick_in, frac_of_ffa, basement_frac_of_cfa, cond_base_surfaces)
 
-    spaces = Geometry.get_conditioned_spaces(model.getSpaces)
+    living_space = Geometry.get_conditioned_space(model.getSpaces)
```
yzhou601: Came across this part which was more intended for separate zone approach. Did some cleanup, wondering maybe we could pass the conditioned zone object from `measure.rb` instead of doing extra search every time?

yzhou601:
Locally checked diffs between idfs generated from this branch and from master branch, all reflected expected construction/material creation & solar absorptance changes for conditioned basement surfaces only.

measure.rb
```diff
@@ -663,6 +671,48 @@ def self.check_for_errors(runner, model)
...
+        exterior_material = exterior_material.clone.to_StandardOpaqueMaterial.get
+        new_const.to_LayeredConstruction.get.setLayer(0, exterior_material)
+      end
+      exterior_material.setSolarAbsorptance(0.0)
+      exterior_material.setVisibleAbsorptance(0.0)
+    end
+    return true
+  end
+
+  def self.if_share_mat(model, surface1, mat1)
```
shorowit: I think you can simply use `mat1.directUseCount > 1` to see if the material is being used by multiple objects, which would obviate the need for this method. But I haven't tested it...

measure.rb
```diff
@@ -663,6 +671,48 @@ def self.check_for_errors(runner, model)
     return true
   end
 
+  def self.modify_cond_basement_surface_properties(runner, model)
+    # modify conditioned basement surface properties here
+    # zero out solar absorptance in conditioned basement
+    @cond_bsmnt_surfaces.each do |cond_bsmnt_surface|
+      # zero out solar absorptance for some surfaces
+      mat_share = false
```
shorowit: This line isn't needed as it just gets overridden later.

measure.rb
```diff
@@ -3593,6 +3659,7 @@ def self.set_surface_interior(model, spaces, surface, surface_id, interior_adjac
       surface.setSpace(create_or_get_space(model, spaces, Constants.SpaceTypeUnconditionedBasement))
     elsif ["basement - conditioned"].include? interior_adjacent_to
       surface.setSpace(create_or_get_space(model, spaces, Constants.SpaceTypeLiving))
+      @cond_bsmnt_surfaces << surface
```
shorowit: We should also do this in the `set_surface_exterior` method.

Interzonal surfaces (e.g., a surface between two zones) doesn't really have a convention for what's "interior" vs "exterior". So, a FoundationWall in HPXML that is between conditioned basement and garage could be described as interior=garage and exterior=conditioned basement.

resources/constructions.rb
```diff
@@ -1004,36 +1004,37 @@ def self.apply_skylight(runner, model, subsurfaces, constr_name, weather,
   end
 
   def self.apply_partition_walls(runner, model, surfaces, constr_name,
-                                 drywall_thick_in, frac_of_ffa)
+                                 drywall_thick_in, frac_of_ffa, basement_frac_of_cfa, cond_base_surfaces, living_space)
 
-    spaces = Geometry.get_conditioned_spaces(model.getSpaces)
-
-    return true if spaces.empty?
+    return true if living_space.nil?
```
shorowit: Remove this, shouldn't be needed.

resources/constructions.rb
```diff
@@ -1004,36 +1004,37 @@ def self.apply_skylight(runner, model, subsurfaces, constr_name, weather,
...
+      # as internal mass object.
+      obj_name = "#{living_space.name.to_s} Living Partition"
+      imdef = create_os_int_mass_and_def(runner, model, obj_name, living_space, addtl_surface_area_lv)
+      imdefs << imdef
+    end
+
+    if addtl_surface_area_base > 0
+      # Add remaining partition walls within spaces (those without geometric representation)
+      # as internal mass object.
+      obj_name = "#{living_space.name.to_s} Base Partition"
```
shorowit: "Base" -> "Basement"

resources/constructions.rb
```diff
@@ -1090,25 +1090,52 @@ def self.apply_furniture(runner, model, frac_of_ffa, mass_lb_per_sqft = 8.0,
...
+        base_surface_area = furnAreaFraction * space.floorArea * basement_frac_of_cfa
+        # living furniture mass
+        if living_surface_area > 0
+          living_obj_name = mass_obj_name_space + " living"
+          imdef = create_os_int_mass_and_def(runner, model, living_obj_name, space, living_surface_area)
+          imdefs << imdef
+        end
+        # basement furniture mass
+        if base_surface_area > 0
+          base_obj_name = mass_obj_name_space + " base"
```
shorowit: "base" -> "basement"

resources/constructions.rb
```diff
@@ -1090,25 +1090,52 @@ def self.apply_furniture(runner, model, frac_of_ffa, mass_lb_per_sqft = 8.0,
...
         return false
       end
-
-      runner.registerInfo("Assigned internal mass object '#{mass_obj_name_space}' to space '#{space.name}'.")
     end
 
     return true
   end
 
+  def self.create_os_int_mass_and_def(runner, model, object_name, space, area)
```
shorowit: 👍

resources/geometry.rb
```diff
@@ -280,16 +280,6 @@ def self.zone_is_of_type(zone, space_type)
     end
   end
 
-  def self.get_conditioned_spaces(spaces)
```
shorowit: 👍

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
...
     activity_sch = nil
 
-    # Get spaces
-    ffa_spaces = self.get_conditioned_spaces(model.getSpaces)
-
-    ffa_spaces.each do |space|
-      space_obj_name = "#{Constants.ObjectNameOccupants}|#{space.name.to_s}"
-
-      space_num_occ = num_occ * UnitConversions.convert(space.floorArea, "m^2", "ft^2") / cfa
+    space_obj_name = "#{Constants.ObjectNameOccupants}|#{space.name.to_s}"
```
shorowit: I would get rid of this line now and just use `Constants.ObjectNameOccupants` now. Since there's only one object created, no need to name it based on the space its associated with.

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
     people_sch = nil
     activity_sch = nil
 
-    # Get spaces
-    ffa_spaces = self.get_conditioned_spaces(model.getSpaces)
-
-    ffa_spaces.each do |space|
-      space_obj_name = "#{Constants.ObjectNameOccupants}|#{space.name.to_s}"
```
shorowit: Good cleanup!

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
...
-        if not people_sch.validated?
-          return false
-        end
-      end
+    space_num_occ = num_occ * UnitConversions.convert(space.floorArea, "m^2", "ft^2") / cfa
 
-      if activity_sch.nil?
-        # Create schedule
-        activity_sch = OpenStudio::Model::ScheduleRuleset.new(model, activity_per_person)
+    if people_sch.nil?
```
shorowit: No longer need `people_sch = nil` above and this line since we're no longer looping. Just create the schedule, period.

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
...
-      occ_def.setMeanRadiantTemperatureCalculationType("ZoneAveraged")
-      occ_def.setCarbonDioxideGenerationRate(0)
-      occ_def.setEnableASHRAE55ComfortWarnings(false)
-      occ.setActivityLevelSchedule(activity_sch)
-      occ.setNumberofPeopleSchedule(people_sch.schedule)
-
-      total_num_occ += space_num_occ
-
-      runner.registerInfo("Space '#{space.name}' been assigned #{space_num_occ.round(2)} occupant(s).")
+    if activity_sch.nil?
```
shorowit: No longer need `activity_sch = nil` above and this line since we're no longer looping. Just create the schedule, period.

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
...
+    occ_def.setNumberofPeople(space_num_occ)
+    occ_def.setFractionRadiant(occ_rad)
+    occ_def.setSensibleHeatFraction(occ_sens)
+    occ_def.setMeanRadiantTemperatureCalculationType("ZoneAveraged")
+    occ_def.setCarbonDioxideGenerationRate(0)
+    occ_def.setEnableASHRAE55ComfortWarnings(false)
+    occ.setActivityLevelSchedule(activity_sch)
+    occ.setNumberofPeopleSchedule(people_sch.schedule)
+
+    total_num_occ += space_num_occ
```
shorowit: Since there's just one space, these are the same. No need to have two different variables, remove one of them.

resources/geometry.rb
```diff
@@ -517,48 +507,43 @@ def self.process_occupants(model, runner, num_occ, occ_gain, sens_frac, lat_frac
...
+
+    runner.registerInfo("Space '#{space.name}' been assigned #{space_num_occ.round(2)} occupant(s).")
+
     runner.registerInfo("The building has been assigned #{total_num_occ.round(2)} total occupant(s).")
```
shorowit: Remove this line (or the line above). Just need to provide the info once now. (Before it was once for each space plus the total for the building.)

measure.rb
yzhou601: Thanks for pointing to this function! It's good to use with an extra same check for construction to avoid this edge case: material is attached once to one construction, but this construction is attached to multiple objects. I'm testing it!

measure.rb
yzhou601: Good point. I didn't notice it's putting surface to adjacent zone. Thanks.

yzhou601:
Update:
It looked like zeroing out solar absorptance of exterior material layer of some conditioned basement surfaces also had an impact on results, this was not expected because these surfaces are defined as `NoSun` for input `Sun Exposure`. Talked with Mike Witte to dig into this issue, finally found this was caused by their kiva foundation boundary condition which was ignoring `Sun Exposure` input, so the `NoSun` kiva surfaces actually was more or less exposed to sun. Therefore, zeroing out exterior solar absorptance of kiva surfaces did make a difference. With this knowledge, now we need to do:
1. Need to redo the test for interior solar absorptance impact only to eliminate kiva surface impacts. Kiva impacted previous manual tests because I zeroed out both interior and exterior surface solar absorptance. But the exterior was not something diffs between single zone and separate zone approaches. So the comparison, though went to the direction that make single zone approach closer to separate zone approach, was suspicious. Therefore whatever the exterior setting should be for kiva surfaces, it's still needed to redo that test. 
2. To improve our model, need to decide whether we want to control the sun exposure of kiva surfaces; (current E+ was not enabling `Sun Exposure` input no matter whether regarding it as a defect or lack of documentation, we could still zero out solar radiation by manually changing solar absorptance of these surfaces' outermost layer material like what I did previously)
3. If we did want to control exterior solar absorptance and interior solar absorptance separately, what should we do with those surface constructions that only have one layer? (zeroing out interior solar absorptance at the same time zeroed out exterior as well)

yzhou601:
@shorowit Hi Scott, do you mind reviewing this again?

PR #186, opened 2019-10-02
Zero view factors between conditioned basement surfaces and living surfaces

yzhou601:
- [x] Test OS SDK of wrapping up "👀 factor" object
- [x] Implement E+ approximate view factor algorithm
- [x] Implement the whole workflow from calculating approximate view factor numbers to running OS simulation
- [x] Check results of implementation
- [ ] Request review & modify code

### DIFF
--- a/measure.rb
+++ b/measure.rb
@@ -829,15 +829,12 @@ class OSModel
              (s_tilt - s2_tilt).abs > same_ang_limit
             if surface2.subSurfaces.size != 0
               # calculate surface and its sub surfaces view factors
-              parent_surface_a = surface2.grossArea
+              parent_surface_a = surface2.netArea
+              if parent_surface_a > 0.01 # base surface of a sub surface: window/door etc.
+                surface_vf_map[surface2] = parent_surface_a / zone_seen_area
+              end
               surface2.subSurfaces.each do |sub_surface|
-                parent_surface_a -= sub_surface.grossArea
-                if parent_surface_a < 0.01 # base surface of a sub surface: window/door etc.
-                  surface_vf_map[sub_surface] = surface2.grossArea / zone_seen_area
-                else
-                  surface_vf_map[surface2] = parent_surface_a / zone_seen_area
-                  surface_vf_map[sub_surface] = sub_surface.grossArea / zone_seen_area
-                end
+                surface_vf_map[sub_surface] = sub_surface.grossArea / zone_seen_area
               end
             else # no subsurface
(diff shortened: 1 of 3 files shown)


yzhou601:
The implemented approximate view factors(output by E+) looks correct. So OS interpretation should be good.

yzhou601:
This PR failed the CI by a partial HVAC check, only `base-hvac-air-to-air-heat-pump-2-speed-33percent.xml` and `base-hvac-air-to-air-heat-pump-var-speed-33percent.xml` went off tolerance (5%). I checked load numbers they are the same as base whole system, so I guess it's not relevant to this PR (impacted more on load instead of systems). Not sure if it's ok to relax a bit of tolerance? @shorowit

shorowit:
@yzhou601 You should now be able to compare CI results for [this branch](https://circleci.com/gh/NREL/OpenStudio-HPXML/1086) and [master](https://circleci.com/gh/NREL/OpenStudio-HPXML/1083). Can you look them over?

shorowit:
For what it's worth, I pulled these changes over to the ERI repo and the ASHRAE 140 basement cases now show good agreement with older results using the 2-zone model. (All the non-basement cases are correctly unchanged.)

image

It's only one building in one location, but it's encouraging!

yzhou601:
@shorowit I think it's hard to judge in which scale the view factors are negligible, eg. for a model with hundreds of surfaces, 0.01 or 0.001 might not mean "tiny", though in that case the performance issue will be more significant and assuming a limitation will give more performance gain.

measure.rb
```diff
@@ -829,15 +829,12 @@ def self.calc_approximate_view_factor(runner, model, all_surfaces)
...
               # calculate surface and its sub surfaces view factors
-              parent_surface_a = surface2.grossArea
+              parent_surface_a = surface2.netArea
+              if parent_surface_a > 0.01 # base surface of a sub surface: window/door etc.
+                surface_vf_map[surface2] = parent_surface_a / zone_seen_area
+              end
               surface2.subSurfaces.each do |sub_surface|
-                parent_surface_a -= sub_surface.grossArea
-                if parent_surface_a < 0.01 # base surface of a sub surface: window/door etc.
-                  surface_vf_map[sub_surface] = surface2.grossArea / zone_seen_area
```
yzhou601: This line is not correct if it has more than one sub surfaces, I reverted sub surface view factor calculation the same as before.

PR #194, opened 2019-10-11
Remove HVAC electric parasitics

shorowit:
Removes electric parasitics from fuel furnaces/boilers; they are already captured in EAE.


shorowit:
Turns out they weren't actually being used. Great. So results are unchanged on the CI.

PR #199, opened 2019-10-15
Allow tankless system to serve more than 5 bedrooms & 0 bedroom

yzhou601:
fix runtime fraction error when tankless system serves more than 5 bedrooms

### DIFF
--- a/resources/waterheater.rb
+++ b/resources/waterheater.rb
@@ -1072,7 +1072,11 @@ class Waterheater
       # on-cycle and off-cycle parasitics.
       # Values used here are based on the average across 10 units originally used when modeling MF buildings
       avg_runtime_frac = [0.0268, 0.0333, 0.0397, 0.0462, 0.0529]
-      runtime_frac = avg_runtime_frac[nbeds - 1]
+      if nbeds <= 5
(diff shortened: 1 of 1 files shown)


resources/waterheater.rb
```diff
@@ -1072,7 +1072,11 @@ def self.create_new_heater(name, cap, fuel, vol, ef, re, jacket_r, t_set, space,
       # on-cycle and off-cycle parasitics.
       # Values used here are based on the average across 10 units originally used when modeling MF buildings
       avg_runtime_frac = [0.0268, 0.0333, 0.0397, 0.0462, 0.0529]
-      runtime_frac = avg_runtime_frac[nbeds - 1]
+      if nbeds <= 5
+        runtime_frac = avg_runtime_frac[nbeds - 1]
+      else
+        runtime_frac = avg_runtime_frac[4]
+      end
```
shorowit: We also need to accommodate zero bedrooms. Seems like that would currently use `avg_runtime_frac[-1]`, which is incorrect.

jmaguire1: And we'd want to treat that as if it were a 1 bedroom house, so I think it's just 1 extra special case we need to catch

PR #201, opened 2019-10-15
Handle conditioned ducts w/ leakage to outside

shorowit:
When there is leakage to the outside but all ducts are in conditioned space, model the leakage as actually going outside. This can happen, e.g., for ducts in floor cavities which have leaky rims. Previously the leakage was being ignored.


shorowit:
CI results only changed for the desired case. Good to go.

PR #206, opened 2019-10-18
Bugfix for possible ducts error

shorowit:
Bugfix for locations that don't have both supply & return ducts. Added test file.

### DIFF
--- a/resources/airflow.rb
+++ b/resources/airflow.rb
@@ -1045,7 +1045,7 @@ class Airflow
   def self.create_ducts_objects(model, runner, building, ducts, mech_vent, tin_sensor, pbar_sensor, adiabatic_const, air_loop, duct_programs, duct_lks, air_loop_objects)
     return true if ducts.size == 0 # No ducts
 
-    duct_zones = ducts.map { |duct| duct.zone }.uniq!
+    duct_zones = ducts.map { |duct| duct.zone }.uniq
     living_space = building.living.zone.spaces[0]
 
     # All duct zones are in living space?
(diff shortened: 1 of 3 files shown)


resources/airflow.rb
```diff
@@ -1045,7 +1045,7 @@ def self.create_sens_lat_load_actuator_and_equipment(model, name, space, is_late
   def self.create_ducts_objects(model, runner, building, ducts, mech_vent, tin_sensor, pbar_sensor, adiabatic_const, air_loop, duct_programs, duct_lks, air_loop_objects)
     return true if ducts.size == 0 # No ducts
 
-    duct_zones = ducts.map { |duct| duct.zone }.uniq!
+    duct_zones = ducts.map { |duct| duct.zone }.uniq
```
shorowit: No idea how this code worked before...

resources/airflow.rb
```diff
@@ -1306,10 +1306,17 @@ def self.create_ducts_objects(model, runner, building, ducts, mech_vent, tin_sen
...
           duct_subroutine.addLine("  Set f_sup = #{UnitConversions.convert(leakage_cfm25s[Constants.DuctSideSupply], "cfm", "m^3/s").round(6)} / (#{fan_mfr_max_var.name} * 1.0135)") # frac
+        else
+          duct_subroutine.addLine("  Set f_sup = 0.0") # frac
+        end
+        if not leakage_fracs[Constants.DuctSideReturn].nil?
+          duct_subroutine.addLine("  Set f_ret = #{leakage_fracs[Constants.DuctSideReturn]}") # frac
+        elsif not leakage_cfm25s[Constants.DuctSideReturn].nil?
           duct_subroutine.addLine("  Set f_ret = #{UnitConversions.convert(leakage_cfm25s[Constants.DuctSideReturn], "cfm", "m^3/s").round(6)} / (#{fan_mfr_max_var.name} * 1.0135)") # frac
+        else
+          duct_subroutine.addLine("  Set f_ret = 0.0") # frac
```
shorowit: The code previously assumed there was both supply and return leakage in each duct location zone. Now it sets `f_sup` (or `f_ret`) to zero if there's no supply (or return) leakage.

resources/airflow.rb
yzhou601: Yeah.. I googled for nuances between w/ bang `!` and w/o `!`. With `!`,  it only worked when it did remove repeating stuff (otherwise it will return `nil`), maybe it worked because supply and return ducts are always placed in the same zone, so there's always a repeating element to be removed. Anyways, we are no longer in dangerous status!

resources/airflow.rb
yzhou601: 👍

PR #215, opened 2019-10-30
Reducing HPWH number of nodes from 12 to 6

jmaguire1:
When we made the change to use the AO Smith control logic, changing from 12 nodes to 6 became a bit easier so it's now pretty much a 1 line change. I did run some simulations comparing the lab data for a simulated use test, Energy consumption difference between E+ 9.2 and the lab data is 13%, compared to 4% for version 8.2 compared to lab data. Reducing to 6 nodes changed the discrepency to 11%. It looks like with the latest version of E+, the HPWH needs to run longer to reheat the tank. I'm comfortable with this level of discrepancy across versions.

image

### DIFF
--- a/resources/waterheater.rb
+++ b/resources/waterheater.rb
@@ -84,9 +84,6 @@ class Waterheater
     int_factor = 1.0 # unitless
     temp_depress = 0.0 # F
     ducting = "none"
-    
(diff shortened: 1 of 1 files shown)


jmaguire1:
Removed the flag per our conversation,

PR #221, opened 2019-11-05
Dual fuel heat pumps

shorowit:
Add support for dual fuel heat pumps (with an optional switchover temperature).


jonwinkler:
@shorowit Since we agreed to hold off on dual fuel GSHPs, do you want to revise EPvalidator and test files?

shorowit:
I ran tests for all system types and timeseries results looked good. Code changes look great, especially creating a method for applying the supplemental coil and clearly differentiating between min compressor temperature and max supplemental temperature. Nice work, @jonwinkler.

I'm going to work on incorporating this in ERI (which may require some additional changes to get loads seen by the heat pump separate from the backup system) and will merge when it's all done.

shorowit:
@jonwinkler Shouldn't the DFHP supplemental coil be autosized differently than the typical ASHP supplemental coil? I.e., using the switchover temperature?

jonwinkler:
@shorowit According the Manual S (version 1; I'm not sure where my v2 is), dual fuel heat pumps should be sized based on the cooling load, which is the same guidance for typical ASHPs. But we're sizing typical ASHPs using the ERI guidance to size on max cooling or heating load. The max heating load for a dual fuel heat pump would occur at the switchover temperature, so unless ERI states otherwise perhaps we should be using the switchover temp.

shorowit:
This is tested and good to go. Thanks, @jonwinkler!

PR #222, opened 2019-11-06
Dehumidifier

shorowit:
## Pull Request Description

Adds ability to model a dehumidifier.

## Checklist

Not all may apply:


shorowit:
Tried modeling a dehumidifier that meets 50% of the load two different ways:


  | Heating [GJ] | Cooling [GJ]
-- | -- | --
No Dehumidifier | 1.76 | 7.21
Dehumidifier - 100% | 1.12 | 14.96
Dehumidifier - 50% (using sequential load fractions backup IdealAir dehumidification) | 1.12 | 11.57
Dehumidifier - 50% (doubling energy factor to reduce electricity consumption) | 1.28 | 11.57

Test building is in Dallas, TX. Humidity setpoint is 40% to exaggerate the effect.

@jonwinkler

jonwinkler:
@shorowit In the above results, I would expect the heating energy to increase when comparing Dehumidifier - 100% to Dehumidifier - 50% because the dehumidifier will provide less sensible heating in the 50% case. Based on that, the Dehumidifier - 50% (using sequential load fractions plus backup IdealAir dehumidification) does not look right to me.

resources/hvac.rb
```diff
@@ -2137,6 +2135,28 @@ def self.apply_dehumidifier(model, runner, energy_factor, integrated_energy_fact
...
+    rh = 60.0 # for both EF and IEF test conditions, %
+
+    # Independent ariables applied to curve equations
+    var_array_ief = [1, ief_db, ief_db * ief_db, rh, rh * rh, ief_db * rh]
+
+    # Curved values under EF test conditions
+    curve_value_ef = 1 # Curves are nomalized to 1.0 under EF test conditions, 80F, 60%
+    # Curve values under IEF test conditions
+    ef_curve_value_ief = var_array_ief.zip(ef_coeff).map { |var, coeff| var * coeff }.inject(0, :+)
+    water_removal_curve_value_ief = var_array_ief.zip(w_coeff).map { |var, coeff| var * coeff }.inject(0, :+)
```
yzhou601: @jonwinkler If the IEF is specified instead of EF, I am assuming the `Capacity` input is also tested under new IEF test conditions, so I shifted it as well. Is it reasonable?

resources/hvac.rb
```diff
@@ -2118,40 +2104,57 @@ def self.apply_dehumidifier(model, runner, energy_factor, water_removal_rate,
...
+    zone_hvac.setRatedAirFlowRate(UnitConversions.convert(air_flow_rate, "cfm", "m^3/s"))
+    zone_hvac.setMinimumDryBulbTemperatureforDehumidifierOperation(10)
+    zone_hvac.setMaximumDryBulbTemperatureforDehumidifierOperation(40)
+
+    zone_hvac.addToThermalZone(control_zone)
+  end
 
-      zone_hvac.addToThermalZone(control_zone)
-    end
+  def self.dehumidifier_ief_to_ef_inputs(w_coeff, ef_coeff, ief, water_removal_rate)
```
yzhou601: Changing performance curve to be normalized to 1.0 at IEF test point is more complicated than directly applying IEF conditions to the existing curve and calculating the E+ rated inputs backwards. So I ended up with the latter approach. They should yield the same results if we want to keep the same performance curve shape. 
@jonwinkler can you take a quick look at the approach to make sure it's following what you suggested?

yzhou601:
@shorowit Oops, good catch. 
This reminds me that maybe the airflow is also supposed to be calculated based on shifted capacity.

resources/hvac.rb
```diff
@@ -2110,11 +2110,11 @@ def self.apply_dehumidifier(model, runner, energy_factor, integrated_energy_fact
     water_removal_curve = create_curve_biquadratic(model, w_coeff, "DXDH-WaterRemove-Cap-fT", -100, 100, -100, 100)
     energy_factor_curve = create_curve_biquadratic(model, ef_coeff, "DXDH-EnergyFactor-fT", -100, 100, -100, 100)
     part_load_frac_curve = create_curve_quadratic(model, pl_coeff, "DXDH-PLF-fPLR", 0, 1, 0.7, 1)
+    water_removal_rate = UnitConversions.convert(water_removal_rate, "pint", "L")
 
     if energy_factor.nil?
       # shift inputs tested under IEF test conditions to those under EF test conditions with performance curves
-      water_removal_rate_l = UnitConversions.convert(water_removal_rate, "pint", "L")
-      energy_factor, water_removal_rate = dehumidifier_ief_to_ef_inputs(w_coeff, ef_coeff, integrated_energy_factor, water_removal_rate_l)
+      energy_factor, water_removal_rate = dehumidifier_ief_to_ef_inputs(w_coeff, ef_coeff, integrated_energy_factor, water_removal_rate)
```
jonwinkler: @yzhou601 It's not clear to me why the water removal rate is only adjusted when the energy_factor is nil. Shouldn't it be adjust all the time?

yzhou601: Oh it means we didn't get EF inputs directly. Since we require either EF or IEF, it's equivalent to say that we got IEF. In that case we'll need to adjust inputs to be values under EF test conditions.

jonwinkler: Oh, I see. We're continuing to support EF_80. So we're assuming that if IEF is passed in, water_removal_rate at 65 is also passed in.

jonwinkler:
@shorowit @yzhou601 I believe two things changed in the CFR, etc.:
1) The rated condition from 80°F to 65°F
2) The use of an IEF, which accounts for off-cycle and parasitic energy. 

Regarding 1), using the performance curves to make this adjustment to the E+ rated condition of 80°F makes sense. This is the approach Scott and I discussed awhile back. For example, EF_80 = EF_65 / perf_curve_value. 
Regarding 2), EnergyPlus does not have a great way to handle this. Yes, parasitic energy use can be modeled but not off-cycle power use. The current implementation might be the best option but we're using an integrated metric in an hourly simulation, which is not really how it should be used.

resources/hvac.rb
yzhou601: Yes you got my question, if a person is passing IEF, do you agree to assume all those other inputs are also under 65?

shorowit: I believe the answer is yes. See the glossary at https://www.energystar.gov/products/appliances/dehumidifiers/key_efficiency_criteria, where both IEF and Product Capacity reference the new test spec.

jonwinkler: Sorry if I missed that question. That might be a question for @shorowit . But in my opinion, it's a little confusing and will depend on where the user is getting information. Maybe there is a better way to distinguish the rated operating conditioned associated with the inputs.

yzhou601: No worries at all. True, currently these inputs are got from `extension` elements, if the element under `extension` is`IEF`, it might infer that the `capacity` required in its neighbor is also associated with IEF. But this could be a good suggestion to this hpxml proposal? https://github.com/hpxmlwg/hpxml/pull/208

yzhou601:
CI failed on component load sum check. I have investigated how `EnergyTransfer` meters report for dehumidifier, next step is to learn more about component output vars to know what causes discrepancy and how to address it (or skip it?).

measure.rb
```diff
@@ -3535,14 +3539,10 @@ def self.add_component_loads_output(runner, model)
...
-
-    liv_load_sensors = {}
-
-    liv_load_sensors[:htg] = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Heating:EnergyTransfer:Zone:#{@living_zone.name.to_s.upcase}")
-    liv_load_sensors[:htg].setName("htg_load_liv")
-
-    liv_load_sensors[:clg] = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Cooling:EnergyTransfer:Zone:#{@living_zone.name.to_s.upcase}")
-    liv_load_sensors[:clg].setName("clg_load_liv")
+    # Need to resolve conflict after new component load approach is merged
+    liv_load_sensors = OpenStudio::Model::EnergyManagementSystemSensor.new(model, "Zone Predicted Sensible Load to Setpoint Heat Transfer Rate")
```
yzhou601: It looks like using the new adjusted loads reported in EMS is more proper for dehumidifier, will test switching.

yzhou601: Switching to adjusted loads made discrepancy smaller. Switched for dehumidifier, and reverted these lines for other systems.

tests/hpxml_translator_test.rb
```diff
@@ -415,8 +424,8 @@ def _get_results(rundir, sim_time, workflow_time)
 
     sqlFile.close
 
-    assert_operator(compload_results["Heating - Residual"], :<, 0.2)
-    assert_operator(compload_results["Cooling - Residual"], :<, 0.2)
+    assert_operator(compload_results["Heating - Residual"], :<, 0.35)
```
yzhou601: Dehumidifier load discrepancy was reduced to:
Dehumidifier(EF) Heating: 1.685GJ vs 1.652GJ
Dehumidifier(EF) Cooling: 24.636GJ vs 24.356GJ
Dehumidifier(IEF) Heating: 1.687GJ vs 1.653GJ
Dehumidifier(IEF) Cooling: 24.686GJ vs 24.384GJ
Cooling is not passing 0.2GJ tolerance, but the discrepancy is relatively small. So I think the new approach is working correctly. (Previously, it broke at heating discrepancy being more than 0.2GJ, couldn't recall cooling)

yzhou601:
Interaction with latest master branch completed. Blocked by component load check (new approach) again. Need further investigation before moving forward to dehumidification load fraction.

SimulationOutputReport/measure.rb
```diff
@@ -451,6 +452,18 @@ def get_outputs(timeseries_frequency,
...
 
+    # Dehumidifier
+    outputs[:hpxml_dehumidifier_ids].each do |sys_id|
+      end_use = @end_uses[[FT::Elec, EUT::Dehumidifier]]
+      vars = get_all_var_keys(end_use.variable)
+      ep_output_names = @hvac_map[sys_id]
+      keys = ep_output_names.map(&:upcase)
+      end_use.annual_output_by_system[sys_id] = get_report_variable_data_annual_mbtu(keys, vars)
+      if include_timeseries_end_use_consumptions
+        end_use.timeseries_output_by_system[sys_id] = get_report_variable_data_timeseries(keys, vars, UnitConversions.convert(1.0, 'J', end_use.timeseries_units), 0, timeseries_frequency)
```
shorowit: Since there can only be a single dehumidifier, you don't need to use `annual_output_by_system` and `timeseries_output_by_system`. Just use `annual_output` and `timeseries_output`.

SimulationOutputReport/tests/output_report_test.rb
```diff
@@ -26,6 +26,7 @@ class SimulationOutputReportTest < MiniTest::Test
     "Electricity: Lighting Exterior (MBtu)",
     "Electricity: Mech Vent (MBtu)",
     "Electricity: Whole House Fan (MBtu)",
+    "Electricity: Dehumidifier (MBtu)",
```
shorowit: Eventually it should also be included in:
- TimeseriesColsEndUses: "Electricity: Dehumidifier"
- AnnualRows: "Component Load: Heating: Dehumidifier (MBtu)"
- AnnualRows: "Component Load: Cooling: Dehumidifier (MBtu)"
- TimeseriesColsComponentLoads: "Component Load: Heating: Dehumidifier"
- TimeseriesColsComponentLoads: "Component Load: Cooling: Dehumidifier"

yzhou601: In component load program, we're including dehumidifier into internal gains. Do you mean separating it out here? Or both places?

shorowit: Ah gotcha, I didn't realize that. In that case, only TimeseriesColsEndUses above needs to be addressed.

yzhou601: Do you mean the TimeseriesColsEndUses is also expected in hpxmls where no dehumidifier exists? Now dehumidifier column is added into annual results even if there's no dehumidifier (because it's populated as 0 when is nil), but not into timeseries. Should I change it to :
1. No humidifier hpxml: no dehumidifier annual, no dehumidifier timeseries;
2. Has humidifier hpxml: have dehumidifier annual, have dehumidifier timeseries;

OR:
1. No humidifier hpxml: have dehumidifier annual, have dehumidifier timeseries;
2. Has humidifier hpxml: have dehumidifier annual, have dehumidifier timeseries;

shorowit: The latter. I'm trying to keep the list of end uses static across all buildings, as well as keep the list identical between annual and timeseries. Note that "Electricity: Whole House Fan" is in both lists even though the test file has no whole house fan.

(I hope you meant dehumidifier above and not humidifier! 😆)

yzhou601:
Fraction of load served next steps:

- [ ] Do sensible load comparison of no dehumidifier + 100% dehumidifier + 50% dehumidifier with current approach (doubling EF w/o EMS)
- [ ] Decide whether implement EMS to adjust sensible heating load reported by dehumidifier object to meet the fraction
- [ ] Finish implementation and if EMS, compare again the step 1

yzhou601:
| Heating Load | Cooling Sensible Load
-- | -- | --
Dallas-NoDehumidifier | 1.822 | 28.227
Dallas-100%Dehumidifier | 1.187 | 40.307
Dallas-100%Dehumidifier-EFx2 | 1.374 | 36.670
Dallas-100%Dehumidifier-EFx2-EMS | 1.536 | 34.844
Average 0-100% | 1.505 | 34.267

The same test conditions:
Test building is in Dallas, TX. Humidity setpoint is 40%.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2807,6 +2808,25 @@ def self.add_ceiling_fans(runner, model, weather)
...
+    water_removal_rate = dehumidifier.capacity
+    energy_factor = dehumidifier.energy_factor
+    integrated_energy_factor = dehumidifier.integrated_energy_factor
+    humidity_setpoint = dehumidifier.rh_setpoint
+    fraction_served = dehumidifier.fraction_served
+
+    # Calculate air flow rate by assuming 2.75 cfm/pint/day (based on experimental test data)
+    air_flow_rate = 2.75 * water_removal_rate
+
+    HVAC.apply_dehumidifier(model, runner, energy_factor, integrated_energy_factor, water_removal_rate,
```
shorowit: Now that we have HPXML objects, how about just passing the `dehumidifier` object as the argument to `apply_dehumidifier`? Then we don't have all of these local variables doing hardly anything. The air_flow_rate calc can also go inside the `apply_dehumidifier` method. I('m going to start moving more of our code to this approach of passing objects when it makes sense.)

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -641,6 +642,16 @@ def self.run_validator(hpxml_doc)
         'RatedAnnualkWh | extension/AdjustedAnnualkWh' => one_or_more,
       },
 
+      # [Dehumidifier]
+      '/HPXML/Building/BuildingDetails/Appliances/Dehumidifier' => {
+        'SystemIdentifier' => one, # Required by HPXML schema
+        # TODO: "[Location='living space' or Location='basement']" => one,
```
shorowit: Let's remove the TODO.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3376,6 +3378,59 @@ def from_rexml(refrigerator)
...
+    end
+
+    def to_rexml(doc)
+      return if nil?
+
+      appliances = XMLHelper.create_elements_as_needed(doc, ['HPXML', 'Building', 'BuildingDetails', 'Appliances'])
+      dehumidifier = XMLHelper.add_element(appliances, 'Dehumidifier')
+      sys_id = XMLHelper.add_element(dehumidifier, 'SystemIdentifier')
+      XMLHelper.add_attribute(sys_id, 'id', @id)
+      XMLHelper.add_element(dehumidifier, 'Capacity', @capacity) unless @capacity.nil?
```
shorowit: Missing Float(...)

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3376,6 +3378,59 @@ def from_rexml(refrigerator)
...
+      XMLHelper.add_element(dehumidifier, 'EnergyFactor', Float(@energy_factor)) unless @energy_factor.nil?
+      XMLHelper.add_element(dehumidifier, 'IntegratedEnergyFactor', Float(@integrated_energy_factor)) unless @integrated_energy_factor.nil?
+      XMLHelper.add_element(dehumidifier, 'DehumidistatSetpoint', @rh_setpoint) unless @rh_setpoint.nil?
+      XMLHelper.add_element(dehumidifier, 'FractionDehumidificationLoadServed', @fraction_served) unless @fraction_served.nil?
```
shorowit: These two are missing Float(...) too

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3376,6 +3378,59 @@ def from_rexml(refrigerator)
     end
   end
 
+  class Dehumidifiers < BaseArrayElement
```
shorowit: Nice job on the new class! Looks great. 🎉

SimulationOutputReport/tests/output_report_test.rb
```diff
@@ -29,6 +29,7 @@ class SimulationOutputReportTest < MiniTest::Test
     'Electricity: Mech Vent (MBtu)',
     'Electricity: Whole House Fan (MBtu)',
     'Electricity: Refrigerator (MBtu)',
+    'Electricity: Dehumidifier (MBtu)',
```
shorowit: `AnnualRows` and `TimeseriesColsComponentLoads` should have new "Component Load: Heating: Dehumidifier" and "Component Load: Cooling: Dehumidifier" rows too, no? Even if the file has no dehumidifier, they should be present but have zero values.

tasks.rb
```diff
@@ -57,6 +57,9 @@ def create_hpxmls
     'invalid_files/water-heater-location-other.xml' => 'base.xml',
     'invalid_files/slab-zero-exposed-perimeter.xml' => 'base.xml',
 
+    'base-appliances-dehumidifier.xml' => 'base-location-dallas-tx.xml',
+    'base-appliances-dehumidifier-ief.xml' => 'base-appliances-dehumidifier.xml',
+    'base-appliances-dehumidifier-50percent.xml' => 'base-appliances-dehumidifier.xml',
```
shorowit: 👍

SimulationOutputReport/tests/output_report_test.rb
yzhou601: Yeah I understood why it was expected here, but currently, dehumidifier was included in internal gain subcategory for component loads program. 
It is not hard to separate if it kept being confusing. 😛

shorowit: Ah, I see. We can keep it in internal gains for now until someone asks for it, that's fine.

jonwinkler:
@shorowit I'll run the new test files today or tomorrow and take a look at the results.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2178,6 +2180,45 @@ def self.dehumidifier_ief_to_ef_inputs(w_coeff, ef_coeff, ief, water_removal_rat
...
+    dehumidifier_load_adj.setSchedule(model.alwaysOnDiscreteSchedule)
+
+    dehumidifier_load_adj_act = OpenStudio::Model::EnergyManagementSystemActuator.new(dehumidifier_load_adj, 'OtherEquipment', 'Power Level')
+    dehumidifier_load_adj_act.setName("#{zone_hvac.name} sens htg adj act")
+
+    # EMS program
+    program = OpenStudio::Model::EnergyManagementSystemProgram.new(model)
+    program.setName("#{zone_hvac.name} load adj program")
+    program.addLine("If #{dehumidifier_sens_htg.name} > 0")
+    program.addLine("  Set #{dehumidifier_load_adj_act.name} = - (#{dehumidifier_sens_htg.name} - #{dehumidifier_power.name}) * (1 - #{fraction_served})")
```
jonwinkler, line 2211: @yzhou601 I'm not sure I understand this equation. The dehumidifier power (read by the sensor) already accounts for the fraction of load served, correct? So is this double counting? Would it be simpler for the sensible load adjustment to equal fraction_served * m_dot_water_dehumidifier * h_fg? I'm happy to discuss.

yzhou601: Good catch! Yes the power portion is already adjusted, I will try your suggestions.

yzhou601: @jonwinkler Ah wait, the total sensible heating is also only accounting for the power adjusted, so it is:

    Qsensible,avg=Mdotwater,avg * hfg+(fractioned)Pdehumid,avg

The **Qsensible,avg - (fractioned)Pdehumid,avg** should yield the same results as **m * h_fg**.

And here in the equation, I'm adjusting the load by subtracting the fraction shouldn't be handled, so the (1 - fraction) is applied

jonwinkler: @yzhou601 Ahh, yes, you're right! This equation is correct.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2109,66 +2109,118 @@ def self.get_default_compressor_type(seer)
...
-      fail 'Invalid water removal rate value entered.'
-    end
-    if (energy_factor != Constants.Auto) && (energy_factor.to_f < 0)
-      fail 'Invalid energy factor value entered.'
-    end
-    if (air_flow_rate != Constants.Auto) && (air_flow_rate.to_f < 0)
-      fail 'Invalid air flow rate value entered.'
-    end
+    # Calculate air flow rate by assuming 2.75 cfm/pint/day (based on experimental test data)
+    air_flow_rate = 2.75 * water_removal_rate
```
shorowit, line 2117: Delete this? Seems to be overridden by the calculation below.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2109,66 +2109,118 @@ def self.get_default_compressor_type(seer)
...
+    water_removal_curve_value_ief = var_array_ief.zip(w_coeff).map { |var, coeff| var * coeff }.inject(0, :+)
+
+    # E+ inputs under EF test conditions
+    ef_input = ief / ef_curve_value_ief * curve_value_ef
+    water_removal_rate_input = water_removal_rate / water_removal_curve_value_ief * curve_value_ef
+
+    return ef_input, water_removal_rate_input
+  end
+
+  def self.adjust_dehumidifier_load_EMS(fraction_served, zone_hvac, model, living_space)
```
shorowit, line 2187: Please add comments describing why this EMS program is needed and what it does.

shorowit: Also, what do we think about skipping EMS when the dehumidifier serves 100% of the load?

shorowit:
@jonwinkler This is ready for final testing.

jonwinkler:
@shorowit @yzhou601 I'm running the three dehumidifier related test files. Two quick questions:

1. Should the dehumidifier energy use be a column in the results.csv?
2. I noticed in EPvalidator.rb that we require the dehumidifier capacity. So I'm assuming we're currently not support dehumidifier autosizing, correct?

shorowit:
1. Yes, it should. What file are you running and how are you running it?
2. I don't think we have a use case identified that needs dehumidifier autosizing, so it's not worth prioritizing right now. We can implement it as needed in the future.

jonwinkler:
@shorowit I was running all three dehumidifier test files (base-appliances-dehumidifier*.xml) from hpxml_translator_test. I also manually created a temporary file (base-appliances-dehumidifier-none), where I removed the dehumidifier inputs, for making energy comparisons.

shorowit:
Ah, right. The results.csv file is currently populated by the E+ End Uses By Subcategory output table. And I don't think it's possible to have the dehumidifier broken out as a new subcategory.

That said, we now have an OpenStudio reporting measure which does a more granular breakout than the E+ output table, so we should probably switch to using that. I'll put that on my list.

jonwinkler:
I see; no problem. I just noticed that results.csv had things like dishwasher, cooking range, etc. and thought that was a custom set of outputs we read in. I did notice that Electricity:Dehumidifier (MBtu) is in the results_annual.csv for each individual run.

Other than that, everything dehumidifier related looks good to me.

shorowit:
Looks good to go. Will merge once the CI finishes. Thanks @yzhou601 and @jonwinkler!

PR #227, opened 2019-11-14
Sort view factors

shorowit:
Sorts view factors to eliminate IDF diffs from one run to the next. CI should produce identical results.


shorowit:
CI results are unchanged.

PR #238, opened 2019-11-25
Foundation wall insulation

shorowit:
Prevent possibility of tiny slivers of rigid insulation on foundation walls.

### DIFF
--- a/measure.rb
+++ b/measure.rb
@@ -1637,6 +1637,9 @@ class OSModel
         drywall_thick_in = 0.0
         rigid_r = assembly_r - Material.Concrete(concrete_thick_in).rvalue - Material.GypsumWall(drywall_thick_in).rvalue - film_r
       end
+      if rigid_r < 0.1
+        rigid_r = 0.0
+      end
       if rigid_r < 0
         rigid_r = 0.0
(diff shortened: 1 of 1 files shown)


measure.rb
```diff
@@ -1637,6 +1637,9 @@ def self.add_foundation_wall(runner, model, spaces, fnd_wall_values, slab_frac,
         drywall_thick_in = 0.0
         rigid_r = assembly_r - Material.Concrete(concrete_thick_in).rvalue - Material.GypsumWall(drywall_thick_in).rvalue - film_r
       end
+      if rigid_r < 0.1
+        rigid_r = 0.0
+      end
```
yzhou601: Hi, @shorowit , this is a bit confusing. "If" here will break the logic of next check "if rigid_r < 0" because rigid_r will always be >=0, and match will always be true, right? (Though the codecov showed our tests didn't ever step into any of them.)

shorowit: Good catch. I fixed this [here](https://github.com/NREL/OpenStudio-HPXML/pull/253/commits/442380dc3e52e5ebb2196c387f37d21cbd14389f).

PR #244, opened 2019-12-05
Use standby loss to calculate indirect water heater ua

yzhou601:
Use [standby loss](https://www.ahridirectory.org/NewSearch?programId=28&searchTypeId=3), to calculate indirect water heater UA. 
If standby loss is missing, use an equation fit from this database to calculate assumed UA.

Reference: [Standby loss test standard](http://www.ahrinet.org/App_Content/ahri/files/standards%20pdfs/Indirect-Fired%20Water%20Heater%20Testing%20Standard03.pdf)


shorowit:
We'll also have to update https://github.com/NREL/OpenStudio-ERI/blob/master/workflow/energy_rating_index.rb#L905-L911 when we bring it over to ERI.

yzhou601:
CI results only shows differences on indirect system results.

resources/EPvalidator.rb
```diff
@@ -498,6 +498,7 @@ def self.run_validator(hpxml_doc)
         "RelatedHVACSystem" => one, # HeatingSystem (boiler)
         "TankVolume" => one,
         "WaterHeaterInsulation/Jacket/JacketRValue" => zero_or_one, # Capable to model tank wrap insulation
+        "extension/StandbyLoss" => zero_or_one,
```
shorowit: Since this is an extension element and not (yet) defined in the HPXML schema, can you add a comment that defines the units and where the input can be obtained from?

resources/waterheater.rb
```diff
@@ -575,15 +575,42 @@ def self.apply_heatpump(model, runner, space, weather, t_set, vol, ef,
...
-                          nbeds, boiler, boiler_plant_loop, boiler_fuel_type, dhw_map, sys_id, wh_type, jacket_r)
+                          nbeds, boiler, boiler_plant_loop, boiler_fuel_type, dhw_map, sys_id, wh_type, jacket_r, standby_loss)
     obj_name_indirect = Constants.ObjectNameWaterHeater
     convlim = model.getConvergenceLimits
     convlim.setMinimumPlantIterations(3) # add one more minimum plant iteration to achieve better energy balance across plant loops.
 
     if wh_type == "space-heating boiler with storage tank"
       tank_type = Constants.WaterHeaterTypeTank
+      # Actual tank volume = 95% nominal tank volume
+      act_vol = 0.95 * vol
```
shorowit: Please reuse the existing `calc_actual_tankvol` method here so we have consistency.

resources/waterheater.rb
```diff
@@ -832,13 +857,12 @@ def self.add_desuperheater(model, t_set, tank, desuperheater_clg_coil, wh_type,
       return [dsh_energy_output_var, dsh_load_output_var]
     else # need to test after switch
       # create a storage tank
-      storage_vol = 50 # FIXME: Input vs assumption?
+      storage_vol_actual = 50 * 0.95 # FIXME: Input vs assumption?
```
shorowit: Use the `calc_actual_tankvol` method here too.

resources/waterheater.rb
```diff
@@ -1199,11 +1231,19 @@ def self.create_new_schedule_manager(t_set, model, wh_type)
...
-    u, ua, eta_c = calc_tank_UA(act_vol, fuel, ef, re, cap, wh_type, cyc_derate, jacket_r, runner)
+    if ua.nil?
+      act_vol = calc_actual_tankvol(vol, fuel, wh_type)
+      u, ua, eta_c = calc_tank_UA(act_vol, fuel, ef, re, cap, wh_type, cyc_derate, jacket_r, runner)
+      new_heater.setHeaterThermalEfficiency(eta_c)
+      new_heater.setHeaterFuelType(HelperMethods.eplus_fuel_map(fuel))
+    else
+      # indirect water heater has standby loss for ua calculation
+      # actual volume is already calculated
+      act_vol = vol
```
shorowit: This seems very complicated that now the `vol` passed in is sometimes nominal and sometimes actual. It would be simpler to rename the `vol` variable in this method to `act_vol` and always pass in the actual volume (and move the `calc_actual_tankvol` calls outside this method).

resources/waterheater.rb
```diff
@@ -1215,8 +1255,6 @@ def self.create_new_heater(name, cap, fuel, vol, ef, re, jacket_r, t_set, space,
 
     new_heater.setHeaterMinimumCapacity(0.0)
     new_heater.setHeaterMaximumCapacity(UnitConversions.convert(cap, "kBtu/hr", "W"))
-    new_heater.setHeaterFuelType(HelperMethods.eplus_fuel_map(fuel))
-    new_heater.setHeaterThermalEfficiency(eta_c)
```
shorowit: I see what you're trying to do here, but it's a bit dangerous to only set the fuel type or thermal efficiency based on whether the UA is passed. That assumes that we'll never want to set the fuel/efficiency when UA is passed and will always want to set the fuel/efficiency when it's not passed. That is true of the current systems, but may not be true in the future. Rather, I would keep these setters here but only apply them if their value (`fuel` or `eta_c`) is not nil. Then someone could create a water heater in the future where they supply fuel, efficiency, and UA, for example.

resources/waterheater.rb
```diff
@@ -1199,11 +1231,19 @@ def self.create_new_schedule_manager(t_set, model, wh_type)
...
 
-  def self.create_new_heater(name, cap, fuel, vol, ef, re, jacket_r, t_set, space, oncycle_p, offcycle_p, ec_adj, wh_type, cyc_derate, nbeds, model, runner)
+  def self.create_new_heater(name, cap, fuel, vol, ef, re, jacket_r, t_set, space, oncycle_p, offcycle_p, ec_adj, wh_type, cyc_derate, nbeds, model, runner, ua)
     new_heater = OpenStudio::Model::WaterHeaterMixed.new(model)
     new_heater.setName(name)
-    act_vol = calc_actual_tankvol(vol, fuel, wh_type)
-    u, ua, eta_c = calc_tank_UA(act_vol, fuel, ef, re, cap, wh_type, cyc_derate, jacket_r, runner)
+    if ua.nil?
+      act_vol = calc_actual_tankvol(vol, fuel, wh_type)
+      u, ua, eta_c = calc_tank_UA(act_vol, fuel, ef, re, cap, wh_type, cyc_derate, jacket_r, runner)
```
shorowit: How about moving the `calc_tank_UA` call outside of this method, so that we always pass in UA instead? Then we can remove some arguments (ef, re, others?) that were only used to calculate UA. That should simplify everything.

resources/waterheater.rb
```diff
@@ -936,6 +960,17 @@ def self.calc_ef_from_uef(uef, type, fuel_type)
     return nil
   end
 
+  def self.calc_tank_areas(vol)
```
shorowit: Great!

resources/waterheater.rb
```diff
@@ -1155,6 +1164,29 @@ def self.calc_tank_UA(vol, fuel, ef, re, pow, wh_type, cyc_derate, jacket_r, run
     return u, ua, eta_c
   end
 
+  def self.apply_tank_jacket(jacket_r, ef, fuel, ua_pre, a_side)
```
shorowit: 👍 👍

resources/waterheater.rb
yzhou601: Thanks @shorowit .
Good idea of removing calc_ua and calc_vol outside! In this way, create_new_tanks should be more general for being called by different types of water heating systems. I did have a hard time trying to call create_new_heater function for indirect wh. Sorry for ending up with so many if/else statements. Will simplify it!

resources/waterheater.rb
```diff
@@ -588,24 +588,8 @@ def self.apply_indirect(model, runner, space, cap, vol, t_set, oncycle_p, offcyc
       tank_type = Constants.WaterHeaterTypeTank
       # Actual tank volume = 95% nominal tank volume
       act_vol = calc_storage_tank_actual_vol(vol, nil)
-      # Tank geometry
-      surface_area, a_side = calc_tank_areas(act_vol)
-
-      if standby_loss.nil? # Swiched to standby_loss equation fit from AHRI database
-        # calculate independent variable SurfaceArea/vol(physically linear to standby_loss/skin_u under test condition) to fit the linear equation from AHRI database
-        sqft_by_gal = surface_area / act_vol # sqft/gal
-        standby_loss = 2.9721 * sqft_by_gal - 0.4732 # linear equation assuming a constant u
```
jmaguire1: I'd put units in the comment as well

jmaguire1: If someone were to mess up the inputs and get a sqft_by_gal value of < 0.159 (.4732/2.9721) you'd get standby loss < 0, which would cause the simulation to crash. If you do the math out for how surface area is calculated, there's probably a critical volume where the sign changes. I'd check for that case and throw an instructive error message about bad inputs to help users identify why the run crashed.

yzhou601: Good point! I followed our area/volume calculation and solved this math, when act_v is over 787.8gal (that is, nominal volume over 787.8/0.95=829gal), the standby loss will be negative here. This value is obviously way too large than reality, I will add an error message if bad volume input is defined.

resources/waterheater.rb
```diff
@@ -64,7 +66,9 @@ def self.apply_tankless(model, runner, space, fuel_type, cap, ef,
     new_manager = create_new_schedule_manager(t_set, model, Constants.WaterHeaterTypeTankless)
     new_manager.addToNode(loop.supplyOutletNode)
 
-    new_heater = create_new_heater(Constants.ObjectNameWaterHeater, cap, fuel_type, 1, ef, 0, nil, t_set, space, oncycle_p, offcycle_p, ec_adj, Constants.WaterHeaterTypeTankless, cd, nbeds, model, runner)
+    act_vol = 1
```
shorowit: There are a lot of places where integers are used instead of doubles. Here, height = 48, t_amb = 70, etc. Some were introduced by you and some probably weren't. But can you look through the file and replace them all with doubles? 

Using integers is risky because of integer division where, for example, 5/2 = 2 in ruby. If it's 5.0 or 2.0, but then you get 2.5.

resources/waterheater.rb
```diff
@@ -575,15 +579,26 @@ def self.apply_heatpump(model, runner, space, weather, t_set, vol, ef,
...
     if wh_type == "space-heating boiler with storage tank"
       tank_type = Constants.WaterHeaterTypeTank
+      # Actual tank volume = 95% nominal tank volume
+      act_vol = calc_storage_tank_actual_vol(vol, nil)
+      a_side = calc_tank_areas(act_vol)[1]
+      ua = calc_indirect_tank_ua(act_vol, standby_loss)
+
+      # Tank jacket
+      # assume indirect water heater skin to be insulated with 2inch R5 (the same as electric)
+      ua = apply_tank_jacket(jacket_r, nil, Constants.FuelTypeElectric, ua, a_side)
```
shorowit: Can we move this method call inside of calc_indirect_tank_ua? Then it will be consistent with calc_tank_ua.

resources/waterheater.rb
```diff
@@ -1119,30 +1161,15 @@ def self.calc_tank_UA(vol, fuel, ef, re, pow, wh_type, cyc_derate, jacket_r, run
...
         ua = q_load * (1 / ef - 1) / ((t - t_env) * 24)
         eta_c = 1.0
-        skin_insulation_t = 2.0 # inch
       end
-      # water heater wrap calculation based on:
-      # Modeling Water Heat Wraps in BEopt DRAFT Technical Note
-      # Authors:  Ben Polly and Jay Burch (NREL)
-      u_pre_skin = 1 / (skin_insulation_t * skin_insulation_R + 1.0 / 1.3 + 1.0 / 52.8) # Btu/hr-ft^2-F = (1 / hout + kins / tins + t / hin)^-1
-      ua -= jacket_r / (1 / u_pre_skin + jacket_r) * u_pre_skin * a_side unless jacket_r.nil?
+      ua = apply_tank_jacket(jacket_r, ef, fuel, ua, a_side)
```
shorowit: Nice. This method is much cleaner now.

resources/waterheater.rb
yzhou601: If doing so, ERI eec calculation will call this function with wrap applied. It's probably harder to revert jacket changes in ERI.

shorowit: Wouldn't we be able to pass in jacket_r = 0 when calc_indirect_tank_ua is called from ERI?

yzhou601: Oh true. I will address this.

resources/waterheater.rb
```diff
@@ -936,6 +935,52 @@ def self.calc_ef_from_uef(uef, type, fuel_type)
...
+    height = 48.0 # inches
+    diameter = 24.0 * ((act_vol * 0.1337) / (height / 12.0 * pi))**0.5 # inches
+    a_top = pi * (diameter / 12.0)**(2.0) / 4.0 # sqft
+    a_side = pi * (diameter / 12.0) * (height / 12.0) # sqft
+    surface_area = 2.0 * a_top + a_side # sqft
+
+    return surface_area, a_side
+  end
+
+  def self.calc_indirect_tank_ua(act_vol, standby_loss, jacket_r, a_side, runner)
```
yzhou601: Sorry forgot to pass in runner before. Now locally tested bad standby samples, it gave error/warning as expected.

yzhou601: @shorowit Should I add a runner.nil check here to make sure energy_rating_index can call it without runner? Error-checking should be completed prior to ERI call.

shorowit: Since ERI shouldn't ever hit an error, I think it's ok to just pass nil for runner.

resources/waterheater.rb
```diff
@@ -936,6 +935,52 @@ def self.calc_ef_from_uef(uef, type, fuel_type)
...
+    rho = 8.216 # lb/gal
+    t_amb = 70.0 # F
+    t_tank_avg = 135.0 # F, Test begins at 137-138F stop at 133F
+
+    # UA calculation
+    q = standby_loss * cp * act_vol * rho # Btu/hr
+    ua = q / (t_tank_avg - t_amb) # Btu/hr-F
+
+    # jacket
+    ua = apply_tank_jacket(jacket_r, nil, nil, ua, a_side)
```
shorowit: How about moving the two standby_loss <= 0 error messages to a check on the final calculated UA that occurs here? Then you could have a simple generic error message like [here](https://github.com/NREL/OpenStudio-HPXML/blob/indirect_standby_loss/resources/waterheater.rb#L1170-L1172).

yzhou601: Sure I can. Just tried to be more descriptive for each specific error message. I'm ok to combine them.

resources/waterheater.rb
yzhou601: I was worried by the warning instead of error condition actually.

yzhou601:
@shorowit  For sure! @jmaguire1 Do you want to take a final look at this?

shorowit:
If @jmaguire1 has any final comments, let's address this in a new PR.

jmaguire1:
@shorowit nope I'm good, Yueyue already addressed everything I had. Thanks for merging!

PR #246, opened 2019-12-06
Push from HES

shorowit:


resources/hvac.rb
```diff
@@ -4237,6 +4237,11 @@ def self.get_unitary_system_air_loops(model, runner, thermal_zone)
       system = get_unitary_system_from_air_loop_hvac(air_loop)
       next if system.nil?
 
+      # skip evap cooler dummy unitary system
+      if system.additionalProperties.getFeatureAsString(Constants.SizingInfoHVACCoolType).is_initialized
+        next if system.additionalProperties.getFeatureAsString(Constants.SizingInfoHVACCoolType).get == Constants.ObjectNameEvaporativeCooler
+      end
```
shorowit: @yzhou601 I pushed the subtree from HEScore and this diff showed up. I'm thinking this code can be deleted, right?

yzhou601: Yes it shouldn't be here. Surprised to see it in HEScore, sorry about this.

PR #247, opened 2019-12-06
Desuperheater combinations

shorowit:
Only allows desuperheaters to be connected to conventional storage and tankless water heaters.


shorowit:
I checked that DSH + HPWH is correctly caught by the EPvalidator.rb change. (Prior to this change, the EnergyPlus simulation was successful.)

PR #254, opened 2019-12-17
Updates to support BuildResidentialHPXML

joseph-robertson:


BuildResidentialHPXML/measure.rb
```diff
@@ -1091,6 +1091,12 @@ def arguments(model)
     arg.setDefaultValue(24)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeBoolArgument("mech_vent_used_for_seasonal_cooling_load_reduction", true)
```
shorowit: This shouldn't be a measure argument.

For mechanical ventilation, you'll always want to set `used_for_whole_building_ventilation` to True. For whole house fans, if you decide to allow them in the measure, you'll always want to set `mech_vent_used_for_seasonal_cooling_load_reduction` to true.

joseph-robertson: I'm not totally sure what you mean. So you're saying we should have a separate argument where you can set either "mechanical ventilation" or "whole house fan"?

shorowit: You are already calling out a bunch of inputs as "Mechanical Ventilation: Foo". For those inputs, they should create a HPXML ventilation fan where `used_for_whole_building_ventilation` is true. Down the road, if you add "Whole House Fan: Foo" inputs, then you'd create a separate HPXML ventilation fan where `used_for_seasonal_cooling_load_reduction` is true.

joseph-robertson: Oh. So a "whole house fan" is not considered "mechanical ventilation" even though it falls into the MechanicalVentilation block of HPXML?

shorowit: I agree that HPXML is a little confusing in this regard. But yes, you could end up with two VentilationFan objects in HPXML's MechanicalVentilation section, one for what we call mechanical ventilation and one for a whole house fan.

joseph-robertson: Should we have a test file that includes both?

shorowit: I don't think it's necessary.

shorowit:
@joseph-robertson In build_residential_hpxml_test.rb, instead of running simulations, how about just calling the BuildResidentialHPXML measure and then comparing the resulting HPXML file to the HPXML file you're trying to mimic? This will cut down significantly on runtime and also help ensure that the the HPXML and OSW files are kept in sync.

shorowit:
And then there will be no need to store any additional artifacts.

joseph-robertson:
I think that's a good idea, and was originally my plan But I'm not sure how/what to compare across the hpxml files as they can look quite different. Attached are "base.xml" and "base-build-res-hpxml.xml".


[base.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/4181627/base.xml.txt)
[base-build-res-hpxml.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/4181630/base-build-res-hpxml.xml.txt)

shorowit:
I ran your file through the HPXML.collapse_enclosure() method, which combines like surfaces, and the resulting differences are far fewer. 

[base-build-res-hpxml-collapsed.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/4181767/base-build-res-hpxml-collapsed.xml.txt)

If we skip comparing certain elements (ClimateandRiskZones, Attics, Foundations, extension/UnitMultiplier, extension/Schedules*, etc.) and update a few geometry values, it seems like we are pretty close?

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("heating_system_flow_rate", true)
+    arg.setDisplayName("Heating System: Flow Rate")
+    arg.setDescription("The flow rate of the heating system.")
+    arg.setUnits("CFM")
+    arg.setDefaultValue(0)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("cooling_system_flow_rate", true)
```
shorowit: Remove these `heating_system_flow_rate` and `cooling_system_flow_rate` arguments.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    arg.setDisplayName("HVAC Distribution: Annual Cooling Distribution System Efficiency")
+    arg.setDescription("The annual cooling efficiency of the distribution system.")
+    arg.setDefaultValue(0.7)
+    args << arg
+
+    hvac_control_type_choices = OpenStudio::StringVector.new
+    hvac_control_type_choices << "manual thermostat"
+    hvac_control_type_choices << "programmable thermostat"
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument("hvac_control_type", hvac_control_type_choices, true)
```
shorowit: Remove this argument, doesn't affect the model.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("heating_setback_hours_per_week", true)
+    arg.setDisplayName("Heating Setback Hours per Week")
+    arg.setDescription("Specify the heating setback number of hours per week value.")
+    arg.setDefaultValue(0)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("heating_setback_start_hour", true)
+    arg.setDisplayName("Heating Setback Start Hour")
+    arg.setDescription("Specify the heating setback start hour value. 0 = midnight, 12 = noon")
+    arg.setDefaultValue(0)
```
shorowit: Use a better default value like 23 (11pm).

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("cooling_setup_hours_per_week", true)
+    arg.setDisplayName("Cooling Setup Hours per Week")
+    arg.setDescription("Specify the cooling setup number of hours per week value.")
+    arg.setDefaultValue(0)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("cooling_setup_start_hour", true)
+    arg.setDisplayName("Cooling Setup Start Hour")
+    arg.setDescription("Specify the cooling setup start hour value. 0 = midnight, 12 = noon")
+    arg.setDefaultValue(0)
```
shorowit: Use a better default value like 9 (9am).

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    duct_leakage_units_choices << "CFM25"
+    duct_leakage_units_choices << "Percent"
+
+    duct_location_choices = OpenStudio::StringVector.new
+    duct_location_choices << Constants.Auto
+    duct_location_choices << "living space"
+    duct_location_choices << "attic - vented"
+    duct_location_choices << "attic - unvented"
+    duct_location_choices << "crawlspace - vented"
+    duct_location_choices << "outside"
```
shorowit: There are choices missing here like "garage", "crawlspace - unvented", ...

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument("return_duct_location", duct_location_choices, true)
+    arg.setDisplayName("Return Duct: Location")
+    arg.setDescription("The location of the return duct.")
+    arg.setDefaultValue(Constants.Auto)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("supply_duct_surface_area", true)
+    arg.setDisplayName("Supply Duct: Surface Area")
+    arg.setDescription("The surface area of the first supply duct.")
```
shorowit: Remove "first"?

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_tested_flow_rate", true)
+    arg.setDisplayName("Mechanical Ventilation: Tested Flow Rate")
+    arg.setDescription("The tested flow rate of the mechanical ventilation.")
+    arg.setUnits("CFM")
+    arg.setDefaultValue(110)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_rated_flow_rate", true)
```
shorowit: Shouldn't have both tested and rated inputs; just ask for one.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_total_recovery_efficiency", true)
+    arg.setDisplayName("Mechanical Ventilation: Total Recovery Efficiency")
+    arg.setDescription("The total recovery efficiency of the mechanical ventilation.")
+    arg.setUnits("Frac")
+    arg.setDefaultValue(0.48)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_adjusted_total_recovery_efficiency", true)
```
shorowit: Shouldn't have both total and adjusted total; just ask for one.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_sensible_recovery_efficiency", true)
+    arg.setDisplayName("Mechanical Ventilation: Sensible Recovery Efficiency")
+    arg.setDescription("The sensible recovery efficiency of the mechanical ventilation.")
+    arg.setUnits("Frac")
+    arg.setDefaultValue(0.72)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("mech_vent_adjusted_sensible_recovery_efficiency", true)
```
shorowit: Shouldn't have both sensible and adjusted sensible; just ask for one.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+      arg.setDefaultValue(Constants.SizingAuto)
+      args << arg
+
+      arg = OpenStudio::Measure::OSArgument::makeStringArgument("water_heater_energy_factor_#{n}", true)
+      arg.setDisplayName("Water Heater #{n}: Rated Energy Factor")
+      arg.setDescription("Ratio of useful energy output from water heater #{n} to the total amount of energy delivered from the water heater.")
+      arg.setDefaultValue(Constants.Auto)
+      args << arg
+
+      arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("water_heater_uniform_energy_factor_#{n}", true)
```
shorowit: Shouldn't have both EF and UEF; just ask for one.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3573 @@
...
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument("hot_water_distribution_pipe_r_value", true)
+    arg.setDisplayName("Hot Water Distribution: Insulation Nominal R-Value")
+    arg.setUnits("h-ft^2-R/Btu")
+    arg.setDescription("Nominal R-value of the insulation on the DHW distribution system.")
+    arg.setDefaultValue(0.0)
+    args << arg
+
+    dwhr_facilities_connected_choices = OpenStudio::StringVector.new
+    dwhr_facilitie
```
shorowit: Either remove this input or add description that if it's provided, it overrides all the other Solar Thermal inputs (and then implement that logic).

shorowit: Shouldn't have both MEF and IMEF; just ask for one.

shorowit: Shouldn't have both EF and CEF; just ask for one.

shorowit: Shouldn't have both EF and kWh; just ask for one.

shorowit: Remove this argument.

BuildResidentialHPXML/measure.rb
joseph-robertson: So just enter `nil` for `heating_cfm` and `cooling_cfm`?

BuildResidentialHPXML/measure.rb
joseph-robertson: OK.

BuildResidentialHPXML/measure.rb
shorowit: They already default to nil, don't need to even pass them to the HPXML.add methods.

BuildResidentialHPXML/measure.rb
joseph-robertson: OK.

BuildResidentialHPXML/measure.rb
joseph-robertson: OK.

BuildResidentialHPXML/measure.rb
joseph-robertson: Any others that are missing? (I don't think I saw any test files that have ducts in these missing locations...)

BuildResidentialHPXML/measure.rb
joseph-robertson: OK.

BuildResidentialHPXML/measure.rb
shorowit: Stop looking at the test files and look at the [EPvalidator specification](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/EPvalidator.rb#L437) (here and for other elements with enumerations).

PR #255, opened 2019-12-17
Code cleanup & performance improvements

shorowit:
1. Replaces `runner.registerError` with `fail` and simplifies corresponding code (e.g., no longer need to return true/false, no longer need to pass around runner in most situations, etc).
2. Improves performance (of FT and writing IDF) for schedules by extending ScheduleRules for consecutive months with identical values
3. Prevents creating an interior shading schedule for each window/skylight.
4. Significantly streamlines the code that creates heating/cooling setpoint schedules.

No CI diffs are expected.


shorowit:
No CI diffs. Good to go.

PR #256, opened 2019-12-23
Natural ventilation improvements

shorowit:
Greatly simplifies code. Adjusts controls to reduce potential for increased heating due to natural ventilation operation.


shorowit:
Incorporating in https://github.com/NREL/OpenStudio-HPXML/pull/258 instead.

PR #257, opened 2019-12-23
EPvalidator xpath bugfix

yzhou601:
Queries with "|" ("or" logic) are not checking counts of target elements in EP validator, instead, they're mostly checking their parent elements. This PR is aimed to fix that issue. Shouldn't have any CI diff.


yzhou601:
Changes tested to be good, EPvalidater is correctly checking counts of target elements and giving error for incorrect counts. 
I kept those xpaths of "choice" elements untouched, they're functionally good.
Ready for review.

resources/EPvalidator.rb
```diff
@@ -49,8 +49,7 @@ def self.run_validator(hpxml_doc)
 
         "/HPXML/Building/BuildingDetails/ClimateandRiskZones/WeatherStation" => one, # See [WeatherStation]
 
-        "/HPXML/Building/BuildingDetails/Enclosure/AirInfiltration[AirInfiltrationMeasurement[HousePressure=50]/BuildingAirLeakage[UnitofMeasure='ACH' or UnitofMeasure='CFM']/AirLeakage | AirInfiltrationMeasurement/extension/ConstantACHnatural]" => one, # ACH50, CFM50, or constant nACH; see [AirInfiltration]
-        "/HPXML/Building/BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement/InfiltrationVolume" => zero_or_one, # Assumes InfiltrationVolume = ConditionedVolume if not provided
+        "/HPXML/Building/BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement" => one, # see [AirInfiltration]
```
shorowit: This change has a subtle difference that is importance. Previously, you could provide an HPXML file that had additional `AirInfiltrationMeasurement` elements beyond what was needed for simulation and they would be ignored. With this change, you would not be able to have extra information stored in the HPXML file since only one `AirInfiltrationMeasurement` element is allowed.

So I think the top line needs to be left where it is (and apply your xpath bugfix there).

resources/EPvalidator.rb
```diff
@@ -127,8 +128,8 @@ def self.run_validator(hpxml_doc)
       },
 
       ## [VentedAttic]
-      "/HPXML/Building/BuildingDetails/Enclosure/Roofs/Roof[InteriorAdjacentTo='attic - vented']" => {
-        "../../Attics/Attic[AtticType/Attic[Vented='true']]/VentilationRate[[UnitofMeasure='SLA']/Value | extension/ConstantACHnatural]" => zero_or_one,
+      "/HPXML/Building/BuildingDetails/Enclosure/Roofs/Roof[InteriorAdjacentTo='attic - vented']/../../Attics/Attic[AtticType/Attic[Vented='true']]" => {
+        "VentilationRate[UnitofMeasure='SLA']/Value | extension/ConstantACHnatural" => zero_or_one,
```
shorowit: There's a subtle difference here too. Previously, if there is a roof adjacent to vented attic, we checked for a ventilation rate on an Attic element (of vented attic). With your change, the check only occurs if the Attic element exists too. So if an HPXML file has the roof element but no corresponding Attic element, the attic ventilation check no longer occurs.

I think the reference to the Attic needs to remain where it was and the xpath bugfix needs to be applied differently.

resources/EPvalidator.rb
yzhou601: Isn't only one AirInfultrationMeasurement allowed in HPXML?

resources/EPvalidator.rb
yzhou601: I think the syntax is functionally the same, if there's no attic element, the attic ventilation won't be checked in either case because attic ventilation is under `Attic` element. My understanding is that the difference is, my change is stopped at searching parent element, but the previous code is stopped at searching sub element. If this hash has checks other than attic elements, it does have difference, as we want to execute those checks even when attic doesn't exist, but not in this case. 

The reason why I split these lines into one-layer relationship is because I want the format of "parent element" => {"sub element" => count} be only checking specific parent element structure, preventing checking the whole tree from root or from much higher level (as we know, it will cause issues when multiple tree branches are specified). It also looks more clear. If we still want it to be executed one by one from roof -> attic -> attic ventilation, it might require more hashes like: 

    "/HPXML/Building/BuildingDetails/Enclosure/Roofs/Roof[InteriorAdjacentTo='attic - vented']" => {
        "../../Attics/Attic[AtticType/Attic[Vented='true']] => zero_or_more,
    },

    "/HPXML/Building/BuildingDetails/Enclosure/Attics/Attic[AtticType/Attic[Vented='true']]" => {
    "VentilationRate[UnitofMeasure='SLA']/Value | extension/ConstantACHnatural" => zero_or_one, 
    },

Another bug here is that the `extension` check was under VentilationRate before, but only HPXML schema only allows `Attic` having `extension` element, so I removed the check under Attic. Let me know your thought.

resources/EPvalidator.rb
yzhou601: Sorry the previous comment was deleted because I was thinking about AirInfiltration, which is only allowed to be 1 max occurance. 
You're right, this subtle difference did exist. If I fixed it in the top line, then the check will be across all measurement elements, probably be fine if the information in additional elements is not relevant to our required inputs.

measure.rb
```diff
@@ -2981,6 +2981,8 @@ def self.add_airflow(runner, model, building, weather, spaces)
       building.elements.each("BuildingDetails/Enclosure/Attics/Attic[AtticType/Attic[Vented='true']]") do |vented_attic|
         vented_attic_values = HPXML.get_attic_values(attic: vented_attic,
                                                      select: [:vented_attic_sla, :vented_attic_constant_ach])
+        next if vented_attic_values[:vented_attic_sla].nil? and vented_attic_values[:vented_attic_sla].nil? # skip additional attic elements
```
yzhou601: A potential bug here when the last element in the loop is an additional element, then the useful information in another element might be ignored because `vented_attic_sla` and `vented_attic_const_ach` are eventually assigned to be nil.

measure.rb
```diff
@@ -4564,11 +4566,10 @@ def self.assign_space_to_subsurface(surface, subsurface_id, wall_idref, building
 
   def self.get_infiltration_volume(building)
     infilvolume = nil
-    building.elements.each("BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement") do |air_infiltration_measurement|
-      air_infiltration_measurement_values = HPXML.get_air_infiltration_measurement_values(air_infiltration_measurement: air_infiltration_measurement,
-                                                                                          select: [:infiltration_volume])
-      infilvolume = air_infiltration_measurement_values[:infiltration_volume] unless air_infiltration_measurement_values[:infiltration_volume].nil?
-    end
+    air_infiltration_measurement = building.elements["BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement[HousePressure=50 and BuildingAirLeakage[UnitofMeasure='ACH' or UnitofMeasure='CFM']/AirLeakage | extension/ConstantACHnatural]"] # only one measurement element has useful inputs
```
yzhou601: Only one element is desired, removed the loop.

yzhou601: Syntax error, need to fix

measure.rb
shorowit:
```suggestion
        next if vented_attic_values[:vented_attic_sla].nil? and vented_attic_values[:vented_attic_constant_ach].nil? # skip additional attic elements
```

measure.rb
```diff
@@ -4564,11 +4566,10 @@ def self.assign_space_to_subsurface(surface, subsurface_id, wall_idref, building
...
-      air_infiltration_measurement_values = HPXML.get_air_infiltration_measurement_values(air_infiltration_measurement: air_infiltration_measurement,
-                                                                                          select: [:infiltration_volume])
-      infilvolume = air_infiltration_measurement_values[:infiltration_volume] unless air_infiltration_measurement_values[:infiltration_volume].nil?
-    end
+    air_infiltration_measurement = building.elements["BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement[(HousePressure=50 and BuildingAirLeakage[UnitofMeasure='ACH' or UnitofMeasure='CFM']/AirLeakage) | extension/ConstantACHnatural]"] # only one measurement element has useful inputs
+    air_infiltration_measurement_values = HPXML.get_air_infiltration_measurement_values(air_infiltration_measurement: air_infiltration_measurement,
+                                                                                        select: [:infiltration_volume])
+    infilvolume = air_infiltration_measurement_values[:infiltration_volume] unless air_infiltration_measurement_values[:infiltration_volume].nil?
```
shorowit: There doesn't seem to be any point in having the `unless air_infiltration_measurement_values[:infiltration_volume].nil?` anymore, right? If it is nil, then infilvolume is assigned nil, which is fine.

```suggestion
    infilvolume = air_infiltration_measurement_values[:infiltration_volume]
```

resources/EPvalidator.rb
```diff
@@ -104,12 +103,13 @@ def self.run_validator(hpxml_doc)
       "/HPXML/Building/BuildingDetails/ClimateandRiskZones/WeatherStation" => {
         "SystemIdentifier" => one, # Required by HPXML schema
         "Name" => one, # Required by HPXML schema
-        "[WMO | extension/EPWFileName]" => one_or_more, # Reference weather/data.csv for the list of acceptable WMO station numbers
+        "WMO | extension/EPWFileName" => one_or_more, # Reference weather/data.csv for the list of acceptable WMO station numbers
```
shorowit: I can't figure out why this is `one_or_more`. I know this isn't your change, but let's try changing it to `one` and see if anything complains.

```suggestion
        "WMO | extension/EPWFileName" => one, # Reference weather/data.csv for the list of acceptable WMO station numbers
```

resources/EPvalidator.rb
```diff
@@ -204,8 +204,8 @@ def self.run_validator(hpxml_doc)
         "Thickness" => one, # Use zero for dirt floor
         "ExposedPerimeter" => one,
         "PerimeterInsulationDepth" => one,
-        "[UnderSlabInsulationWidth | [UnderSlabInsulationSpansEntireSlab='true']]" => one,
-        "[DepthBelowGrade | [InteriorAdjacentTo!='living space' and InteriorAdjacentTo!='garage']]" => one_or_more, # DepthBelowGrade only required when InteriorAdjacentTo is 'living space' or 'garage'
+        "UnderSlabInsulationWidth | UnderSlabInsulationSpansEntireSlab='true'" => one,
```
shorowit: This doesn't look right. I tried a HPXML that has
```
            <UnderSlabInsulationWidth>2.0</UnderSlabInsulationWidth>
            <UnderSlabInsulationSpansEntireSlab>false</UnderSlabInsulationSpansEntireSlab>
```
and the EPvalidator incorrectly threw an error
`Error: C:/GitHub/OpenStudio-HPXML/tests/base-foundation-slab.xml: Expected [1] element(s) but found 2 element(s) for xpath: /HPXML/Building/BuildingDetails/Enclosure/Slabs/Slab/UnderSlabInsulationWidth | UnderSlabInsulationSpansEntireSlab='true'`

I think it needs to be
```suggestion
        "UnderSlabInsulationWidth | [UnderSlabInsulationSpansEntireSlab='true']" => one,
```

yzhou601: True, I thought they're the same. Will fix it.

measure.rb
yzhou601: Sorry for this silly error 🤦‍♀ Good catch!

shorowit:
Thanks @yzhou601!

PR #263, opened 2020-01-02
Fixes running simulations when there is no weather .cache file

### LINKED ISSUE #261: WeatherProcess.new fails for new weather files
When adding a new epw weather file to the weather directory (which doesn't yet have a .cache file), this line fails: https://github.com/NREL/OpenStudio-HPXML/blob/master/measure.rb#L164.

shorowit:
Fixes #261.

Fixes running simulations when there is no weather .cache file. Removes obsolete code. Adds test.


shorowit:
@joseph-robertson Does this fix it for you?

PR #266, opened 2020-01-03
Improved EPvalidator to be able to check referenced nodes

yzhou601:
Allowed EPvalidator to check referenced node, eg. only check referenced distribution system element for heating/cooling system.

### DIFF
--- a/measure.rb
+++ b/measure.rb
@@ -112,9 +112,9 @@ class HPXMLTranslator < OpenStudio::Measure::ModelMeasure
 
     hpxml_doc = XMLHelper.parse_file(hpxml_path)
 
-      if not validate_hpxml(runner, hpxml_path, hpxml_doc, schemas_dir)
-        return false
-      end
+    if not validate_hpxml(runner, hpxml_path, hpxml_doc, schemas_dir)
+      return false
+    end
 
     begin
       # Weather file
--- a/measure.rb
+++ b/measure.rb
@@ -2299,8 +2299,6 @@ class OSModel
       end
       @total_frac_remaining_cool_load_served -= load_frac
 
-      check_distribution_system(building, cooling_system_values)
-
       sys_id = cooling_system_values[:id]
       @hvac_map[sys_id] = []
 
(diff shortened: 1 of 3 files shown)


measure.rb
```diff
@@ -2835,25 +2829,6 @@ def self.add_ceiling_fans(runner, model, building, weather)
                             @cfa, @living_space)
   end
 
-  def self.check_distribution_system(building, system_values)
```
yzhou601: New EPvalidator will check the id reference as well as distribution system type, this test can be removed, right?

shorowit: Seems like it. 🚀

yzhou601:
Didn't see available syntax to directly specify child XPath with parent node in its predicates( `current()` looks not supported yet). 
Is the current approach of using XPath variables acceptable?

resources/EPvalidator.rb
```diff
@@ -694,12 +694,10 @@ def self.run_validator(hpxml_doc)
...
-            # check referenced node by using parent's attribute(idref) value. Parameter name follows $ChildElementName_AttributeName
-            if child.include? "$"
-              var_name = child.split("$")[1].split(/\W+/)[0]
-              element_name = var_name.split("_")[0]
-              attribute_name = var_name.split("_")[1]
-              var[var_name] = REXML::XPath.first(parent_element, element_name).attributes[attribute_name] unless REXML::XPath.first(parent_element, element_name).nil?
+            # check referenced HVACDistribution
+            if child.include? "$dist_idref" and not REXML::XPath.first(parent_element, "DistributionSystem").nil?
+              # give id value to variable $dist_idref
+              var["dist_idref"] = REXML::XPath.first(parent_element, "DistributionSystem").attributes["idref"]
```
yzhou601: One last simplification before reverting.

shorowit:
I think it's awesome that you found a xpath-based solution, but have to admit that I find it to be a bit opaque/complex. I'm going to close this in lieu of your other, simpler approach.

yzhou601:
> I think it's awesome that you found a xpath-based solution, but have to admit that I find it to be a bit opaque/complex. I'm going to close this in lieu of your other, simpler approach.

Sounds good. Understood the preference of not taking Xpath adventures if we can.

PR #267, opened 2020-01-03
Purge constants

shorowit:
Code cleanup that replaces many constants (e.g., Constants.FuelTypeElectric) with HPXML enumeration values (e.g., "electricity"). Should be no CI diffs.

FYI @joseph-robertson


shorowit:
No CI diffs, good to go.

PR #271, opened 2020-01-06
Add distribution system type check.

yzhou601:
## Pull Request Description

Alternate of #266 for validating distribution system type. Close one of these two.

## Checklist

Not all may apply:

- [x] `rake update_measures` has been run


measure.rb
```diff
@@ -2805,21 +2806,33 @@ def self.add_ceiling_fans(runner, model, building, weather)
...
-    dist_id = system_values[:distribution_system_idref]
+  def self.check_distribution_system(building, dist_id, system_id, system_type)
     return if dist_id.nil?
 
+    hvac_distribution_type_map = { "Furnace" => ["AirDistribution", "DSE"],
+                                   "Boiler" => ["HydronicDistribution", "DSE"],
+                                   "central air conditioner" => ["AirDistribution", "DSE"],
+                                   "air-to-air" => ["AirDistribution", "DSE"],
+                                   "mini-split" => ["AirDistribution", "DSE"],
+                                   "ground-to-air" => ["AirDistribution", "DSE"] }
```
shorowit: Hmm, why is evaporative cooler not in this list? It is similar to mini-split in that it can be optionally ducted.

measure.rb
```diff
@@ -2805,21 +2806,33 @@ def self.add_ceiling_fans(runner, model, building, weather)
...
     # Get attached distribution system
     found_attached_dist = false
+    type_match = true
     building.elements.each("BuildingDetails/Systems/HVAC/HVACDistribution") do |dist|
       hvac_distribution_values = HPXML.get_hvac_distribution_values(hvac_distribution: dist)
       next if dist_id != hvac_distribution_values[:id]
 
       found_attached_dist = true
+      if not hvac_distribution_type_map[system_type].nil?
+        type_match = hvac_distribution_type_map[system_type].include? hvac_distribution_values[:distribution_system_type]
```
shorowit: Why do we need to check for nil here? What systems aren't we handling? My inclination is that we should throw an error if we don't have a system included in the map that can be ducted, but maybe I'm missing a nuance here.

measure.rb
```diff
@@ -2805,21 +2806,33 @@ def self.add_ceiling_fans(runner, model, building, weather)
...
+      end
     end
 
     if not found_attached_dist
-      fail "Attached HVAC distribution system '#{dist_id}' cannot be found for HVAC system '#{system_values[:id]}'."
+      fail "Attached HVAC distribution system '#{dist_id}' cannot be found for HVAC system '#{system_id}'."
+    elsif not type_match
+      fail "Incorrect HVAC distribution system type for HVAC type: '#{system_type}'. Should be one of: #{hvac_distribution_type_map[system_type]}"
```
shorowit: Nice error message.

measure.rb
yzhou601: Yeah the nil check is also for skipping evap cooler. Evap cooler didn't check system type in EP-validator so I didn't address it here. I will add it to both places.

measure.rb
yzhou601: No need to check nil here. Removed. 
Distribution counts are restricted in EPvalidator, so no possibility to have an illegal ducted system specified in HPXML.

PR #277, opened 2020-01-10
Bugfix basement surface construction

yzhou601:
## Pull Request Description

There's a bug in changing solar/visible absorptance properties of basement surfaces. 
If there's only one material layer in the basement surface construction, it didn't check material/construction share before splitting the single layer. Therefore, it might cause errors to other surfaces when shared material layer was splitted into half. It should be moved after share check.

## Checklist

Not all may apply:


yzhou601:
Defect file is now working correctly.

yzhou601:
My conclusion is that E+ kiva is not allowing any offset. If there's more than one custom block layer, for x position inputs of each one, its's required to include foundation wall thickness + inner layer thicknesses without any gap. Our workflow is satisfying it so it should be fine for us after this bugfix.

PR #282, opened 2020-01-15
Improved calculation of component loads

shorowit:
## Pull Request Description

The current approach to calculating heating/cooling component loads was to try to sum components heat transfers for any hour with a heating (or cooling) load. However, any hours with offsetting component heat transfers (i.e., net sum of zero, and thus, no heating/cooling loads) were not captured. Furthermore, the natural ventilation component load would be calculated as essentially zero since its operation would typically eliminate the cooling load for the hour.

A new approach was devised where every hour of the year is assigned to either heating or cooling. The algorithm can be described in pseudo code as follows:

```
Loop through each hour:
    If hour has heating load:
        Assign hour to heating season
    Else if hour has cooling load or natural ventilation or whole house fan:
       Assign hour to cooling season
    Else if hour has no load:
       Assign hour to heating/cooling season based on heating/cooling season definition (e.g., used for interior shading)
```

This PR also fixes a few bugs where some energy transfers were not captured.

## Checklist

Not all may apply:


tests/hpxml_translator_test.rb
```diff
@@ -415,8 +415,8 @@ def _get_results(rundir, sim_time, workflow_time)
 
     sqlFile.close
 
-    assert_operator(compload_results["Heating - Residual"], :<, 0.2)
-    assert_operator(compload_results["Cooling - Residual"], :<, 0.2)
+    assert_operator(compload_results["Heating - Residual"], :<, 0.35)
+    assert_operator(compload_results["Cooling - Residual"], :<, 0.35)
```
shorowit: @yzhou601 😱

shorowit:
Need to double-check a building with a setpoint change.

tests/hpxml_translator_test.rb
shorowit: It appears that the values are off by more than before because the residual value checks above are absolute values. In reality, the heating residuals are negative and cooling residuals are positive, with the total residuals very close to 0 for most files (but a few higher, max of 0.23 GJ). 

The "problem" is likely due to choosing the "wrong" season for an hour with no heating/cooling load. If a month in the Fall is assigned to heating and a month in the Spring is assigned to cooling, and these months see opposite trends to the living space temperature, their net sum loads will be non-zero. I'm going to try tweaking the heating/cooling seasons a bit but the issue will never be eliminated.

yzhou601: True, we're assigning modes with definition for that situation, along with signs(+ for heating and - for cooling). If a wrong sign is picked, they won't be offset during runperiod. Just one thought: why can't we create another bin/mode for no heating/cooling load situation (eg. float_mode)? It can distinguish the general heat transfer happens all the time from load seen by hvac, and help go through the residual check. Though it might output a new set of component load variables with prefix "float_" or something.

measure.rb
```diff
@@ -3922,73 +3914,46 @@ def self.add_component_loads_output(runner, model)
...
-      program.addLine("Set #{prev_hr_setpoint_vars[mode].name} = #{setpoint_sensors[mode].name}")
     end
 
+    # EMS program: Total loads
+    program.addLine("Set loads_htg_tot = 0")
+    program.addLine("Set loads_clg_tot = 0")
+    program.addLine("If #{liv_load_sensors[:htg].name} > 0")
+    program.addLine("  Set loads_htg_tot = #{tot_load_sensors[:htg].name} - #{tot_load_sensors[:clg].name}")
+    program.addLine("ElseIf #{liv_load_sensors[:clg].name} > 0")
+    program.addLine("  Set loads_clg_tot = #{tot_load_sensors[:clg].name} - #{tot_load_sensors[:htg].name}")
```
yzhou601: Is it possible to have non-zero Heating:EnergyTransfer and non-zero Cooling:EnergyTransfer simultaneously in EMS (operates in time step resolution)?

shorowit: Yes, there can be non-zero Heating:EnergyTransfer in one zone and non-zero Cooling:EnergyTransfer in another zone. For example, this can happen if the ducts are in the attic and the heating system is running, but the attic zone is actually warmer than the living zone (heating setpoint). In this case, Heating:EnergyTransfer:Zone:Living and Cooling:EnergyTransfer:Zone:PlenumZone are both non-zero. 

After making the above code change, a couple test files show much better agreement between total loads and sum of component loads.

yzhou601: Thanks for clarification. I thought E+ might also have offset cooling and heating between zones, like what it did of summing up all pieces of hvac equipment no matter if it's cooling only or heating only😅. Good to see better agreement!

PR #285, opened 2020-01-18
Improved calculation of component loads, alternative

shorowit:
## Pull Request Description

Alternative approach suggested by @yzhou601. We'll need to choose either this PR or the other one.

The current approach to calculating heating/cooling component loads was to try to sum components heat transfers for any hour with a heating (or cooling) load. However, any hours with offsetting component heat transfers (i.e., net sum of zero, and thus, no heating/cooling loads) were not captured. Furthermore, the natural ventilation component load would be calculated as essentially zero since its operation would typically eliminate the cooling load for the hour.

A new approach was devised where every hour of the year is assigned to either heating or cooling. The algorithm can be described in pseudo code as follows:

```
Loop through each hour:
    If living temperature < average setpoint temperature:
        Assign hour to heating season
    Else
       Assign hour to cooling season
```

## Checklist

Not all may apply:


shorowit:
Final decision: let's go with the original approach (using seasons).

PR #298, opened 2020-02-12
ViewFactor approach improvement

yzhou601:
## Pull Request Description

To improve view factor approach implemented for single conditioned space workflow.

## Checklist

Not all may apply:


yzhou601:
One error check added here. Fixed a "subsurface in basement" issue, not captured in test cases. So no CI diff here.

shorowit:
Were you going to add a check that the .err file doesn't have warnings like "Warning ** FixViewFactors: View factors not complete"? Similar to [this check](https://github.com/NREL/OpenStudio-HPXML/blob/master/workflow/tests/hpxml_translator_test.rb#L617-L623).

yzhou601:
> Were you going to add a check that the .err file doesn't have warnings like "Warning ** FixViewFactors: View factors not complete"? Similar to [this check](https://github.com/NREL/OpenStudio-HPXML/blob/master/workflow/tests/hpxml_translator_test.rb#L617-L623).

Sure I can. I thought we came to the conclusion to add error check in workflow. 
I will also add that one for testing files.

shorowit:
Thanks. Seems best to have both!

shorowit:
Confirmed that the basement window test file produces a change. All other files are unchanged.

PR #300, opened 2020-02-14
Allow small tolerance of hvac fraction sum

yzhou601:
## Pull Request Description

Currently, EPvalidator allows heating/cooling system load fractions to be sum to <=1.01, (0.01 tolerance), but EnergyPlus simulation will break when 1.01 occurs, thus cause a bug. This PR aims to fix this.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -2470,15 +2482,27 @@ def self.add_heat_pump(runner, model, building, weather)
...
       load_frac_cool = heat_pump_values[:fraction_cool_load_served]
       if @total_frac_remaining_cool_load_served > 0
-        sequential_load_frac_cool = load_frac_cool / @total_frac_remaining_cool_load_served # Fraction of remaining load served by this system
+        if (load_frac_cool - @total_frac_remaining_cool_load_served).abs <= 0.010001
+          # Last equipment to handle all the remaining load (within 0.01 tolerance)
+          load_frac_cool = @total_frac_remaining_cool_load_served
+          sequential_load_frac_cool = 1.0 # Fraction of remaining load served by this system
+        else
+          sequential_load_frac_cool = load_frac_cool / @total_frac_remaining_cool_load_served # Fraction of remaining load served by this system
+        end
```
shorowit: Seems like similar code added to 4 different places. Can we create a method that is called by all 4 places?

yzhou601: Sounds like a good idea

shorowit:
@yzhou601 Can you see if similar code is needed for the DHW load fractions?

yzhou601:
I inspected dhw fractions when I was fixing this issue. It looks like for dhw, it follows a manner that we're assigning different portions of hot water draws(needs) to each system before E+ takes care of it, so in E+ it won't break simulation nor report any issue as hvac fraction does. And I double checked EPvalidator, the dhw load sum check is in its place, so I assume dhw is good in shape.

shorowit:
> I inspected dhw fractions when I was fixing this issue. It looks like for dhw, it follows a manner that we're assigning different portions of hot water draws(needs) to each system before E+ takes care of it, so in E+ it won't break simulation nor report any issue as hvac fraction does. And I double checked EPvalidator, the dhw load sum check is in its place, so I assume dhw is good in shape.

Ah right, we don't pass the DHW fractions through to E+ inputs, so it can't possibly have the same problem. Thanks!

PR #301, opened 2020-02-15
Add check for FoundationWall and Slab attachment

### LINKED ISSUE #297: HPXML file w/ invalid foundation configuration
If an HPXML file has foundation walls adjacent to basement/crawlspace but no corresponding slab, the file still runs (but the model is weird). We need to throw an error for this case and add a test.

yzhou601:
Fix #297 

## Checklist

Not all may apply:


yzhou601:
No CI diff, ready for review

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1447,6 +1450,9 @@ def self.add_foundation_walls_slabs(runner, model, building, spaces)
                                 drywall_thick_in, film_r, mat_ext_finish)
       end
     end
+    if fndwall_slab_size < fndwall_size
+      fail "There existing foundation wall(s) not attached to any slab. Please double check."
```
shorowit: Can we give a more specific error message? Something like: "Foundation wall 'foo' is adjacent to 'basement - unconditioned' but no corresponding slab was found adjacent to 'basement - unconditioned'".

yzhou601:
@shorowit changed as suggested. One thing need to note: the foundation walls we can get here might already be collapsed. That means the foundation wall ids listed in the error message might be incomplete.

shorowit:
> @shorowit changed as suggested. One thing need to note: the foundation walls we can get here are already collapsed. That means the foundation wall ids listed in the error message might be incomplete.

True. Ideally we would have these checks in EPvalidator, but the xpath expressions won't be so user friendly. (Maybe we can reconsider if/when we switch to schematron.) I'm ok with the current limitation; they will at least get 1 error message.

shorowit:
Will merge once the CI comes back ✔️ from my last change.

PR #310, opened 2020-03-02
Multifamily spaces

yzhou601:
## Pull Request Description

1. Adds new multifamily space types for `ExteriorAdjacentTo` for surfaces:
- "other heated space"
- "other multifamily buffer space"
- "other non-freezing space"
- ("other housing unit" is already supported)
2. Allows ducts and water heaters to be located in these new space types.
3. Adds `Location` input for CookingRange and Dishwasher
4. Allows all appliances to be located in "other", in which internal gains are neglected.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -136,7 +136,7 @@ def self.run_validator(hpxml_doc)
       # [Wall]
       "/HPXML/Building/BuildingDetails/Enclosure/Walls/Wall" => {
         "SystemIdentifier" => one, # Required by HPXML schema
-        "[ExteriorAdjacentTo='outside' or ExteriorAdjacentTo='attic - vented' or ExteriorAdjacentTo='attic - unvented' or ExteriorAdjacentTo='basement - conditioned' or ExteriorAdjacentTo='basement - unconditioned' or ExteriorAdjacentTo='crawlspace - vented' or ExteriorAdjacentTo='crawlspace - unvented' or ExteriorAdjacentTo='garage' or ExteriorAdjacentTo='other housing unit']" => one,
+        "[ExteriorAdjacentTo='outside' or ExteriorAdjacentTo='attic - vented' or ExteriorAdjacentTo='attic - unvented' or ExteriorAdjacentTo='basement - conditioned' or ExteriorAdjacentTo='basement - unconditioned' or ExteriorAdjacentTo='crawlspace - vented' or ExteriorAdjacentTo='crawlspace - unvented' or ExteriorAdjacentTo='garage' or ExteriorAdjacentTo='other housing unit' or ExteriorAdjacentTo='other heated space' or ExteriorAdjacentTo='other multifamily buffer space' or ExteriorAdjacentTo='other non-freezing space']" => one,
```
yzhou601: Do we want to also support foundation walls and rim joists to be "ExteriorAdjacentTo" these new space types?

yzhou601:
@shorowit Just noticed [Windows and GlassDoors are not allowed to have Other Side Coefficients.](https://bigladdersoftware.com/epx/docs/9-2/input-output-reference/group-thermal-zone-description-geometry.html#field-outside-boundary-condition-object-10)

HPXMLtoOpenStudio/resources/EPvalidator.rb
shorowit: Yes, as well as FrameFloors.

tasks.rb
```diff
@@ -141,6 +141,9 @@ def create_hpxmls
     'base-enclosure-windows-inoperable.xml' => 'base.xml',
     'base-enclosure-windows-interior-shading.xml' => 'base.xml',
     'base-enclosure-windows-none.xml' => 'base.xml',
+    'base-enclosure-wall-unrated-heated-space.xml' => 'base.xml',
+    'base-enclosure-two-wall-multifamily-buffer.xml' => 'base.xml',
+    'base-enclosure-wall-non-freezing-space.xml' => 'base.xml',
```
shorowit: Let's just put all the new spaces into a single test file that includes all the possible adjacency scenarios (once you handle foundation walls, rim joists, and frame floors), so we don't create a large proliferation of new test files.

shorowit:
> @shorowit Just noticed [Windows and GlassDoors are not allowed to have Other Side Coefficients.](https://bigladdersoftware.com/epx/docs/9-2/input-output-reference/group-thermal-zone-description-geometry.html#field-outside-boundary-condition-object-10). I guess the reason is, "It should be noted that solar effects are not accounted for when other side coefficients are used."

Hmm. The documentation seems to suggest that it works for SubSurfaces, but not for Windows/GlassDoors. Isn't a Window or GlassDoor a SubSurface? Can you help me understand what they are trying to say?

tasks.rb
shorowit: Could be called 'base-enclosure-attached-multifamily.xml'?

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4245,11 +4246,41 @@ def self.set_surface_exterior(model, spaces, surface, surface_id, exterior_adjac
...
+      end
+      otherside_object.setName(exterior_adjacent_to)
+      otherside_object.setCombinedConvectiveRadiativeFilmCoefficient(30)
+      otherside_object.setMinimumOtherSideTemperatureLimit(temp_min)
+      otherside_object.setZoneAirTemperatureCoefficient(indoor_weight)
+      otherside_object.setExternalDryBulbTemperatureCoefficient(outdoor_weight)
+      surface.setSurfacePropertyOtherSideCoefficients(otherside_object)
+      spaces[exterior_adjacent_to] = otherside_object
+    else
+      surface.setSurfacePropertyOtherSideCoefficients(spaces[exterior_adjacent_to])
```
yzhou601: I'm letting the surfaces adjacent to the same new space type share one single `SurfacePropertyOtherSideCoefficients` object, for surfaces that are not adjacent to the same interior space, this will cause issues. Will switch to one object per surface.

yzhou601:
> Can you also look into reporting the timeseries temperatures for these new spaces in the reporting measure? There is an argument in the reporting measure to report space temperatures right now, but the code only looks at EnergyPlus thermal zones and needs to be updated. You can then create a new test based off of `test_timeseries_hourly_temperatures` in SimulationOutputReport\tests\output_report_test.rb that uses your new test file and shows these additional space temperatures are reported.

The new space is not always only one temperature set per type. For example, there're two surfaces adjacent to two different zones but both adjacent to a eg. Multifamily Buffer Boundary. In this case, is it supposed to report one Multifamily Buffer Boundary temperature or two? If two, how to properly name them in the report?

For the [comment](https://github.com/NREL/OpenStudio-HPXML/pull/310#discussion_r386756589), I further tested two surfaces(adjacent to two thermal zones) attached to one singe instance of `SurfacePropertyOtherSideCoefficients`. It turns out to be that the simulation follows the correct path, which means they actually have two sets of `Surface Other Side Coefficients Exterior Air Drybulb Temperature` for each surface. But weird enough that E+ only reports one column of `Surface Other Side Coefficients Exterior Air Drybulb Temperature` if only one object created.

So back to the question, if we want multiple sets of temperature to be shown in our reporting measure, then I'll switch to one `SurfacePropertyOtherSideCoefficients` object per surface attached.

yzhou601:
⏫ , Should proceed with only simulating one othersidecoefficient temperature per space type with conditioned zone temperature, not following E+ logic of calculating it with surface zone which may vary. 
Will switch to use EMS to achieve it.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4272,8 +4272,33 @@ def self.add_otherside_coefficients(surface, exterior_adjacent_to, model, spaces
       otherside_object.setName(exterior_adjacent_to)
       otherside_object.setCombinedConvectiveRadiativeFilmCoefficient(30)
       otherside_object.setMinimumOtherSideTemperatureLimit(temp_min)
-      otherside_object.setZoneAirTemperatureCoefficient(indoor_weight)
+      otherside_object.setConstantTemperatureCoefficient(indoor_weight)
       otherside_object.setExternalDryBulbTemperatureCoefficient(outdoor_weight)
+
+      if @other_side_indoor_sch.nil?
```
yzhou601: All other side coefficient objects (no matter if of the same space type), share the same EMS program to read conditioned zone air temperature. 
EMS only reads zone temperature, not calculating other side temperature, letting othersidecoefficient object weigh between constant schedule value and outdoor air temperature. In this way, we can avoid duplicated EMS programs.

tasks.rb
```diff
@@ -144,6 +144,7 @@ def create_hpxmls
     'base-enclosure-windows-inoperable.xml' => 'base.xml',
     'base-enclosure-windows-interior-shading.xml' => 'base.xml',
     'base-enclosure-windows-none.xml' => 'base.xml',
+    'base-enclosure-attached-multifamily.xml' => 'base.xml',
```
shorowit: And can you add a window and door to a surface adjacent to one of these new space types? I forget if we decided it might be an issue for our modeling approach...

yzhou601:
HVAC sizing needs to be updated with duct and surface changes

tasks.rb
yzhou601: Door added to base-enclosure-attached-multifamily.xml, window added to a newly added invalid file.

yzhou601:
Current appliance locations for new space type:

OS-HPXML appliances | HPXML Location Element Type | Type Shared With | Location Enumeration For New Spaces
-- | -- | -- | --
ClothesWasher | LaundryMachineLocation | ClothesDryer | other
ClothesDryer | LaundryMachineLocation | ClothesWasher | other
Dishwasher | No | No | No
Refrigerator | RefrigeratorLocation | Freezer | other
CookingRange | No | No | No
Dehumidifier | DehumidifierLocation | No | other

Based on my understanding, we will want to add locations for dishwasher, refrigerator, and cooking range right? Though it seems hard to imagine these kitchen appliances to be located on public spaces.

shorowit:
@yzhou601 Let's allow every major appliance to have a location specified. I agree that it's unlikely for some appliances to be located in these MF common spaces, but better to just allow it for all than to have someone complain in the future.

yzhou601:
CI error is not able to be reproduced locally.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
     elsif [HPXML::LocationBasementConditioned].include? exterior_adjacent_to
       surface.createAdjacentSurface(create_or_get_space(model, spaces, HPXML::LocationLivingSpace))
       @cond_bsmnt_surfaces << surface
+    elsif [HPXML::LocationOtherHeatedSpace, HPXML::LocationOtherMultifamilyBufferSpace, HPXML::LocationOtherNonFreezingSpace].include? exterior_adjacent_to
+      add_otherside_coefficients(surface, exterior_adjacent_to, model, spaces)
+      surface.setSunExposure('NoSun')
+      surface.setWindExposure('NoWind')
```
shorowit, line 4197: Minor suggestion, but it seems like these should be moved inside of add_otherside_coefficients, since you'd never want to call that method without also setting these.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
...
     end
   end
 
+  def self.add_otherside_coefficients(surface, exterior_adjacent_to, model, spaces)
+    if spaces[exterior_adjacent_to].nil?
+      # Create E+ other side coefficient object
+      otherside_object = OpenStudio::Model::SurfacePropertyOtherSideCoefficients.new(model)
+      otherside_object.setName(exterior_adjacent_to)
+      # Fixme: assumption the same as SurfacePropertyConvectionCoefficients of return air plenum
+      otherside_object.setCombinedConvectiveRadiativeFilmCoefficient(30)
```
shorowit, line 4209: Hmm. I'm inclined to say we should just use a value that is <= 0. See https://bigladdersoftware.com/epx/docs/9-2/input-output-reference/group-advanced-surface-concepts.html#field-combined-convectiveradiative-film-coefficient

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
...
+      # Schedule of space temperature, can be shared with water heater/appliances
+      create_outside_boundary_schedule(model, exterior_adjacent_to, spaces)
+      # FIXME: wait for new OS release with bugfix of https://github.com/NREL/OpenStudio/issues/3848
+      otherside_object.setPointer(9, @mf_temp_sch_map[exterior_adjacent_to].handle)
```
shorowit, line 4213: It doesn't seem like we actually need this `@mf_temp_sch_map` global variable. Can't we obtain the schedule from the `SurfacePropertyOtherSideCoefficients` object, which itself can be retrieved from the `spaces` variable if needed?

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
...
+      temp_min = UnitConversions.convert(50, 'F', 'C')
+      indoor_weight = 0.5
+      outdoor_weight = 0.5
+    elsif outside_space == HPXML::LocationOtherNonFreezingSpace
+      # Floating with outdoor air temperature with minimum of 40 deg-F
+      temp_min = UnitConversions.convert(40, 'F', 'C')
+      indoor_weight = 0.0
+      outdoor_weight = 1.0
+    elsif outside_space == HPXML::LocationOtherHousingUnit
+      # For water heater, duct, appliances, etc.
```
shorowit, line 4247: The comments here and elsewhere suggest that this applies to appliances, but I don't think that's true. Yes, ducts and water heaters will point to these space types, but appliances won't.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
...
+
   # Returns an OS:Space, or nil if the location is outside the building
   def self.get_space_from_location(location, object_name, model, spaces)
-    if (location == HPXML::LocationOtherExterior) || (location == HPXML::LocationOutside)
-      return
+    return if (location == HPXML::LocationOtherExterior) || (location == HPXML::LocationOutside) || (location == HPXML::LocationOther)
+
+    if (location == HPXML::LocationOtherHeatedSpace) || (location == HPXML::LocationOtherHousingUnit) || (location == HPXML::LocationOtherMultifamilyBufferSpace) || (location == HPXML::LocationOtherNonFreezingSpace)
+      create_outside_boundary_schedule(model, location, spaces)
+      return @mf_temp_sch_map[location]
```
shorowit, line 4289: I'm tempted to say we shouldn't ever be creating a new MF space schedule here. For regular spaces, we give an error (below) if there is an appliance, water heater, etc. in the building that is adjacent to a space (e.g., garage) that had not previously been created. I'm thinking we should do the same here?

```
if (location == HPXML::LocationOtherHeatedSpace) || (location == HPXML::LocationOtherHousingUnit) || (location == HPXML::LocationOtherMultifamilyBufferSpace) || (location == HPXML::LocationOtherNonFreezingSpace)
  if @mf_temp_sch_map[location].nil?
    fail "#{object_name} location is '#{location}' but building does not have this location specified."
  end
  return @mf_temp_sch_map[location]
end
```

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4169,15 +4191,102 @@ def self.set_surface_exterior(model, spaces, surface, exterior_adjacent_to)
...
+    program.addLine('EndIf')
+
+    program_cm = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    program_cm.setName("#{program.name} calling manager")
+    program_cm.setCallingPoint('EndOfSystemTimestepAfterHVACReporting')
+    program_cm.addProgram(program)
+  end
+
   # Returns an OS:Space, or nil if the location is outside the building
   def self.get_space_from_location(location, object_name, model, spaces)
```
shorowit, line 4284: This is very confusing that the method says it returns a space, the objects assigned (e.g., `duct_space`) suggest it's a space, but it may be a schedule now. Let's update the name of this method to `get_space_or_temp_schedule_from_location` and update the comment above too. Then we should return either` [space, nil]` or `[nil, schedule]` so we can do, e.g.,

duct_space, duct_temp_schedule = get_space_or_temp_schedule_from_location(...)

Then subsequent code can use one or the other appropriately.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -733,6 +733,10 @@ def self.process_ducts(model, ducts, building, air_loop)
       if duct.space.nil? # Outside
         duct.zone = nil
         duct.zone_handle = HPXML::LocationOutside
+      elsif duct.space.is_a? OpenStudio::Model::ScheduleConstant
+        # Pass schedule and its name to duct.zone
+        duct.zone = duct.space
```
shorowit, line 738: By making changes suggested above, we should now be able to clean this up so that it's not a space being assigned to a zone (?!), where the space actually represents a schedule (?!).

HPXMLtoOpenStudio/measure.rb
yzhou601: In that way, the surface exterior temperature will be exactly the new space temperature calculated: instead of assuming there's air between.

HPXMLtoOpenStudio/measure.rb
yzhou601: Good idea, I admit that this part is pretty confusing.

HPXMLtoOpenStudio/measure.rb
shorowit: I realize that. It just seemed like an okay assumption given that the temperatures are already completely arbitrary?

SimulationOutputReport/measure.rb
```diff
@@ -613,17 +614,27 @@ def get_outputs(timeseries_frequency,
     # Get zone temperatures
     if include_timeseries_zone_temperatures
       zone_names = []
+      other_side_obj_names = []
       @model.getThermalZones.each do |zone|
         if zone.floorArea > 1
           zone_names << zone.name.to_s.upcase
         end
       end
+      @model.getSurfacePropertyOtherSideCoefficientss.each do |otherside_obj|
```
shorowit, line 623: For buildings that have a water heater and/or duct in one of the new MF space types, but there are no surfaces adjacent to the MF space type, the space temperatures aren't accounted for here. We need to account for them. (We don't need to worry about appliances, which are only assigned to "other" anyway.)

workflow/sample_files/base-enclosure-other-housing-unit.xml
```diff
@@ -277,7 +284,7 @@
             <SystemIdentifier id='WaterHeater'/>
             <FuelType>electricity</FuelType>
             <WaterHeaterType>storage water heater</WaterHeaterType>
-            <Location>living space</Location>
+            <Location>other housing unit</Location>
```
shorowit, line 287: It seems like everything except duct location is set to 'other housing unit' in this file. We should also set duct location, right?

shorowit:
CI results look good. Code looks good. Thank you @yzhou601!!

joseph-robertson:
Should this affect any of our BuildResHPXML work?

shorowit:
@joseph-robertson I suggest 1) adding location arguments for the dishwasher and clothes washer, so that all appliances have them, and 2) adding HPXML::LocationOther as a choice for appliances.

PR #314, opened 2020-03-11
HPXML class, second pass

shorowit:
Improvements to new HPXML class:
1. Adds helper methods (e.g., wall.windows, window.wall). 
2. Updates class to better mirror HPXML schema.
3. Adds constants.


shorowit:
TODO: Add constants.

PR #327, opened 2020-03-23
Updates to support BuildResidentialHPXML, OpenStudio 2.9.1

joseph-robertson:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


BuildResidentialHPXML/measure.rb
```diff
@@ -2810,8 +2829,19 @@ def self.set_cooling_systems(hpxml, runner, args)
...
+      elsif args[:cooling_system_cooling_efficiency] <= 21.0
+        hpxml.cooling_systems[-1].compressor_type = 'two stage'
+        hpxml.cooling_systems[-1].cooling_shr = 0.73
+      else
+        hpxml.cooling_systems[-1].compressor_type = 'variable speed'
+        hpxml.cooling_systems[-1].cooling_shr = 0.78
+      end
     elsif [HPXML::HVACTypeRoomAirConditioner].include? cooling_system_type
       hpxml.cooling_systems[-1].cooling_efficiency_eer = args[:cooling_system_cooling_efficiency]
+      hpxml.cooling_systems[-1].cooling_shr = 0.65
```
shorowit: No, don't just copy the default logic from the HPXMLtoOS measure. Either 1) expose them as inputs, or 2) if you don't want to add inputs, remove them from your generated HPXMLs (which means they use the default logic in the HPXMLtoOS measure) and exclude them from your test comparisons.

BuildResidentialHPXML/measure.rb
```diff
@@ -2867,9 +2905,26 @@ def self.set_heat_pumps(hpxml, runner, args)
...
+      else
+        if heat_pump_type == HPXML::HVACTypeHeatPumpAirToAir
+          hpxml.heat_pumps[-1].compressor_type = 'variable speed'
+        end
+        hpxml.heat_pumps[-1].cooling_shr = 0.78
+      end
     elsif [HPXML::HVACTypeHeatPumpGroundToAir].include? heat_pump_type
       hpxml.heat_pumps[-1].heating_efficiency_cop = args[:heat_pump_heating_efficiency]
       hpxml.heat_pumps[-1].cooling_efficiency_eer = args[:heat_pump_cooling_efficiency]
+      hpxml.heat_pumps[-1].cooling_shr = 0.73
```
shorowit: Same comment here.

BuildResidentialHPXML/measure.rb
```diff
@@ -3151,7 +3206,8 @@ def self.set_water_heating_systems(hpxml, runner, args)
                                     recovery_efficiency: recovery_efficiency,
                                     related_hvac_idref: related_hvac_idref,
                                     standby_loss: standby_loss,
-                                    jacket_r_value: jacket_r_value)
+                                    jacket_r_value: jacket_r_value,
+                                    temperature: 125.0)
```
shorowit: And here.

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,3472 @@
...
+
+    recirculation_control_type_choices = OpenStudio::StringVector.new
+    recirculation_control_type_choices << HPXML::DHWRecirControlTypeNone
+    recirculation_control_type_choices << HPXML::DHWRecirControlTypeTimer
+    recirculation_control_type_choices << HPXML::DHWRecirControlTypeTemperature
+    recirculation_control_type_choices << HPXML::DHWRecirControlTypeSensor
+    recirculation_control_type_choices << HPXML::DHWRecirControlTypeManual
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('recirculation_control_type', recirculation_control_type_choices, true)
+    arg.set
```
shorowit: And same here.

joseph-robertson: Which are you referring to here?

shorowit: backup_heating_switchover_temp: 25.0

BuildResidentialHPXML/measure.rb
joseph-robertson: I decided to not expose `cooling_shr` and instead exclude it from the test comparisons.

BuildResidentialHPXML/measure.rb
joseph-robertson: I decided to expose `water_heater_setpoint_temperature`.

BuildResidentialHPXML/measure.rb
joseph-robertson: I decided to expose `heat_pump_backup_heating_switchover_temp`.

BuildResidentialHPXML/measure.rb
shorowit: Still need to do something with compressor type.

joseph-robertson: It follows the efficiency value, an exposed argument. So is it actually "hardcoded"?

joseph-robertson: Or maybe you're implying that it's possible, e.g., an efficiency of 16 is entered but the compressor type is `single stage`?

BuildResidentialHPXML/tests/build_residential_hpxml_test.rb
```diff
@@ -176,9 +176,6 @@ def _check_hpxmls(workflow_dir, built_dir, test_dir, hpxml_path)
       hpxml.cooling_systems.each do |cooling_system|
         cooling_system.cooling_shr = nil
       end
-      hpxml.heat_pumps.each do |heat_pump|
-        heat_pump.cooling_shr = nil
-      end
```
shorowit: Why not remove the 3 lines right above it too?

BuildResidentialHPXML/resources/geometry.rb
```diff
@@ -1427,8 +1442,8 @@ def self.create_doors(runner:,
 
       door_sub_surface = OpenStudio::Model::SubSurface.new(door_polygon, model)
       door_sub_surface.setName("#{min_story_avail_wall.name} - Door")
-      door_sub_surface.setSubSurfaceType('Door')
       door_sub_surface.setSurface(min_story_avail_wall)
+      door_sub_surface.setSubSurfaceType('Door')
```
joseph-robertson, line 1446: Whoa. Was this breaking anything before? What about in ResStock?

shorowit: I don't know. It was definitely causing an issue in this measure, where the door was being translated to a HPXML Window.

See https://github.com/NREL/OpenStudio/issues/3936.

joseph-robertson: Hmm, then wouldn't there have been no doors in any of the hpxml files created from test osws?

BuildResidentialHPXML/measure.rb
```diff
@@ -833,9 +833,9 @@ def arguments(model)
 
     arg = OpenStudio::Measure::OSArgument::makeStringArgument('heating_system_heating_capacity', true)
     arg.setDisplayName('Heating System: Heating Capacity')
-    arg.setDescription("The output heating capacity of the heating system. If using '#{Constants.SizingAuto}', the autosizing algorithm will use ACCA Manual S to set the capacity.")
+    arg.setDescription("The output heating capacity of the heating system. If using '#{Constants.Auto}', the autosizing algorithm will use ACCA Manual S to set the capacity.")
     arg.setUnits('Btu/hr')
-    arg.setDefaultValue(Constants.SizingAuto)
```
joseph-robertson: Should `SizingAuto` be removed from constants.rb?

shorowit: The HPXMLtoOS measure uses it.

BuildResidentialHPXML/measure.rb
```diff
@@ -1802,18 +1802,18 @@ def arguments(model)
...
+    arg = OpenStudio::Measure::OSArgument::makeDStringArgument('plug_loads_television_annual_kwh', true)
     arg.setDisplayName('Plug Loads: Television Annual kWh')
     arg.setDescription('The annual energy consumption of the television plug loads.')
     arg.setUnits('kWh/yr')
-    arg.setDefaultValue(620)
+    arg.setDefaultValue(Constants.Auto)
     args << arg
 
-    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('plug_loads_other_annual_kwh', true)
+    arg = OpenStudio::Measure::OSArgument::makeStringArgument('plug_loads_other_annual_kwh', true)
```
shorowit, line 1812: Note that when you make changes like this, the osw files will no longer run via `openstudio run -w foo.osw` because it complains that the osws have arguments with the wrong type (double instead of string). Not sure how important it is, but I try to remember to update tasks.rb for changes like this. The osw's do run via the meta-measure because it does the datatype conversions automatically, so that's why our tests still work.

joseph-robertson: You're saying that in tasks.rb, the setArgument argument needs to be a string, and not a double anymore?

shorowit: Yes. Technically.

joseph-robertson: Ok, I will make those changes.

PR #329, opened 2020-03-24
Skip creating tiny surfaces

yzhou601:
## Pull Request Description

Skip creating surfaces smaller than 0.1 sqft in our workflow.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -1296,11 +1305,13 @@ def self.add_foundation_walls_slabs(runner, model, spaces)
       slabs = []
       @hpxml.foundation_walls.each do |foundation_wall|
```
yzhou601: By adding conditions to skip small surfaces, I noticed it's looping through hpxml elements multiple times here, thus makes me wonder it might be a better way to handle it. Maybe use a hash to map foundation_type with its assigned slab and foundation walls? And this can also be used in the error checking above?

shorowit: Good thinking. I will replace these two loops when I implement the more general error-checks to ensure spaces have minimum adjacent surface types, so it's fine for now.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1889,7 +1896,7 @@ def self.add_conditioned_floor_area(runner, model, spaces)
     end
 
     addtl_cfa = cfa - model_cfa
-    return unless addtl_cfa > 0
+    return unless addtl_cfa > 0.1
```
shorowit, line 1899: Nice find.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1268,6 +1268,8 @@ def self.create_or_get_space(model, spaces, spacetype)
 
   def self.add_roofs(runner, model, spaces)
     @hpxml.roofs.each do |roof|
+      next if roof.net_area < 0.1 # skip surfaces smaller than 0.1 sqft
```
shorowit, line 1271: So I'm seeing checks like this for only 3 surface types: roofs, walls, and foundation_walls. I assume those three are called out because they are the only surfaces that can have subsurfaces, so we're checking the net area here (while the HPXML class already checked gross area).

If that's the logic, then I suggest we change the comments here and below from:
`# skip surfaces smaller than 0.1 sqft`
to:
`# skip modeling net surface area for surfaces comprised entirely of subsurface area`

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3795,6 +3796,18 @@ def delete_partition_surfaces()
     end
   end
 
+  def delete_tiny_surfaces()
+    (@rim_joists + @walls + @foundation_walls + @frame_floors + @roofs + @windows + @skylights + @doors + @slabs).reverse_each do |surface|
+      next if surface.area > 0.1
+
+      # error out if the only wall is too small. It skipped EPvalidator but should be blocked here.
+      if (surface.is_a? Wall) && (@walls.size == 1)
+        fail "The only wall specified:#{surface.id} is smaller than 0.1 sqft."
```
shorowit, line 3805: Likewise can remove this error check.

tasks.rb
```diff
@@ -818,6 +820,17 @@ def set_hpxml_walls(hpxml_file, hpxml)
...
     hpxml.walls[1].insulation_assembly_r_value = 23.0
+  elsif ['base-enclosure-tiny-wall.xml'].include? hpxml_file
+    hpxml.walls.add(id: 'TinyWall',
+                    exterior_adjacent_to: HPXML::LocationOutside,
+                    interior_adjacent_to: HPXML::LocationLivingSpace,
+                    wall_type: HPXML::WallTypeWoodStud,
+                    area: 0.05,
+                    solar_absorptance: 0.7,
+                    emittance: 0.92,
+                    insulation_assembly_r_value: 20.0)
```
shorowit, line 831: This adds another simulation for a fairly uncommon problem. It also only checks walls.

I suggest 1) Removing this test file and adding this surface to the base-enclosure-split-surfaces.xml file instead, and 2) Adding similar tiny surface for roofs, rim joists, foundation walls, windows, etc.

tasks.rb
```diff
@@ -57,6 +57,7 @@ def create_hpxmls
     'invalid_files/water-heater-location.xml' => 'base.xml',
     'invalid_files/water-heater-location-other.xml' => 'base.xml',
     'invalid_files/slab-zero-exposed-perimeter.xml' => 'base.xml',
+    'invalid_files/only-wall-tiny.xml' => 'base-atticroof-flat.xml',
```
shorowit, line 60: Remove this, we'll cover this and more in the PR to ensure spaces have minimum adjacent surface types.

workflow/tests/hpxml_translator_test.rb
```diff
@@ -830,6 +831,7 @@ def _verify_simulation_outputs(runner, rundir, hpxml_path, results)
       end
       query = "SELECT SUM(Value) FROM TabularDataWithStrings WHERE ReportName='EnvelopeSummary' AND ReportForString='Entire Facility' AND TableName='Opaque Exterior' AND (RowName='#{wall_id}' OR RowName LIKE '#{wall_id}:%' OR RowName LIKE '#{wall_id} %') AND ColumnName='Net Area' AND Units='m2'"
       sql_value = UnitConversions.convert(sqlFile.execAndReturnFirstDouble(query).get, 'm^2', 'ft^2')
+      assert_operator(sql_value, :>, 0.01)
```
shorowit, line 834: Good idea! Should there be similar checks for other surface types? Windows, doors, slabs, ...

PR #341, opened 2020-04-07
Report timestamps in timeseries output, try 2

joseph-robertson:
## Pull Request Description

Try 2:

Instead of integers, report either hourly, daily, or timestep timestamps. This will make connecting hpxml-based ResStock to buildstockbatch much easier.

https://trello.com/c/9MJxjnGa/39-timestamps-instead-of-integers-for-timeseries-index

## Checklist

Not all may apply:


joseph-robertson:
The Time table for `test_timeseries_hourly_loads` has only IntervalType=3 and IntervalType=4 rows populated, resulting in error.

@shorowit

shorowit:
@joseph-robertson I fixed `test_annual_only`. The two remaining tests that fail are `test_timeseries_hourly_componentloads` and `test_timeseries_hourly_loads`, both of which are hitting  [a bug I fixed for OpenStudio v3](https://github.com/NREL/OpenStudio-workflow-gem/pull/97). Pretty sure that once we switch to v3, the tests will pass.

joseph-robertson:
Reminder:

Once this PR is merged, delete this branch: https://github.com/NREL/OpenStudio-HPXML/tree/ts-timestamps

shorowit:
@joseph-robertson I found a bug here. I ran the `test_timeseries_timestep_ALL_10min` test, looked at the results_timeseries.csv file, and found the last row doesn't have a timestamp. Maybe everything is offset by 1 timestamp?

image

joseph-robertson:
> @joseph-robertson I found a bug here. I ran the `test_timeseries_timestep_ALL_10min` test, looked at the results_timeseries.csv file, and found the last row doesn't have a timestamp. Maybe everything is offset by 1 timestamp?
> 
> image

Look at this row in a text editor (it looks correct). Perhaps excel isn't recognizing this datetime for some reason?

shorowit:
Ah, you're right. This happens every day too, not just the end of the year. I now remember that [this is an EnergyPlus issue](https://github.com/NREL/EnergyPlus/issues/5360).

I strongly think we should workaround the issue because it makes processing the timestamp in Excel, python, etc. a real pain. Is it easy for you to change from 24:00 to 00:00 and increment the day (which may also require incrementing the month or even year)?

joseph-robertson:
> Ah, you're right. This happens every day too, not just the end of the year. I now remember that [this is an EnergyPlus issue](https://github.com/NREL/EnergyPlus/issues/5360).
> 
> I strongly think we should workaround the issue because it makes processing the timestamp in Excel, python, etc. a real pain. Is it easy for you to change from 24:00 to 00:00 and increment the day (which may also require incrementing the month or even year)?

Yes, it is possible. I think this is part of the reason why I went with the `ts-timestamps` branch (using the `Time.new(...)` route)...

shorowit:
I kind of see. But that was a fair amount of custom code to support and could deviate from EnergyPlus-reported timestamps or have bugs for edge cases. I still think it's best to use the EnergyPlus timestamps directly, just need a small (?) workaround on top.

joseph-robertson:
> I kind of see. But that was a fair amount of custom code to support and could deviate from EnergyPlus-reported timestamps or have bugs for edge cases. I still think it's best to use the EnergyPlus timestamps directly, just need a small (?) workaround on top.

Yeah. I think I can do sort of a hybrid of the two.

PR #356, opened 2020-04-17
Kitchen/Bath fans

shorowit:
## Pull Request Description

Adds ability to model kitchen/bath fans (spot ventilation).

## Checklist

Not all may apply:


shorowit:
@joseph-robertson This was merged into master. If you add these new inputs to the BuildResHPXML measure, we can then use them in ResStock. Note that many of the inputs are optional and you can default to 'auto'.

PR #357, opened 2020-04-17
HPXML Measure Defaults

bpark1327:
## Pull Request Description

Add HPXML defaults including
- Hot water recirculation pump power
- Hot water recirculation loop length
- Hot water recirculation branch length
- Solar thermal storage volume
- Storage water heater capacity
- Storage water heater tank volume

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -535,8 +535,7 @@ def self.run_validator(hpxml_doc)
       ## [WHType=Tank]
       '/HPXML/Building/BuildingDetails/Systems/WaterHeating/WaterHeatingSystem[WaterHeaterType="storage water heater"]' => {
         'FuelType[text()="natural gas" or text()="fuel oil" or text()="propane" or text()="electricity" or text()="wood"]' => one, # If not electricity, see [WHType=FuelTank]
-        'TankVolume' => one,
-        'HeatingCapacity' => one,
+        'TankVolume | HeatingCapacity' => zero_or_two,
```
shorowit, line 538: We can independently allow these to be provided, so you can just use zero_or_one for each of them. Will also need to update measure.rb code and the documentation accordingly.

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -596,9 +595,7 @@ def self.run_validator(hpxml_doc)
       ## [HWDistType=Recirculation]
       '/HPXML/Building/BuildingDetails/Systems/WaterHeating/HotWaterDistribution/SystemType/Recirculation' => {
         'ControlType[text()="manual demand control" or text()="presence sensor demand control" or text()="temperature" or text()="timer" or text()="no control"]' => one,
-        'RecirculationPipingLoopLength' => one,
-        'BranchPipingLoopLength' => one,
-        'PumpPower' => one,
+        'RecirculationPipingLoopLength | BranchPipingLoopLength | PumpPower' => zero_or_three,
```
shorowit, line 598: Same here.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -689,16 +708,16 @@ The dishwasher is assumed to be in the living space.
...
-LabelUsage               4 [cyc/week]
-=======================  =================
+===============================  =================
+Element Name                     Default Value
+===============================  =================
+RatedAnnualkWh [kwh/yr]          467  
+LabelElectricRate [$/kWh]        0.12  
+LabelGasRate [$/therm]           1.09  
+LabelAnnualGasCost [$]           33.12  
+PlaceSettingCapacity [standard]  12
```
shorowit, line 718: This should be `[#]` not `[standard]`.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -517,7 +517,7 @@ Depending on the type of water heater specified, additional elements are require
 ========================================  ===================================  ===========  ==========  ===============  ========================  =================  =================  =========================================
 WaterHeaterType                           UniformEnergyFactor or EnergyFactor  FuelType     TankVolume  HeatingCapacity  RecoveryEfficiency        RelatedHVACSystem  UsesDesuperheater  WaterHeaterInsulation/Jacket/JacketRValue
 ========================================  ===================================  ===========  ==========  ===============  ========================  =================  =================  =========================================
-storage water heater                      required                             <any>        required    <optional>       required if non-electric                     <optional>         <optional>
+storage water heater                      required                             <any>        <optional>  <optional>       required if non-electric                     <optional>         <optional>
```
shorowit, line 520: For TankVolume and HeatingCapacity defaulting, you can point to Table 7 in the [Building America House Simulation Protocols](https://www1.eere.energy.gov/buildings/publications/pdfs/building_america/house_simulation.pdf). I believe it should be identical to the equations, but please verify.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -554,13 +553,26 @@ CFA = conditioned floor area [ft²],
 NCfl = number of conditioned floor levels number of conditioned floor levels in the residence, including conditioned basements, 
 bsmnt = presence = 1.0 or absence = 0.0 of an unconditioned basement in the residence.
 
-For a ``SystemType/Recirculation`` system, the following elements are required:
+For a ``SystemType/Recirculation`` system, the following element is required:
```
shorowit, line 556: Rather than splitting the list into required and optional inputs for recirculation systems, how about using a single list where you call out certain inputs as optional? See this [simple solar hot water](https://github.com/NREL/OpenStudio-HPXML/blob/solar_fraction/docs/source/hpxml_to_openstudio.rst#solar-thermal) section where ConnectedTo is called out as optional.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -592,14 +604,21 @@ If using simple inputs, the following element is required:
...
 - ``CollectorLoopType``: 'liquid indirect' or 'liquid direct' or 'passive thermosyphon'
 - ``CollectorType``: 'single glazing black' or 'double glazing black' or 'evacuated tube' or 'integrated collector storage'
 - ``CollectorAzimuth``
 - ``CollectorTilt``
 - ``CollectorRatedOpticalEfficiency``: FRTA (y-intercept); see Directory of SRCC OG-100 Certified Solar Collector Ratings
-- ``CollectorRatedThermalLosses``: FRUL (slope, in units of Btu/hr-ft^2-R); see Directory of SRCC OG-100 Certified Solar Collector Ratings
-- ``StorageVolume``
+- ``CollectorRatedThermalLosses``: FRUL (slope, in units of Btu/hr-ft²-R); see Directory of SRCC OG-100 Certified Solar Collector Ratings
+
+the following element can be optionally entered:
```
shorowit, line 615: Same as previous comment, can we keep this as a single list?

tasks.rb
```diff
@@ -90,6 +90,7 @@ def create_hpxmls
     'base-dhw-recirc-nocontrol.xml' => 'base.xml',
     'base-dhw-recirc-temperature.xml' => 'base.xml',
     'base-dhw-recirc-timer.xml' => 'base.xml',
+    'base-dhw-recirc-defaults.xml' => 'base-dhw-recirc-demand.xml',
```
shorowit, line 93: Hmm, I guess we can't handle this in base-misc-defaults.xml because there can only be a single hot water distribution system and it's standard in that file. 

How about renaming this file to base-misc-defaults2.xml, so that it will be able to more generally accommodate defaulting of multiple different building components?

tasks.rb
```diff
@@ -99,6 +100,7 @@ def create_hpxmls
     'base-dhw-solar-thermosyphon-evacuated-tube.xml' => 'base.xml',
     'base-dhw-solar-thermosyphon-flat-plate.xml' => 'base.xml',
     'base-dhw-solar-thermosyphon-ics.xml' => 'base.xml',
+    'base-dhw-solar-direct-flat-plate-defaults.xml' => 'base-dhw-solar-direct-flat-plate.xml',
```
shorowit, line 103: Can you remove this file and incorporate the solar system into base-misc-defaults.xml?

shorowit: Actually, I don't see any references to base-dhw-solar-direct-flat-plate-defaults.xml, so it seems like you are already incorporating it in base-misc-defaults.xml....

HPXMLtoOpenStudio/measure.rb
```diff
@@ -477,6 +477,13 @@ def self.set_defaults_and_globals(runner)
         sqft_by_gal = surface_area / act_vol # sqft/gal
         water_heating_system.standby_loss = (2.9721 * sqft_by_gal - 0.4732).round(3) # linear equation assuming a constant u, F/hr
       end
+      if (water_heating_system.water_heater_type == HPXML::WaterHeaterTypeStorage) && water_heating_system.heating_capacity.nil?
+        water_heater_fuel_type = water_heating_system.fuel_type
+        num_water_heaters = @hpxml.water_heating_systems.size
+        nbaths = @hpxml.building_construction.number_of_bathrooms
```
shorowit, line 483: No need for this, you can just reference @nbaths that is assigned above.

tasks.rb
```diff
@@ -528,6 +530,7 @@ def set_hpxml_building_construction(hpxml_file, hpxml)
   elsif ['base-misc-defaults.xml'].include? hpxml_file
     hpxml.building_construction.conditioned_building_volume = nil
     hpxml.building_construction.average_ceiling_height = 8
+    hpxml.building_construction.number_of_bathrooms = 3
```
shorowit, line 533: You should not have to do this, please remove.

tasks.rb
bpark1327: Oops.  I incorporated the solar system into base-misc-defaults.xml and forgot to remove the base-dhw-solar-direct-flat-plate-defaults.xml.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -478,6 +478,21 @@ def self.set_defaults_and_globals(runner)
         sqft_by_gal = surface_area / act_vol # sqft/gal
         water_heating_system.standby_loss = (2.9721 * sqft_by_gal - 0.4732).round(3) # linear equation assuming a constant u, F/hr
       end
+      if (water_heating_system.water_heater_type == HPXML::WaterHeaterTypeStorage)
+        if water_heating_system.heating_capacity.nil?
+          water_heating_system.heating_capacity = Waterheater.get_default_heating_capacity(water_heating_system.fuel_type, @nbeds, @hpxml.water_heating_systems.size, @nbaths)
```
shorowit, line 483: This isn't correct -- note the units being returned from the method vs the units that HPXML uses. When you fix this, double-check the documentation too.

bpark1327: 😱 I fixed it.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -1179,7 +1179,7 @@ def self.get_default_heating_capacity(fuel, num_beds, num_water_heaters, num_bat
       else
         cap_kw = 5.5
       end
-      return UnitConversions.convert(cap_kw, 'kW', 'kBtu/hr')
+      return UnitConversions.convert(cap_kw * 1000.0, 'W', 'Btu/hr')
```
shorowit, line 1182: This correctly changes the unit conversions when the fuel is electricity, but other fuel types are still wrong.

Also, I would recommend keeping the method how it was, since other code is expecting the units it was returning, and instead accommodate the unit conversion in measure.rb.

shorowit:
This looks good now, thanks @bpark1327! Just pulled in master one more time, so I'll merge once the CI comes back with a ✔️.

shorowit:
FYI @joseph-robertson

PR #358, opened 2020-04-20
Desuperheater os-v3

yzhou601:
## Pull Request Description

Desuperheater switch to two tank approach after [E+ 9.3.0 bugfix](https://github.com/NREL/EnergyPlus/pull/7627)

## Checklist

Not all may apply:


yzhou601:
@shorowit  I have run all those five desuperheater test files and looked into their hourly E+ temperature and energy results, all of them are applying Heat Reclaim Recovery Efficiency (0.25 in our assumption) correctly, tank temperatures in both tanks are reasonable, and tank source side energy transfer is consistent with desuperheater heating energy, thus showing that E+ bugfix is working well in our approach. 

The only issue observed is that storage tank is also initialized at its setpoint instead of at mains water temperature, which might impact hot water energy results of the first couple days. But should be easy to resolve with an simple EMS program running only once at the beginning of simulation period. Should we implement that?

shorowit:
Great! Should we do a manual test run for a building with a sub-hourly timesteps too? See base-misc-timestep-10-mins.xml.

The temperature initialization issue sounds pretty minor, so I'd suggest not bothering unless you think it's significant.

yzhou601:
@shorowit I tried base-dhw-desuperheater.xml with 10 mins timestep, giving similar fair results.

yzhou601:
@shorowit I just recalled that the issue for dsh+combi system is that there's only one `RelatedHVAC` node under each`WaterHeatingSystem`, while both desuperheater and combi system require hvac idref information.

shorowit:
@yzhou601 Ah, right. Well of the two, I was definitely more interested in HPWHs. Let's not bother with combi systems then, I don't think it's worth the effort to update the schema and all of our code.

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -558,7 +558,7 @@ def self.run_validator(hpxml_doc)
 
       ## [Desuperheater]
       '/HPXML/Building/BuildingDetails/Systems/WaterHeating/WaterHeatingSystem[UsesDesuperheater="true"]' => {
-        'WaterHeaterType[text()="storage water heater" or text()="instantaneous water heater"]' => one, # Desuperheater is only supported with storage/tankless water heater
+        'WaterHeaterType[text()="storage water heater" or text()="instantaneous water heater" or text()="heat pump water heater"]' => one, # Desuperheater is only supported with storage/tankless water heater
```
shorowit, line 561: Need to update the comment here too, as well as the RTD documentation.

SimulationOutputReport/measure.rb
```diff
@@ -531,29 +531,23 @@ def get_outputs(timeseries_frequency,
...
-          end_use.timeseries_output_by_system[sys_id] = end_use.timeseries_output_by_system[sys_id].zip(dsh_adj_timeseries).map { |x, y| x + y }
         end
         break # only apply once
       end
 
-      # Can only be one desuperheater or solar thermal system
-      if desuperheater_vars.empty?
-        desuperheater_vars = ep_output_names.select { |name| name.include? Constants.ObjectNameDesuperheaterLoad(nil) }
-      end
+      # Can only be one solar thermal system
```
shorowit, line 546: Have we done a test w/ multiple HVAC systems attached to multiple desuperheaters? We should confirm it works correctly.

tasks.rb
```diff
@@ -2651,7 +2652,12 @@ def set_hpxml_water_heating_systems(hpxml_file, hpxml)
   elsif ['base-dhw-uef.xml'].include? hpxml_file
     hpxml.water_heating_systems[0].energy_factor = nil
     hpxml.water_heating_systems[0].uniform_energy_factor = 0.93
-  elsif ['base-dhw-desuperheater.xml'].include? hpxml_file
+  elsif ['base-dhw-desuperheater.xml',
+         'base-dhw-desuperheater-2-speed.xml',
+         'base-dhw-desuperheater-var-speed.xml'].include? hpxml_file
+    hpxml.water_heating_systems[0].uses_desuperheater = true
+    hpxml.water_heating_systems[0].related_hvac_idref = 'CoolingSystem'
+  elsif ['base-dhw-desuperheater-hpwh.xml'].include? hpxml_file
```
shorowit, line 2660: This can be combined with above too, right?

SimulationOutputReport/measure.rb
yzhou601: I've done a manual test with two dshs + two cooling systems.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -1540,50 +1485,11 @@ def self.create_new_heater(name, cap, fuel, act_vol, ef, t_set, space, oncycle_p
...
-      new_heater.setOnCycleParasiticFuelConsumptionRate(avg_elec)
-      new_heater.setOffCycleParasiticFuelConsumptionRate(avg_elec)
-    else
-      new_heater.setOnCycleParasiticFuelConsumptionRate(oncycle_p)
-      new_heater.setOffCycleParasiticFuelConsumptionRate(offcycle_p)
-    end
-    new_heater.setOnCycleParasiticFuelType('electricity')
-    new_heater.setOffCycleParasiticFuelType('electricity')
+    new_heater.setOnCycleParasiticFuelType(HPXML::FuelTypeElectricity)
+    new_heater.setOffCycleParasiticFuelType(HPXML::FuelTypeElectricity)
```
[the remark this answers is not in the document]
yzhou601: Ah no problem, I thought they're strings unintentionally left out without realizing the nuances between E+/HPXML constants logics. Thanks for catching, I will revert.

tasks.rb
yzhou601: I started combining due to the new one but finally left out the new one, haha. 🤦

shorowit:
Thanks @yzhou601!

PR #362, opened 2020-04-22
Allow small tolerance of hvac fraction sum, take 2

shorowit:
## Pull Request Description

This is safer and simpler.

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/measure.rb
+++ b/HPXMLtoOpenStudio/measure.rb
@@ -2460,19 +2460,13 @@ class OSModel
 
   def self.calc_sequential_load_fraction(load_fraction, remaining_fraction)
     if remaining_fraction > 0
-      if (load_fraction - remaining_fraction).abs <= 0.010001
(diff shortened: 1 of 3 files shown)


shorowit:
@yzhou601 The fix here is really [just this](https://github.com/NREL/OpenStudio-HPXML/pull/362/files#diff-f865e7b10a878fbe87058db74b4a5c0eR4501-R4505).

I was looking back at how we addressed this issue previously and got concerned that it still might cause errors under certain circumstances. I could imagine that we still might create an EnergyPlus IDF where the sequential load schedule's value is slightly greater than 1 due to representation of floating-point numbers or the fact that we're doing some rounding. I think this is safer because it ensures we use 1.0 exactly in these cases.

shorowit:
🤯 

Thanks for reviewing. I hadn't even considered that situation, hopefully this more robustly covers all edge cases.

shorowit:
No CI diffs, merging.

PR #372, opened 2020-04-30
BuildResidentialHPXML measure

joseph-robertson:
## Pull Request Description

Adds an OpenStudio measure that can generate an HPXML file. The measure will never support everything that can be described in an HPXML file, but covers the most common situations. It also has simplified geometry inputs (floor area, aspect ratio, etc.) and derives surface areas for the HPXML file automatically.

## Checklist

Not all may apply:

- [x] ~EPvalidator.rb has been updated~


BuildResidentialHPXML/measure.rb
```diff
@@ -3649,30 +3685,42 @@ def self.set_pv_systems(hpxml, runner, args, weather)
...
+                                location: HPXML::LocationInterior,
+                                fration_of_units_in_location: args[:lighting_fraction_cfl_interior].get,
+                                third_party_certification: HPXML::LightingTypeTierI)
+    end
+    if args[:lighting_fraction_cfl_exterior].is_initialized
+      hpxml.lighting_groups.add(id: 'Lighting_TierI_Exterior',
+                                location: HPXML::LocationExterior,
+                                fration_of_units_in_location: args[:lighting_fraction_cfl_exterior].get,
+                                third_party_certification: HPXML::LightingTypeTierI)
+    end
```
shorowit, line 3699: This provides too much flexibility as we need all 6 inputs provided or none provided. If just one input is provided, for example, that's not sufficient and will generate errors from EPvalidator. I suggest making all of the inputs required with default values.

BuildResidentialHPXML/measure.rb
```diff
@@ -1659,6 +1659,36 @@ def arguments(model)
       args << arg
     end
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('lighting_fraction_cfl_interior', false)
+    arg.setDisplayName('Lighting: Fraction CFL Interior')
+    arg.setDescription('Fraction of all lamps (interior) that are compact fluorescent. Lighting not specified as CFL or LED is assumed to be incandescent.')
```
shorowit, line 1664: Replace all "cfl" with "fluorescent" because it's _all_ fluorescent lighting (i.e., linear and compact). Remove any references to "compact".

BuildResidentialHPXML/measure.rb
```diff
@@ -3409,26 +3415,32 @@ def self.set_hvac_distribution(hpxml, runner, args)
...
+
+    if args[:ducts_supply_surface_area] != Constants.Auto
+      ducts_supply_surface_area = args[:ducts_supply_surface_area]
     end
 
-    ducts_return_location = args[:ducts_return_location]
-    if ducts_return_location == Constants.Auto
-      ducts_return_location = get_duct_location_auto(args, hpxml)
+    if args[:ducts_return_surface_area] != Constants.Auto
+      ducts_return_surface_area = args[:ducts_return_surface_area]
```
shorowit, line 3431: How does this work? Isn't it a string and needs to be converted to a float?

shorowit: And now the get_duct_location_auto method can be removed, right?

joseph-robertson: I thought the hpxml.rb resource handled the string-to-float conversion?

joseph-robertson: Yes, I will remove the get_duction_location_auto method.

shorowit: Oh, right. Good point. 🤦‍♂️

BuildResidentialHPXML/measure.rb
```diff
@@ -0,0 +1,4033 @@
...
+        return HPXML::LocationLivingSpace
+      end
+    elsif ['vented attic'].include? space_type
+      return HPXML::LocationAtticVented
+    elsif ['unvented attic'].include? space_type
+      return HPXML::LocationAtticUnvented
+    elsif ['unconditioned basement'].include? space_type
+      return HPXML::LocationBasementUnconditioned
+    elsif ['corridor'].include? space_type
+      return HPXML::LocationLivingSpace # FIXME: update to handle new enum
```
shorowit, line 2887: Seems like this can be updated to HPXML::LocationOtherHousingUnit now?

tasks.rb
```diff
@@ -185,6 +190,8 @@ def create_osws
     'base-hvac-furnace-wood-only.osw' => 'base.osw',
     # 'base-hvac-furnace-x3-dse.osw' => 'base.osw', # Not going to support DSE
     'base-hvac-ground-to-air-heat-pump.osw' => 'base.osw',
+    # 'base-hvac-ground-to-air-heat-pump-cooling-only.osw' => 'base-hvac-ground-to-air-heat-pump.osw', # FIXME
```
shorowit, line 193: What's the problem with this one?

joseph-robertson: https://github.com/NREL/OpenStudio-HPXML/blob/master/workflow/sample_files/base-hvac-ground-to-air-heat-pump-cooling-only.xml#L313

shorowit: Ah, that's a bug. Will fix it.

shorowit: https://github.com/NREL/OpenStudio-HPXML/pull/611

BuildResidentialHPXML/measure.rb
```diff
@@ -3892,10 +3892,8 @@ def self.set_cooling_systems(hpxml, runner, args)
 
     return if cooling_system_type == 'none'
 
-    if cooling_system_type != HPXML::HVACTypeEvaporativeCooler
-      if args[:cooling_system_cooling_capacity] != Constants.Auto
-        cooling_capacity = args[:cooling_system_cooling_capacity]
-      end
+    if args[:cooling_system_cooling_capacity] != Constants.Auto
```
shorowit, line 3895: Also need to update the description for the cooling_system_cooling_capacity argument.

BuildResidentialHPXML/measure.rb
```diff
@@ -1766,7 +1772,11 @@ def arguments(model)
     arg.setDisplayName('Water Heater: First Hour Rating')
     arg.setDescription("Rated gallons of hot water supplied in an hour. Required if Efficiency Type is UniformEnergyFactor and Type is not #{HPXML::WaterHeaterTypeTankless}. Does not apply to space-heating boilers.")
     arg.setUnits('gal/hr')
-    arg.setDefaultValue(56.0)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('water_heater_usage_bin', water_heater_usage_bin_choices, false)
+    arg.setDisplayName('Water Heater: Usage Bin')
+    arg.setDescription("The usage of the water heater. Required if Efficiency Type is UniformEnergyFactor and Type is not #{HPXML::WaterHeaterTypeTankless}. Does not apply to space-heating boilers.")
```
shorowit, line 1779: If you’re adding this argument, then remove the FHR one above.

joseph-robertson: Can we not have the flexibility of providing either option?

shorowit: They are both equally available inputs (on the EnergyGuide label), so there’s really no advantage to that.

joseph-robertson: Which of the two do you think we should provide?

shorowit: I'd recommend providing just Usage Bin (which can be defaulted in OS-HPXML).

BuildResidentialHPXML/resources/constants.rb
```diff
@@ -48,12 +48,4 @@ def self.NumHoursInYear(is_leap_year = false)
     num_hours_in_year = num_days_in_year * 24
     return num_hours_in_year.to_f
   end
-
-  def self.PeakFlowRate
```
joseph-robertson: Hmm, I thought there was a specific reason for using these. @rajeee would probably know.

shorowit: I spoke with @rajeee and my understanding is that Constants provide values in the schedule csv that have some physical basis (i.e., power or flowrate values match the raw data that the schedules are derived from). However, that is a bit misleading in the context of the simulation because the schedule values are only used to calculate _when_ energy is used, not _how much_. It's also confusing to mix units between columns when, at the end of the day, they are treated as fractional inputs to the simulation engine.

Additionally, the values should have no bearing on the simulation results, but testing showed that they actually did. This is because other code in our model generation (specifically the dryer vent model) assumes that schedules are normalized to 1. In EnergyPlus, that convention makes the most sense because then the equipment design levels or peak flow rates are closer to reality.

So in summary, I think it's more intuitive and less error-prone to use normalized columns everywhere in the context of EnergyPlus.

joseph-robertson:
@shorowit [This line](https://github.com/NREL/OpenStudio-HPXML/blob/build-res-hpxml-v3/HPXMLtoOpenStudio/resources/hvac.rb#L960) is a problem for when schedules type is stochastic and there are ceiling fans present.

PR #379, opened 2020-05-04
BuildResidentialHPXML measure: Model "second" (parallel) heating system

joseph-robertson:
## Pull Request Description

https://trello.com/c/5voHNmZK/35-implement-secondary-heating-in-buildresidentialhpxml

Added the following arguments to BuildResidentialHPXML:
* heating_system_type_2 (wall/floor furnace, electric resistance, stove, portable heater, fireplace)
* heating_system_fuel_2 (any)
* heating_system_heating_efficiency_2
* heating_system_heating_capacity_2
* heating_system_fraction_heat_load_served_2 (>50% gives a warning)
* heating_system_electric_auxiliary_energy_2

## Checklist

Not all may apply:


BuildResidentialHPXML/measure.rb
```diff
@@ -1219,6 +1219,60 @@ def arguments(model)
...
+    arg.setDescription('The rated efficiency value of the second Furnace/WallFurnace heating system.')
+    arg.setDefaultValue(0.78)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_heating_efficiency_percent_2', true)
+    arg.setDisplayName('Heating System 2: Rated Percent')
+    arg.setUnits('Percent')
+    arg.setDescription('The rated efficiency value of the second ElectricResistance/Stove/PortableHeater/Fireplace heating system.')
+    arg.setDefaultValue(1.0)
+    args << arg
```
shorowit, line 1255: Since AFUE/Percent have the same units, I suggest combining these into a single `heating_system_heating_efficiency_2` input. No point introducing extra complexity. Then in the code further down, automatically use AFUE or Percent as appropriate based on the heating system type. You can make the same change for the primary heating system.

BuildResidentialHPXML/measure.rb
```diff
@@ -1219,6 +1219,60 @@ def arguments(model)
...
+
+    arg = OpenStudio::Measure::OSArgument::makeStringArgument('heating_system_heating_capacity_2', true)
+    arg.setDisplayName('Heating System 2: Heating Capacity')
+    arg.setDescription("The output heating capacity of the second heating system. If using '#{Constants.Auto}', the autosizing algorithm will use ACCA Manual S to set the capacity.")
```
shorowit, line 1259: I suggest rewording to "the autosizing algorithm will use ACCA Manual J/S to set the capacity to meet its load served." That will help a user understand that a system meeting 50% of the heating load will be autosized to 50% of the capacity it would have if it met 100% of the heating load. Make the same change for the primary HVAC system.

BuildResidentialHPXML/measure.rb
```diff
@@ -1219,6 +1219,60 @@ def arguments(model)
...
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_fraction_heat_load_served_2', true)
+    arg.setDisplayName('Heating System 2: Fraction Heat Load Served')
+    arg.setDescription('The heat load served fraction of the second heating system.')
+    arg.setUnits('Frac')
+    arg.setDefaultValue(0)
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_electric_auxiliary_energy_2', false)
+    arg.setDisplayName('Heating System 2: Electric Auxiliary Energy')
+    arg.setDescription('The electric auxiliary energy of the second heating system.')
```
shorowit, line 1273: Units are kWh/yr.

BuildResidentialHPXML/measure.rb
```diff
@@ -3260,6 +3321,14 @@ def validate_arguments(args)
     error = ((args[:ducts_supply_location] == Constants.Auto) && (args[:ducts_supply_surface_area] != Constants.Auto)) || ((args[:ducts_supply_location] != Constants.Auto) && (args[:ducts_supply_surface_area] == Constants.Auto)) || ((args[:ducts_return_location] == Constants.Auto) && (args[:ducts_return_surface_area] != Constants.Auto)) || ((args[:ducts_return_location] != Constants.Auto) && (args[:ducts_return_surface_area] == Constants.Auto))
     errors << "ducts_supply_location=#{args[:ducts_supply_location]} and ducts_supply_surface_area=#{args[:ducts_supply_surface_area]} and ducts_return_location=#{args[:ducts_return_location]} and ducts_return_surface_area=#{args[:ducts_return_surface_area]}" if error
 
+    # no second heating system, but has positive heat load served fraction
+    error = (args[:heating_system_type_2] == 'none') && (args[:heating_system_fraction_heat_load_served_2] > 0)
+    errors << "heating_system_type_2=none and heating_system_fraction_heat_load_served_2=#{args[:heating_system_fraction_heat_load_served_2]}" if error
```
shorowit, line 3326: Seems like this could just be a warning or even ignored altogether. If the system type is none, I don't think it should matter what the other inputs are.

BuildResidentialHPXML/measure.rb
```diff
@@ -3970,14 +4039,55 @@ def self.set_heating_systems(hpxml, runner, args)
       heating_efficiency_percent = args[:heating_system_heating_efficiency_percent]
     end
 
+    fraction_heat_load_served = args[:heating_system_fraction_heat_load_served]
+    fraction_heat_load_served_2 = args[:heating_system_fraction_heat_load_served_2]
+    fraction_heat_load_served -= fraction_heat_load_served_2
```
shorowit, line 4044: I'm totally confused. Why are you subtracting heat_load_served_2 from heat_load_served?

BuildResidentialHPXML/measure.rb
```diff
@@ -3970,14 +4039,55 @@ def self.set_heating_systems(hpxml, runner, args)
...
                               heating_efficiency_afue: heating_efficiency_afue,
                               heating_efficiency_percent: heating_efficiency_percent)
+
+    heating_system_type_2 = args[:heating_system_type_2]
+
+    return if heating_system_type_2 == 'none'
+
+    heating_capacity_2 = args[:heating_system_heating_capacity_2]
+    if heating_capacity_2 == Constants.Auto
+      heating_capacity_2 = -1
```
shorowit, line 4061: You can just set all autosized heating capacities to `nil` now instead of -1 (though -1 still works for backwards compatibility). Excluding the element is now preferred to setting it as -1.

BuildResidentialHPXML/measure.rb
joseph-robertson: I thought there was a reason we had them separated out. So that different heating system types would get appropriate default values? Why couldn't we just have a single efficiency argument for cooling (even though they're different units), and then set them according to the cooling system type?

BuildResidentialHPXML/measure.rb
joseph-robertson: For example: if the user left the primary at 100% and then entered a secondary system at 40%, this ensures that the total load fraction is 1.0.

shorowit: You are thinking of EnergyPlus's sequential load fractions, which are defined differently than HPXML's fractions load served. The user should enter 60% and 40% for that situation and they should be directly translated to the HPXML file.

BuildResidentialHPXML/measure.rb
shorowit: For cooling system units or heat pump units (SEER/EER, COP/HSPF), the logic definitely makes sense to have different values for different system types because if the default SEER is 13, that wouldn't make sense as a EER value. They are defined differently. In fact, this need to be fixed because you have cooling_system_cooling_efficiency_eer set as 13.

But for heating system units (AFUE/Percent), they are essentially the same. Most users would probably be more confused by treating them differently than vice versa. Also note that the current defaults don't make sense. You have 0.78 for furnaces and 1.0 for other systems. But those default values should be based on _fuel type_, not _system type_. Electric systems will be ~1.0 while fuel systems will generally be in the 0.78 - 0.95 range. Since you have the default fuel type as electric, 0.78 doesn't make sense for an electric furnace. So if you default to electricity, a default of 1.0 can be used for all heating system types; if you default to gas, a default of 0.78 can be used for all heating system types. This needs to be fixed for both primary and secondary heating systems.

BuildResidentialHPXML/measure.rb
joseph-robertson: So should an error be thrown if the two fractions don't add up to 1.0?

shorowit: An error will be issued if they try to use the HPXML file in the HPXMLtoOpenStudio measure. So... I guess not?

PR #382, opened 2020-05-07
HPXML class datatype conversions

shorowit:
Improves datatype conversions used by HPXML class. In particular, prevents automatically converting non-integer values to integers.

Checked that the appropriate errors are raised for float, integer, and boolean, either when reading an HPXML file or writing to an HPXML file.


HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -4261,3 +4243,53 @@ def self.add_extension(parent:,
...
+    fail "Cannot convert '#{value}' to integer."
+  end
+end
+
+def to_boolean(value)
+  if value.is_a? TrueClass
+    return true
+  elsif value.is_a? FalseClass
+    return false
+  elsif (value.downcase.to_s == 'true') || (value == '1') || (value == 1)
```
joseph-robertson, line 4273: 👍

PR #385, opened 2020-05-13
FrameFloors wrt/ MF space

shorowit:
## Pull Request Description

For FrameFloors, removes the "other housing unit above" and "other housing unit below" enumerations. Instead, FrameFloors in MF/attached buildings that are adjacent to any of the "other" space types (i.e., "other housing unit", "other heated space", "other multifamily buffer space", and "other non-freezing space") must now have an `extension/OtherSpaceAboveOrBelow` property set to either "above" or "below".

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/measure.rb
+++ b/HPXMLtoOpenStudio/measure.rb
@@ -3996,10 +3996,8 @@ class OSModel
     elsif [HPXML::LocationBasementConditioned].include? exterior_adjacent_to
       surface.createAdjacentSurface(create_or_get_space(model, spaces, HPXML::LocationLivingSpace))
       @cond_bsmnt_surfaces << surface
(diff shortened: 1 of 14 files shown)


shorowit:
@yzhou601 I realized that we essentially needed "above" and "below" versions of all the new MF space types (like we have for "other housing unit") in order to determine if a FrameFloor is a floor or ceiling surface. So instead of creating a bunch of additional enumerations, I created an extension element that can apply to all the MF space types.

shorowit:
(The reason we don't need "above" and "below" versions of other space types is that we can usually infer whether it's a ceiling or floor. For example, it's a ceiling if the space type is an attic and it's a floor if the space type is a crawlspace/basement/garage. But the MF space types are generic, so it's impossible to infer.)

PR #400, opened 2020-05-23
Other side coefficient related improvement

yzhou601:
## Pull Request Description

Fix some small issues related to the use of SurfaceProperty:OtherSideCoefficients object for mf spaces

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -4045,16 +4045,16 @@ def self.set_surface_interior(model, spaces, surface, interior_adjacent_to)
...
-    elsif [HPXML::LocationGround].include? exterior_adjacent_to
+    elsif exterior_adjacent_to == HPXML::LocationGround
       surface.setOutsideBoundaryCondition('Foundation')
-    elsif [HPXML::LocationBasementConditioned].include? exterior_adjacent_to
+    elsif exterior_adjacent_to == HPXML::LocationOtherHousingUnit
+      surface.setOutsideBoundaryCondition('Adiabatic')
+    elsif exterior_adjacent_to == HPXML::LocationBasementConditioned
       surface.createAdjacentSurface(create_or_get_space(model, spaces, HPXML::LocationLivingSpace))
       @cond_bsmnt_surfaces << surface
-      set_surface_otherside_coefficients(surface, exterior_adjacent_to, model, spaces)
```
yzhou601, line 4055: I think this line is not intended here, might be added by error?

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -244,6 +244,7 @@ def initialize(hpxml_path: nil, collapse_enclosure: true)
     # Clean up
     delete_partition_surfaces()
     delete_tiny_surfaces()
+    delete_adiabatic_subsurfaces()
```
yzhou601, line 247: I ended up implementing adiabatic door deletion inside hpxml class scope. In this way the net area of its attached wall won't be messed up. I have run the other housing unit test file for confirmation.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -4094,10 +4093,11 @@ def self.get_space_temperature_schedule(model, location, spaces)
 
     sch = OpenStudio::Model::ScheduleConstant.new(model)
     sch.setName(location)
+    heating_setpoint = @hpxml.hvac_controls[0].heating_setpoint_temp
```
yzhou601, line 4096: This simply grabs base setpoint temperature without considering setback. @shorowit Good enough?

HPXMLtoOpenStudio/resources/hpxml.rb
shorowit: This is clever, but I worry that it might cause [these error checks](https://github.com/NREL/OpenStudio-HPXML/blob/d26e36c45415d8057c819f45319b3ef9a2bd0504/HPXMLtoOpenStudio/measure.rb#L310-L374) to get triggered? Maybe you can call hpxml.delete_adiabatic_subsurfaces() from the end of the linked method instead of in the HPXML constructor?

HPXMLtoOpenStudio/measure.rb
shorowit: Is it easy to use a sensor on the heating setpoint schedule value? If not, then this seems fine as an approximation.

HPXMLtoOpenStudio/resources/hpxml.rb
yzhou601: In the error checking, I didn't see it do anything with subsurfaces. The deletion only applied to subsurfaces, or more specifically, doors, along with its attachment to its parent surface. Since it didn't delete parent surface, the error-cheking should be fine. Also, I noticed that we also created hpxml class in our [test](https://github.com/NREL/OpenStudio-HPXML/blob/d26e36c45415d8057c819f45319b3ef9a2bd0504/workflow/tests/hpxml_translator_test.rb#L509), if we only implemented deletion in the measure, we might need to make changes to test codes as well to address it. So implementing in class constructor might be simpler.

shorowit: Oops! Disregard what I said. I previously read the method "delete_adiabatic_subsurfaces" as "delete_adiabatic_**surfaces**".

HPXMLtoOpenStudio/measure.rb
yzhou601: Unfortunately the sequence of calling add_setpoints method is after surface creation right before adding hvac, so the heating setpoint schedule is not there at this point. Simply grabbing heating setpoint from OS model is not working. But we can manually pass the `Constants.ObjectNameHeatingSetpoint` as sensor name, which is obviously not that safe.

shorowit: I can't think of any reason why we couldn't move the add_setpoints call to before the surface creation. Is that worth trying?

yzhou601: Sure, I can try changing its calling sequence.

yzhou601:
Before last commit, ci only shows diff on attached-multifamily and other-housing-unit. 
After changing other-heated-space heating setpoint, expect diff on other-heated-space as well. The other heated space schedule hourly values are manually tested, the minimum values are dynamically changed with heating setpoint temperature. Ready for review.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -3652,10 +3653,17 @@ def self.get_space_temperature_schedule(model, location, spaces)
...
```
shorowit, line 3656: Seems like this variable is now unused, right?

yzhou601: Oops, I thought I deleted it. Good catch.

shorowit:
CI results are as you described. I'm going to revert the change to base-enclosure-other-heated-space.xml now that we've shown it's working correctly (I'd prefer that it stay more consistent with the other base-enclosure-foo.xml files) and then will merge.

shorowit:
@yzhou601 Thanks for getting this implemented so quickly.

PR #407, opened 2020-06-01
Multiple VentilationFans

yzhou601:
## Pull Request Description

Allows multiple `VentilationFan` objects for:
- Whole building ventilation (mechanical ventilation)
- Local ventilation (kitchen/bath fans)
- Cooling load reduction (whole house fans)

## Checklist

Not all may apply:


workflow/tests/hpxml_translator_test.rb
```diff
@@ -1002,46 +995,56 @@ def _verify_simulation_outputs(runner, rundir, hpxml_path, results)
...
-        assert_in_delta(hpxml_value, sql_value, 0.01)
+          # CFIS Fan power
+          hpxml_value = fan_cfis[0].fan_power
+          query = "SELECT Value FROM ReportData WHERE ReportDataDictionaryIndex IN (SELECT ReportDataDictionaryIndex FROM ReportDataDictionary WHERE Name='#{@cfis_fan_power_output_var.variableName}' AND ReportingFrequency='Run Period')"
+          sql_value = sqlFile.execAndReturnFirstDouble(query).get
+          assert_in_delta(hpxml_value, sql_value, 0.01)
+          mv_energy -= cfis_energy
+        end
 
-        # CFIS Flow rate
```
yzhou601, line 1027: Deleted this test, QWHV already got tested in test_airflow.

yzhou601:
Only CI diff is the new test file, do I need to revert changes to bath/kitchen fan and whole house fan?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1480,174 +1705,7 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
-      # Fan power is metered under fan cooling and heating meters
-      infil_program.addLine("    Set #{vent_mech_fan_actuator.name} = 0")
-      infil_program.addLine('  EndIf')
-      infil_program.addLine('Else')
-      # The ventilation requirement for the hour has been met
-      infil_program.addLine('  Set QWHV = 0')
-      infil_program.addLine("  Set #{vent_mech_fan_actuator.name} = 0")
-      infil_program.addLine('EndIf')
-    else # Non-CFIS system
-      infil_program.addLine("Set QWHV = #{UnitConversions.convert(vent_mech_cfm.to_f, 'cfm', 'm^3/s').round(4)}")
```
yzhou601, line 1541: Small ci diff in erv/hrv systems is caused by removing the rounding here. Not a big issue applying rounding or not.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1166,33 +1163,34 @@ def self.apply_local_ventilation(model, vent_object, obj_name)
...
+      obj_sch_sensor.setKeyName(obj_sch.schedule.name.to_s)
+      obj_sch_sensors[vent_object.id] = obj_sch_sensor
+
+      equip_def = OpenStudio::Model::ElectricEquipmentDefinition.new(model)
+      equip_def.setName(obj_name)
+      equip = OpenStudio::Model::ElectricEquipment.new(equip_def)
+      equip.setName(obj_name)
+      equip.setSpace(@living_space)
+      if obj_type_name == Constants.ObjectNameMechanicalVentilationBathFan
+        equip_def.setDesignLevel(vent_object.fan_power * vent_object.quantity)
```
yzhou601, line 1178: Bath fan test ci diff is expected, the "quantity" is applied here as a bugfix.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -337,18 +302,24 @@ def self.apply_natural_ventilation_and_whole_house_fan(model, weather, vent_whf,
       vent_program.addLine("Set Tnvsp = #{UnitConversions.convert(73.0, 'F', 'C')}") # Assumption when no HVAC system
     end
     vent_program.addLine("Set NVavail = #{nv_avail_sensor.name}")
-    vent_program.addLine("Set WHFavail = #{whf_avail_sensor.name}")
     vent_program.addLine("Set ClgSsnAvail = #{nv_clg_ssn_sensor.name}")
     vent_program.addLine('If (Wout < MaxHR) && (Phiout < MaxRH) && (Tin > Tout) && (Tin > Tnvsp) && (ClgSsnAvail > 0)')
-    vent_program.addLine("  Set WHF_Flow = #{UnitConversions.convert(whf_cfm, 'cfm', 'm^3/s')}")
+    vent_program.addLine('  Set WHF_Flow = 0')
+    vent_fans_whf.each do |vent_whf|
+      vent_program.addLine("Set WHF_Flow = WHF_Flow + #{UnitConversions.convert(vent_whf.rated_flow_rate, 'cfm', 'm^3/s')} * #{whf_avail_sensors[vent_whf.id].name}")
```
shorowit, line 309: Even though it doesn't make it to the IDF, please try to indent the EMS lines to show how nested they are. I.e., `"Set WHF_Flow ...` -> `"  Set WHF_Flow`

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -337,18 +302,24 @@ def self.apply_natural_ventilation_and_whole_house_fan(model, weather, vent_whf,
...
     vent_program.addLine('  Set Adj = (@Max Adj 0)')
-    vent_program.addLine('  If (WHFavail > 0) && (WHF_Flow > 0)') # If available, prioritize whole house fan
+    vent_program.addLine('  If (WHF_Flow > 0)') # If available, prioritize whole house fan
     vent_program.addLine("    Set #{nv_flow_actuator.name} = 0")
     vent_program.addLine("    Set #{whf_flow_actuator.name} = WHF_Flow*Adj")
     vent_program.addLine("    Set #{liv_to_zone_flow_rate_actuator.name} = WHF_Flow*Adj") unless whf_zone.nil?
-    vent_program.addLine("    Set #{whf_elec_actuator.name} = #{whf_fan_w}*Adj")
+    vent_program.addLine('  Set WHF_W = 0')
+    vent_fans_whf.each do |vent_whf|
+      vent_program.addLine("Set WHF_W = WHF_W + #{vent_whf.fan_power} * #{whf_avail_sensors[vent_whf.id].name}")
```
shorowit, line 320: Both of these lines need proper indentation.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -190,59 +189,33 @@ def self.apply_infiltration_to_unconditioned_space(model, space, ach = nil, ela
...
+
+    # Availability Schedules paired with vent fan class
+    # If whf_num_days_per_week is exposed, can handle multiple fans with different days of operation
+    whf_avail_sensors = {}
+    vent_fans_whf.each do |vent_whf|
       whf_num_days_per_week = 7 # FUTURE: Expose via HPXML?
-      whf_cfm = vent_whf.rated_flow_rate
-      whf_fan_w = vent_whf.fan_power
-    end
+      obj_name = "#{Constants.ObjectNameWholeHouseFan} #{vent_whf.id}"
```
shorowit, line 212: I'm a little concerned that these names could become really long (or have invalid characters?) since we're not in control of them, as they come from the user's HPXML file. How about if we just use index values (1, 2, 3) instead of ids?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -423,46 +414,50 @@ def self.create_sens_lat_load_actuator_and_equipment(model, name, space, frac_la
...
-    @cfis_t_sum_open_var = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, "#{Constants.ObjectNameMechanicalVentilation.gsub(' ', '_')}_cfis_t_sum_open") # Sums the time during an hour the CFIS damper has been open
-    @cfis_f_damper_extra_open_var = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, "#{Constants.ObjectNameMechanicalVentilation.gsub(' ', '_')}_cfis_f_extra_damper_open") # Fraction of timestep the CFIS blower is running while hvac is not operating. Used by infiltration and duct leakage programs
+      @cfis_t_sum_open_var[vent_mech.id] = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, "#{Constants.ObjectNameMechanicalVentilation.gsub(' ', '_')}_cfis_t_sum_open_#{vent_mech.id.gsub(' ', '_')}") # Sums the time during an hour the CFIS damper has been open
+      @cfis_f_damper_extra_open_var[vent_mech.id] = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, "#{Constants.ObjectNameMechanicalVentilation.gsub(' ', '_')}_cfis_f_extra_damper_open_#{vent_mech.id.gsub(' ', '_')}") # Fraction of timestep the CFIS blower is running while hvac is not operating. Used by infiltration and duct leakage programs
```
shorowit, line 443: Use indexes instead of ids here too and review rest of code for similar. (These in particular make me more concerned than the previous example because I think EMS has a limit on the number of characters in a line?)

shorowit: We change to use indexes, then maybe these would become arrays instead of hashes?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -423,46 +414,50 @@ def self.create_sens_lat_load_actuator_and_equipment(model, name, space, frac_la
...
 
-    # CFIS Initialization Program
-    cfis_program = OpenStudio::Model::EnergyManagementSystemProgram.new(model)
-    cfis_program.setName(Constants.ObjectNameMechanicalVentilation + ' cfis init program')
-    cfis_program.addLine("Set #{@cfis_t_sum_open_var.name} = 0")
-    cfis_program.addLine("Set #{@cfis_f_damper_extra_open_var.name} = 0")
+      # CFIS Initialization Program
+      cfis_program = OpenStudio::Model::EnergyManagementSystemProgram.new(model)
+      cfis_program.setName(Constants.ObjectNameMechanicalVentilation + ' cfis init program')
```
shorowit, line 447: Add an index to the name?

HPXMLtoOpenStudio/resources/airflow.rb
shorowit: 😮 Good catch!

shorowit: How about allowing an optional quantity input for kitchen fans too? That would simplify our code so we don't have to handle the bath fan as a special case. 99.9% of the time it will be 1, but that's okay.

Maybe for completeness, we should allow optional quantity inputs for all vent fan types?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1273,7 +1271,119 @@ def self.calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
...
-  def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent_kitchen, vent_bath, has_flue_chimney, air_infils, vented_attic, vented_crawl)
+  def self.get_mech_vent_cfm(vent_mech)
+    if not vent_mech.tested_flow_rate.nil?
+      return vent_mech.tested_flow_rate
+    else
+      return vent_mech.rated_flow_rate
+    end
+  end
+
+  def self.apply_cfis(infil_program, vent_mech, cfis_fan_actuator)
```
shorowit, line 1282: How about something more descriptive like apply_cfis_to_infil_program?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1273,7 +1271,119 @@ def self.calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
     return vent_mech_sens_eff, vent_mech_lat_eff, vent_mech_apparent_sens_eff
   end
 
-  def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent_kitchen, vent_bath, has_flue_chimney, air_infils, vented_attic, vented_crawl)
+  def self.get_mech_vent_cfm(vent_mech)
```
shorowit, line 1274: Maybe we should add a `flow_rate` method to the HPXML VentilationFan class that does this? Then we could always just call that method for every VentilationFan object without worrying about whether it was considered a tested or rated value.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: Good catch, adding id into the name is good for debugging as you can see if things in idf are consistent with HPXML, but yeah it might introduce issues.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1265,7 +1281,111 @@ def self.calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
...
+    infil_program.addLine("      Set #{@cfis_t_sum_open_var[vent_mech.id].name} = CFIS_t_min_hr_open")
+    infil_program.addLine('    Else')
+    # Damper is open and using call for heat/cool to supply fresh air
+    infil_program.addLine('      Set cfis_fan_runtime = fan_rtf_hvac*ZoneTimeStep*60')
+    infil_program.addLine('      Set cfis_f_damper_open = fan_rtf_hvac')
+    infil_program.addLine('      Set QWHV_cfis = QWHV_cfis + cfis_f_damper_open * CFIS_Q_duct')
+    infil_program.addLine("      Set #{@cfis_t_sum_open_var[vent_mech.id].name} = #{@cfis_t_sum_open_var[vent_mech.id].name}+cfis_fan_runtime")
+    infil_program.addLine('    EndIf')
+    # Fan power is metered under fan cooling and heating meters
+    infil_program.addLine("    Set #{cfis_fan_actuator.name} = #{cfis_fan_actuator.name}")
```
shorowit, line 1328: Is this needed?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1265,7 +1281,111 @@ def self.calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
...
+    infil_program.addLine('  EndIf')
+    infil_program.addLine('Else')
+    # The ventilation requirement for the hour has been met
+    infil_program.addLine("  Set #{cfis_fan_actuator.name} = #{cfis_fan_actuator.name}")
```
shorowit, line 1332: Is this needed?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1315,31 +1413,47 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+    model.getBuilding.additionalProperties.setFeature(Constants.SizingInfoMechVentWholeHouseRateBalanced, tot_bal_cfm)
+    model.getBuilding.additionalProperties.setFeature(Constants.SizingInfoMechVentWholeHouseRateUnbalanced, unbal_cfm.abs)
+    model.getBuilding.additionalProperties.setFeature(Constants.SizingInfoMechVentExist, (not vent_fans_mech.empty?))
+
+    # Fan Actuators
+    # Do we still want to create ee if there's no fan in HPXML? Reporting purpose?
+    sup_fan_actuator = add_ee_for_vent_fan_power(model, Constants.ObjectNameMechanicalVentilationHouseFanSupply, HPXML::MechVentTypeSupply)
+    exh_fan_actuator = add_ee_for_vent_fan_power(model, Constants.ObjectNameMechanicalVentilationHouseFanExhaust, HPXML::MechVentTypeExhaust)
+    bal_fan_actuator = add_ee_for_vent_fan_power(model, Constants.ObjectNameMechanicalVentilationHouseFanBalanced, HPXML::MechVentTypeBalanced)
+    cfis_fan_actuator = add_ee_for_vent_fan_power(model, Constants.ObjectNameMechanicalVentilationHouseFanCFIS, HPXML::MechVentTypeCFIS)
```
shorowit, line 1456: I suggest simplifying this. At the end of the day, we need: 
1) one EE object for CFIS systems (since it has FractionLost=0.0 and needs to be actuated), and
2) one EE object for all non-CFIS systems (where we don't need to use EMS/actuator and we can calculate FractionLost from the various mech vent system types and their fan powers)

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1359,6 +1473,127 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+      infil_program.addLine('Set ERVSecInEnth = (@HFnTdbW ERVSecInTemp ERVSecInW)')
+    end
+
+    # Calculate cfm weighted average effectiveness for hvac sizing
+    # Please review this, and its integration with hvac_sizing
+    weighted_vent_mech_tot_eff = 0.0
+    weighted_vent_mech_lat_eff = 0.0
+    weighted_vent_mech_apparent_sens_eff = 0.0
+
+    # Apply balanced mechanical ventilation
```
shorowit, line 1517: It looks like this is actually only applying ERV/HRV load, not handling balanced (w/o recovery) systems, right?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1359,6 +1473,127 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+    vent_mech_erv_hrv.each do |vent_mech|
+      vent_mech_cfm = vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0)
+      vent_mech_fan_w = vent_mech.fan_power * (vent_mech.hours_in_operation / 24.0)
+      vent_mech_sens_eff, vent_mech_lat_eff, vent_mech_apparent_sens_eff = calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
+      # No need to add balanced system here to average because their effectivenesses are all 0.0
+      weighted_vent_mech_tot_eff += vent_mech_cfm / tot_bal_cfm * vent_mech.total_recovery_efficiency.to_f
+      weighted_vent_mech_lat_eff += vent_mech_cfm / tot_bal_cfm * vent_mech_lat_eff
+      weighted_vent_mech_apparent_sens_eff += vent_mech_cfm / tot_bal_cfm * vent_mech_apparent_sens_eff
+
+      infil_program = apply_balanced_mech_vent_to_infil_program(model, infil_program, vent_mech_cfm, vent_mech_sens_eff, vent_mech_lat_eff, erv_sens_load_actuator, erv_lat_load_actuator)
```
shorowit, line 1527: Rename to `apply_erv_hrv_load_to_infil_program`?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1315,31 +1413,47 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+    # get cfms
+    sup_cfm = vent_mech_sup.map { |vent_mech| vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0) }.inject(0.0, :+)
+    exh_cfm = vent_mech_exh.map { |vent_mech| vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0) }.inject(0.0, :+)
+    bal_cfm = vent_mech_bal.map { |vent_mech| vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0) }.inject(0.0, :+)
+    erv_hrv_cfm = vent_mech_erv_hrv.map { |vent_mech| vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0) }.inject(0.0, :+)
+    cfis_cfm = vent_mech_cfis.map { |vent_mech| vent_mech.flow_rate }.inject(0.0, :+)
+
+    # Combine exhaust and supply
+    combined_bal_cfm = [sup_cfm, exh_cfm].min
+    combined_unbal_cfm = sup_cfm - exh_cfm
```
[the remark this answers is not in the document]
yzhou601: It is not supposed to apply `.abs` at this point, because it is added with cfis later, which is supply. The `.abs` was applied when creating the additional properties feature.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: Right, balanced without recovery is applied later in the airflow calculation(where only erv/hrv airlow is subtracted).

HPXMLtoOpenStudio/resources/airflow.rb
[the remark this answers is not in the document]
shorowit: Ah, true enough. I didn't look closely at how it was used, just noticed that `combined_unbal_cfm` could go negative which seemed wrong.

HPXMLtoOpenStudio/resources/airflow.rb
shorowit: Please update the comment then.

HPXMLtoOpenStudio/resources/airflow.rb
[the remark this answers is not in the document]
yzhou601: I see.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: Ah got it. Will do.

HPXMLtoOpenStudio/tests/test_airflow.rb
```diff
@@ -211,16 +269,84 @@ def test_ventilation_bath_kitchen_fans
...
+    assert_in_epsilon(1, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationRangeFan).size, 0.01)
+    assert_in_epsilon(kitchen_fan_power, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationRangeFan)[0].designLevel.get, 0.01)
+    assert_in_epsilon(1.0, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationRangeFan)[0].fractionLost, 0.01)
     assert_in_epsilon(bath_fan_cfm, UnitConversions.convert(program_values['Qbath'].to_f, 'm^3/s', 'cfm'), 0.01)
+    assert_in_epsilon(1, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationBathFan).size, 0.01)
+    assert_in_epsilon(bath_fan_power, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationBathFan)[0].designLevel.get, 0.01)
+    assert_in_epsilon(1.0, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationBathFan)[0].fractionLost, 0.01)
+  end
+
+  def test_multiple_mechvent
```
yzhou601, line 296: Multiple cfis systems are not as explicit as other systems for testing, is it ok just manually checking its EMS program?

shorowit: You can just follow what we did in the `test_mechanical_ventilation_cfis` test.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1315,31 +1409,51 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+    # Balanced and unbalanced airflow calculated for hvac sizing
+    tot_bal_cfm = bal_cfm + erv_hrv_cfm + combined_bal_cfm
+    unbal_cfm = combined_unbal_cfm + cfis_cfm
+
+    # Store info for HVAC Sizing measure
+    # Please review this, and its integration with hvac_sizing
```
shorowit, line 1450: Remove comment, this seems reasonable to me.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1359,6 +1473,123 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+      lat_name = "#{Constants.ObjectNameERVHRV} latent load"
+      erv_lat_load_var = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, lat_name.gsub(' ', '_'))
+      erv_lat_load_actuator = create_sens_lat_load_actuator_and_equipment(model, lat_name, @living_space, 1.0, 0.0)
+
+      infil_program.addLine("Set #{erv_sens_load_actuator.name} = 0.0")
+      infil_program.addLine("Set #{erv_lat_load_actuator.name} = 0.0")
+
+      # ERV/HRV EMS load model
+      # E+ ERV model is using standard density for MFR calculation, caused discrepancy with other system types.
+      # E+ ERV model also does not meet setpoint perfectly.
```
shorowit, line 1493: Add to comment that this may now be addressed? https://github.com/NREL/EnergyPlus/issues/7507

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1359,6 +1473,123 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_mech, vent
...
+    weighted_vent_mech_lat_eff = 0.0
+    weighted_vent_mech_apparent_sens_eff = 0.0
+
+    # Apply ERV/HRV mechanical ventilation
+    vent_mech_erv_hrv.each do |vent_mech|
+      vent_mech_cfm = vent_mech.flow_rate * (vent_mech.hours_in_operation / 24.0)
+      vent_mech_fan_w = vent_mech.fan_power * (vent_mech.hours_in_operation / 24.0)
+      vent_mech_sens_eff, vent_mech_lat_eff, vent_mech_apparent_sens_eff = calc_hrv_erv_effectiveness(vent_mech, vent_mech_cfm, vent_mech_fan_w)
+      # No need to add balanced system here to average because their effectivenesses are all 0.0
+      weighted_vent_mech_tot_eff += vent_mech_cfm / tot_bal_cfm * vent_mech.total_recovery_efficiency.to_f
```
shorowit, line 1523: I assume this works correctly if the HPXML input is AdjustedTotalRecoveryEfficiency instead of TotalRecoveryEfficiency? I can't immediately tell.

HPXMLtoOpenStudio/tests/test_airflow.rb
```diff
@@ -24,23 +24,42 @@ def get_ems_values(ems_objects, name)
...
       end
     end
     assert_operator(values.size, :>, 0)
     return values
   end
 
+  def get_eed_for_ventilation(model, ee_name)
+    eeds = []
+    model.getElectricEquipmentDefinitions.each do |eed|
+      next if eed.name.to_s.include? 'cfis'
```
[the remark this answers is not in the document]
yzhou601: This line handles the situation where cfis object is returned when we actually want the combined electric equipment object for all other whole house ventilation fans. This is because CFIS fan constant name is `mech vent house fan cfis`, while other whole house ventilation fan name is `mech vent house fan`, which is a substring of cfis name. 🤦 
I've tried out other approaches, eg. using equal("==") instead of "include?" but the bath/kitchen fans can be multiple, which ended up with `mech vent bath fan 1` etc. Therefore, in order to implement the minimum change, I ended up with two lines here.

HPXMLtoOpenStudio/tests/test_airflow.rb
```diff
@@ -24,23 +24,42 @@ def get_ems_values(ems_objects, name)
...
     args_hash = {}
     args_hash['hpxml_path'] = File.absolute_path(File.join(sample_files_dir, 'base.xml'))
     model, hpxml = _test_measure(args_hash)
 
     # Check infiltration/ventilation program
     program_values = get_ems_values(model.getEnergyManagementSystemPrograms, "#{Constants.ObjectNameInfiltration} program")
-    assert_in_epsilon(0.0436, Float(program_values['c']), 0.01)
-    assert_in_epsilon(0.0573, Float(program_values['Cs']), 0.01)
-    assert_in_epsilon(0.1446, Float(program_values['Cw']), 0.01)
+    assert_in_epsilon(0.0436, program_values['c'].sum, 0.01)
```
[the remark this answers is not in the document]
yzhou601: Yay, would you like me to try replacing all those `.inject()` here?

shorowit:
Hmm, something doesn't look quite right.

I ran `openstudio workflow/run_simulation.rb -x workflow/sample_files/base-mechvent-bath-kitchen-fans.xml --hourly ALL` and looked at the hourly mech vent airflow rates. They correctly showed the bath/kitchen fan spikes for certain hours.

Then I ran `openstudio workflow/run_simulation.rb -x workflow/sample_files/base-mechvent-multiple.xml --hourly ALL` and didn't see any bath/kitchen fan airflow spikes (i.e., every hour had the same value of 433 cfm) even though there are bath/kitchen fans defined in the HPXML.

Can you look at that?

HPXMLtoOpenStudio/tests/test_airflow.rb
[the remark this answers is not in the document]
shorowit: Assuming it's equivalent, sure! It's definitely more readable.

HPXMLtoOpenStudio/tests/test_airflow.rb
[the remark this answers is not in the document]
shorowit: Ah, I understand. Works for me.

yzhou601:
> Hmm, something doesn't look quite right.
> 
> I ran `openstudio workflow/run_simulation.rb -x workflow/sample_files/base-mechvent-bath-kitchen-fans.xml --hourly ALL` and looked at the hourly mech vent airflow rates. They correctly showed the bath/kitchen fan spikes for certain hours.
> 
> Then I ran `openstudio workflow/run_simulation.rb -x workflow/sample_files/base-mechvent-multiple.xml --hourly ALL` and didn't see any bath/kitchen fan airflow spikes (i.e., every hour had the same value of 433 cfm) even though there are bath/kitchen fans defined in the HPXML.
> 
> Can you look at that?

This is because `supply` is always larger than exhaust in the multiple test file, so Qfan is always equal to Qsupply, which won't have any spike. I can increase the exhaust cfms to better observe this behavior.

shorowit:
> This is because `supply` is always larger than `exhaust` in the multiple test file, so Qfan is always equal to Qsupply, which won't have any spike. I can increase the exhaust cfms to better observe this behavior.

Ah, right, this has to do with the ASHRAE 62.2 equations! Thanks, that explains it. I don't think the test file needs to be updated to show spikes (though if you've already made those changes, it's fine to commit them).

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: This line was deleted based on above comments.

PR #420, opened 2020-06-12
Schematron validation

bpark1327:
## Pull Request Description

Replace custom validation.rb files with Schematron.

Schematron is a rule-based validation language for making assertions about the presence or absence of patterns in XML trees. It is a structural schema language expressed in XML using a small number of elements and XPath.

## Checklist

Not all may apply:


bpark1327:
A sample Schematron EP_validator (i.e. EPvalidator.stron) has added.  And a ruby file (i.e. schematron_test.rb) has temporarily added in "resource" directory for testing the schematron validator.

HPXMLtoOpenStudio/resources/EPvalidator.stron
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
shorowit, line 1: Where did the .stron file extension come from? Seems like it's typically .sch or .xml.

HPXMLtoOpenStudio/resources/EPvalidator.stron
```diff
@@ -0,0 +1,14 @@
...
+<sch:schema xmlns:sch="http://purl.oclc.org/dsdl/schematron">
+  <sch:title>HPXML Schematron Validator</sch:title>
+  <sch:ns uri='http://hpxmlonline.com/2019/10' prefix='hpxml_v3'/>
```
shorowit, line 4: I suggest just using prefix='h' instead, since it's going to show up so often in the file.

HPXMLtoOpenStudio/resources/EPvalidator.stron
```diff
@@ -0,0 +1,31 @@
...
+  <sch:pattern>
+    <sch:title>EnergyPlus Validator</sch:title>
+    <sch:rule context="/hpxml_v3:HPXML/hpxml_v3:XMLTransactionHeaderInformation">
+      <sch:assert test="count(hpxml_v3:XMLType) = 1">element 'hpxml_v3:XMLType' is REQUIRED</sch:assert>
```
shorowit, line 8: Hmm, I kind of expected to just see "element 'XMLType' is REQUIRED" instead of "element 'hpxml_v3:XMLType' is REQUIRED". Not sure what the pros/cons are.

HPXMLtoOpenStudio/schematron_test.rb
```diff
@@ -0,0 +1,17 @@
+require 'pathname'
```
shorowit, line 1: This file should be HPXMLtoOpenStudio/tests/test_schematron.rb and it should be structured like all the other test files (i.e., have a class that inherits from MiniTest::Test. And then you can have proper asserts in the tests.

HPXMLtoOpenStudio/schematron_test.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'schematron-nokogiri'
+
+file_dir = File.dirname(__FILE__)
+
+# load the schematron xml
+stron_doc = Nokogiri::XML File.open(File.join(file_dir, 'resources', 'EPvalidator.stron'))  # "/path/to/schema.stron"
+# make a schematron object
+stron = SchematronNokogiri::Schema.new stron_doc
+# load the xml document you wish to validate
+xml_doc = Nokogiri::XML File.open(File.join(File.dirname(file_dir), 'workflow', 'sample_files', 'base.xml'))  # "/path/to/xml_document.xml"
```
shorowit, line 11: Seems like we should be looping through all sample files, not just testing the one file.

HPXMLtoOpenStudio/schematron_test.rb
```diff
@@ -0,0 +1,17 @@
...
+# validate it
+results = stron.validate xml_doc
+# print out the results
+results.each do |error|
+  puts "#{error[:line]}: #{error[:message]}"
```
shorowit, line 16: Convert to an `assert`

HPXMLtoOpenStudio/schematron_test.rb
```diff
@@ -0,0 +1,17 @@
...
+file_dir = File.dirname(__FILE__)
+
+# load the schematron xml
+stron_doc = Nokogiri::XML File.open(File.join(file_dir, 'resources', 'EPvalidator.stron'))  # "/path/to/schema.stron"
+# make a schematron object
+stron = SchematronNokogiri::Schema.new stron_doc
+# load the xml document you wish to validate
+xml_doc = Nokogiri::XML File.open(File.join(File.dirname(file_dir), 'workflow', 'sample_files', 'base.xml'))  # "/path/to/xml_document.xml"
+# validate it
+results = stron.validate xml_doc
```
shorowit, line 13: I think that basically every rule that's added to the schematron file should have a test that hits it. So, if the schematron file has:
```
<sch:rule context="/hpxml_v3:HPXML/hpxml_v3:Building/hpxml_v3:ProjectStatus">
      <sch:assert test="count(hpxml_v3:EventType) = 1">element 'hpxml_v3:ProjectStatus/hpxml_v3:EventType' is REQUIRED</sch:assert>
</sch:rule>
```
We should have a test where we get a hpxml object from base.xml, remove the ProjectStatus element, validate it against the schematron file, and assert that the schematron error message is generated.

HPXMLtoOpenStudio/resources/EPvalidator.stron
bpark1327: The former is likely to be easier to understand.

shorowit: There's only a single namespace in HPXML though, right? So I don't really see what additional information is being provided. I would guess that most users/developers are largely unaware that there's even a namespace. To that point, our current EPvalidator does not provide error messages that include the namespace.

Aside from the BuildingSync example, I'm not really finding any examples where the namespace is included in the error messages. I am leaning towards thinking we shouldn't include them.

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -743,7 +743,7 @@ def self.run_validator(hpxml_doc)
       },
 
       ## [LightingGroup]
-      'LightingGroup[LightingType[LightEmittingDiode | CompactFluorescent | FluorescentTube] and Location[text()="interior" or text()="exterior" or text()="garage"]]' => {
+      '/HPXML/Building/BuildingDetails/Lighting/LightingGroup[LightingType[LightEmittingDiode | CompactFluorescent | FluorescentTube] and Location[text()="interior" or text()="exterior" or text()="garage"]]' => {
```
shorowit, line 746: Good catch! Can you issue a separate PR with just this bugfix? I'd like to get this fix into master sooner than waiting on the full schematron capability.

shorowit:
@bpark1327 Can you pull the latest master into your branch and accommodate any changes to EPvalidator.rb that have been made?

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -0,0 +1,980 @@
...
+    <!-- [FoundationWallInsLayer] -->
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Enclosure/h:FoundationWalls/h:FoundationWall/h:Insulation/h:Layer[h:InstallationType="continuous - exterior" or h:InstallationType="continuous - interior"]'>
+      <sch:assert test="count(h:NominalRValue) = 1">Expected 1 element(s): NominalRValue</sch:assert>
+      <sch:assert test="count(h:extension/h:DistanceToTopOfInsulation) = 1">Expected 1 element(s): extension/DistanceToTopOfInsulation</sch:assert> <!-- ft -->
+      <sch:assert test="count(h:extension/h:DistanceToBottomOfInsulation) = 1">Expected 1 element(s): extension/DistanceToBottomOfInsulation</sch:assert> <!-- ft -->
+    </sch:rule>
+  </sch:pattern>
+
+  <sch:pattern name="[FrameFloorAdjacentToOther]">
+    <!-- [FrameFloorAdjacentToOther] -->
```
shorowit, line 557: Seems like we can remove all of these comments since it's identical to the pattern name on the line above it.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,618 @@
...
+      hpxml_doc = hpxml.to_oga()
+      _test_ruby_validation(hpxml_doc)
+      # Schematron validation
+      _test_schematron_validation(hpxml_path)
+    end
+    puts
+  end
+
+  def test_invalid_files
+    expected_error_msgs = {
```
shorowit, line 120: Call this expected_error_msgs_required, to contrast with expected_error_msgs_optional below.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,618 @@
...
+      ['/HPXML/Building/BuildingDetails/Appliances/ClothesDryer/FuelType[text()="natural gas" or text()="fuel oil" or text()="propane" or text()="electricity" or text()="wood" or text()="wood pellets"]'] => 'Expected 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/ClothesDryer/FuelType[text()="natural gas" or text()="fuel oil" or text()="fuel oil 1" or text()="fuel oil 2" or text()="fuel oil 4" or text()="fuel oil 5/6" or text()="diesel" or text()="propane" or text()="kerosene" or text()="coal" or text()="coke" or text()="bituminous coal" or text()="anthracite coal" or text()="electricity" or text()="wood" or text()="wood pellets"]',
+      ['/HPXML/Building/BuildingDetails/Appliances/Dishwasher/SystemIdentifier'] => 'Expected 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/Dishwasher/SystemIdentifier',
+      ['/HPXML/Building/BuildingDetails/Appliances/Dishwasher/RatedAnnualkWh'] => 'Expected 0 or 6 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/Dishwasher/RatedAnnualkWh | EnergyFactor | LabelElectricRate | LabelGasRate | LabelAnnualGasCost | LabelUsage | PlaceSettingCapacity',
```
shorowit, line 326: Why are there some rows like this that include "Expected 0", which implies they are optional, but they are located in the expected_error_msgs hash instead of expected_error_msgs_optional?

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,618 @@
...
+      ['/HPXML/Building/BuildingDetails/Appliances/ClothesWasher', nil] => 'Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/ClothesWasher',
+      ['/HPXML/Building/BuildingDetails/Appliances/ClothesDryer', nil] => 'Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/ClothesDryer',
+      ['/HPXML/Building/BuildingDetails/Appliances/Dishwasher', nil] => 'Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/Dishwasher',
+      ['/HPXML/Building/BuildingDetails/Appliances/Dehumidifier', nil] => 'Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/Dehumidifier',
+      ['/HPXML/Building/BuildingDetails/Appliances/CookingRange', nil] => 'Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/CookingRange',
+      ['/HPXML/Building/BuildingDetails/Appliances/Oven', nil] => 'Expected 1 element(s) for xpath: /HPXML/Building/BuildingDetails/Appliances/CookingRange/../Oven',
```
shorowit, line 371: Why are there some rows like this that include "Expected 1", which implies they are required, but they are located in the expected_error_msgs_optional hash instead of expected_error_msgs?

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -0,0 +1,980 @@
...
+    </sch:rule>
+    <sch:rule context="/h:HPXML/h:Building/h:BuildingDetails/h:BuildingSummary">
+      <sch:assert test="not(h:Site) or count(h:Site) = 1">Expected 0 or 1 element(s): Site</sch:assert> <!-- See [Site] -->
+      <sch:assert test="count(h:BuildingOccupancy) &lt;= 1">Expected 0 or 1 element(s): BuildingOccupancy</sch:assert> <!-- See [BuildingOccupancy]-->
+      <sch:assert test="count(h:BuildingConstruction) = 1">Expected 1 element(s): BuildingConstruction</sch:assert> <!-- See [BuildingConstruction] -->
+    </sch:rule>
+    <sch:rule context="/h:HPXML/h:Building/h:BuildingDetails/h:ClimateandRiskZones">
+      <sch:assert test="count(h:ClimateZoneIECC) &lt;= 1">Expected 0 or 1 element(s): ClimateZoneIECC</sch:assert> <!-- See [ClimateZone] -->
+      <sch:assert test="count(h:WeatherStation) = 1">Expected 1 element(s): WeatherStation</sch:assert> <!-- See [WeatherStation] -->
+    </sch:rule>
```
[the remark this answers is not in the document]
bpark1327: You're right.  They don't necessarily have to be broken up.  I will address this.

HPXMLtoOpenStudio/tests/test_validation.rb
bpark1327: Sorry, I think the hash name was misleading. My intention here was to test the validation by deleting the elements of the keys. For example, in Dishwasher case, deleting `RatedAnnualkWh `element is expected to generate an error message because only five elements remain.

HPXMLtoOpenStudio/tests/test_validation.rb
bpark1327: Oops.  This looks wrong.  The effort to remove this huge hash is expected to eliminate this kind of errors.

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -348,57 +348,57 @@ def self.run_validator(hpxml_doc)
       '/HPXML/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem[HeatingSystemType/Furnace]' => {
         '../../HVACDistribution[DistributionSystemType/AirDistribution | DistributionSystemType[Other="DSE"]]' => one_or_more, # See [HVACDistribution]
         'DistributionSystem' => one,
-        'HeatingSystemFuel[text()="natural gas" or text()="fuel oil" or text()="fuel oil 1" or text()="fuel oil 2" or text()="fuel oil 4" or text()="fuel oil 5/6" or text()="diesel" or text()="propane" or text()="kerosene" or text()="electricity" or text()="wood" or text()="wood pellets"]' => one,
+        'HeatingSystemFuel[text()="natural gas" or text()="fuel oil" or text()="fuel oil 1" or text()="fuel oil 2" or text()="fuel oil 4" or text()="fuel oil 5 or 6" or text()="diesel" or text()="propane" or text()="kerosene" or text()="electricity" or text()="wood" or text()="wood pellets"]' => one,
```
shorowit, line 351: Not sure why you're wanting to change all of these, but it's not correct. Refer to the [HXPML enumeration](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem/HeatingSystemFuel)

bpark1327: Hmm.  In the case where a child element is `foo/bar[text()=a or ...]`, I wanted to split the element by slash and choose only bar[text()=a or ...], but it didn't work when there is Fuel 5/6 and I ended up changing the name of this fuel type. I'll have to think of another way.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+      # Schematron validation
+      _test_schematron_validation(@stron_doc, hpxml_doc.to_xml, expected_error_message)
+    end
+
+    puts
+  end
+
+  private
+
+  def _test_schematron_validation(stron_doc, hpxml, expected_error_msgs = nil)
```
shorowit, line 145: `expected_error_msgs` should be `expected_error_msg`?

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+    if expected_error_msgs.nil?
+      assert_empty(results)
+    else
+      idx_of_interest = results.index { |i| i[:message].gsub(': ', [': ', i[:context_path].gsub('h:', '').concat(': ').gsub('/*: ', '')].join('')) == expected_error_msgs }
+      error_msg_of_interest = results[idx_of_interest][:message].gsub(': ', [': ', results[idx_of_interest][:context_path].gsub('h:', '').concat(': ').gsub('/*: ', '')].join(''))
+      assert_equal(expected_error_msgs, error_msg_of_interest)
+    end
+  end
+
+  def _test_ruby_validation(hpxml_doc, expected_error_msgs = nil)
```
shorowit, line 160: `expected_error_msgs` should be `expected_error_msg`?

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+      end
+      next unless found_children
+
+      return _deep_copy_object(hpxml_doc)
+    end
+
+    fail "Could not find HPXML file for target_xpath: #{target_xpath}."
+  end
+
+  def _get_element_name(assertion)
```
shorowit, line 210: The method is called "_get_element_name" but it returns an array. Please update the method name and provide a brief comment describing what the method does.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+
+  def _get_expected_error_message(parent_xpath, assertion)
+    if parent_xpath == '' # root element
+      return [[assertion.partition(': ').first, parent_xpath].join(': '), assertion.partition(': ').last].join() # return "Expected x element(s) for xpath: foo"
+    else
+      return [[assertion.partition(': ').first, parent_xpath].join(': '), assertion.partition(': ').last].join(': ') # return "Expected x element(s) for xpath: foo: bar"
+    end
+  end
+
+  def _get_child_element_with_value(target_xpath, max_number_of_elements_allowed)
```
shorowit, line 245: The method is called "_get_child_element_with_value" but it returns an array. Please update the method name and provide a brief comment describing what the method does.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+          element_name = _get_element_name(assertion)
+          target_xpath = [parent_xpath, element_name]
+          expected_error_message = _get_expected_error_message(parent_xpath, assertion)
+
+          if assertion.start_with?('Expected 0') || assertion.partition(': ').last.start_with?('[not') # FIXME: Is there another way to do this?
+            next if assertion.start_with?('Expected 0 or more') # no tests needed
+
+            expected_error_msgs_by_element_addition[target_xpath] = expected_error_message
+          elsif assertion.start_with?('Expected 1') || assertion.start_with?('Expected 9')
+            expected_error_msgs_by_element_deletion[target_xpath] = expected_error_message
```
shorowit, line 65: I don't understand why it's either element addition or deletion but never both. If we have an "Expected 1" situation, for example, shouldn't we check that it fails both element addition _and_ deletion?

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -0,0 +1,264 @@
...
+    puts
+  end
+
+  def test_schematron_asserts
+    # build up expected error messages hashes by parsing EPvalidator.xml
+    doc = XMLHelper.parse_file(@stron_path)
+    expected_error_msgs_by_element_addition = {}
+    expected_error_msgs_by_element_deletion = {}
+    XMLHelper.get_elements(doc, '/sch:schema/sch:pattern').each do |pattern|
+      XMLHelper.get_elements(pattern, 'sch:rule').each do |rule|
```
shorowit, line 51: Combine as `XMLHelper.get_elements(doc, '/sch:schema/sch:pattern/sch:rule').each do |rule|`

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -0,0 +1,1125 @@
...
+      <sch:assert test='count(h:XMLTransactionHeaderInformation/h:Transaction) = 1'>Expected 1 element(s) for xpath: /HPXML/XMLTransactionHeaderInformation/Transaction</sch:assert> <!-- Required by HPXML schema -->
+      <sch:assert test='count(h:SoftwareInfo/h:extension/h:SimulationControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: /HPXML/SoftwareInfo/extension/SimulationControl</sch:assert> <!-- See [SimulationControl] -->
+    
+      <sch:assert test='count(h:Building) = 1'>Expected 1 element(s) for xpath: /HPXML/Building</sch:assert>
+      <sch:assert test='count(h:Building/h:BuildingID) = 1'>Expected 1 element(s) for xpath: /HPXML/Building/BuildingID</sch:assert> <!-- Required by HPXML schema -->
+      <sch:assert test='count(h:Building/h:ProjectStatus/h:EventType) = 1'>Expected 1 element(s) for xpath: /HPXML/Building/ProjectStatus/EventType</sch:assert> <!-- Required by HPXML schema -->
+    
+      <sch:assert test='not(h:Building/h:BuildingDetails/h:BuildingSummary/h:Site) or count(h:Building/h:BuildingDetails/h:BuildingSummary/h:Site) = 1'>Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/BuildingSummary/Site</sch:assert> <!-- See [Site] -->
```
shorowit, line 17: Why the use of `not` here? Isn't this simply:

```
<sch:assert test='count(h:Building/h:BuildingDetails/h:BuildingSummary/h:Site) &lt;= 1'>Expected 0 or 1 element(s) for xpath: /HPXML/Building/BuildingDetails/BuildingSummary/Site</sch:assert> <!-- See [Site] -->
```

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -0,0 +1,1125 @@
...
+    </sch:rule>
+
+    <!-- [WeatherStation] -->
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:ClimateandRiskZones/h:WeatherStation'>
+      <sch:assert test='count(h:SystemIdentifier) = 1'>Expected 1 element(s) for xpath: SystemIdentifier</sch:assert> <!-- Required by HPXML schema -->
+      <sch:assert test='count(h:Name) = 1'>Expected 1 element(s) for xpath: Name</sch:assert> <!-- Required by HPXML schema -->
+      <sch:assert test='(count(h:WMO) = 1 or count(h:extension/h:EPWFilePath) = 1)'>Expected 1 element(s) for xpath: WMO | extension/EPWFilePath</sch:assert> <!-- Reference weather/data.csv for the list of acceptable WMO station numbers -->
```
shorowit, line 125: Seems like sometimes this pattern is used, other times you use the following pattern. Would be nice to use the same pattern everywhere. 

```
<sch:assert test='(count(h:SystemType/h:Standard) + count(h:SystemType/h:Recirculation) = 1)'>
```

shorowit (commented):
Some initial comments.

HPXMLtoOpenStudio/resources/EPvalidator.xml
bpark1327: This pattern was used for the first implementation. That'll be replaced with the new pattern to be consistent.

HPXMLtoOpenStudio/tests/test_validation.rb
bpark1327: If I remember correctly, we talked about this before. I think you're right. We should check if all assertion passes or fails by both element addition and deletion. Let me address this.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -0,0 +1,1131 @@
...
+      <sch:assert test='count(h:extension/h:UsageMultiplier) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/UsageMultiplier</sch:assert>
+      <sch:assert test='count(h:extension/h:WeekdayScheduleFractions) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/WeekdayScheduleFractions</sch:assert>
+      <sch:assert test='count(h:extension/h:WeekendScheduleFractions) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/WeekendScheduleFractions</sch:assert>
+      <sch:assert test='count(h:extension/h:MonthlyScheduleMultipliers) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/MonthlyScheduleMultipliers</sch:assert>
+    </sch:rule>
+  </sch:pattern>
+  
+  <sch:pattern name='[Neighbors]'>
```
shorowit, line 522: @bpark1327 You have some `sch:rule`s pulled out into separate `sch:pattern`s as needed. I'm sure it works correctly, but I think most developers will find it difficult to understand/know when to extract into a separate `sch:pattern` and when not to. Is there any reason not to have _every_ `sch:rule` in its own `sch:pattern`? That would both simplify the file and as well as allow these `sch:pattern`s you've pulled out to be reinserted at a more logical location (e.g., WindowOverhang adjacent to Window) rather than simply tacked to the end of the file.

bpark1327: Good point. There's actually no reason not to have every `sch:rule` in its own `sch:pattern`. As far as I know, we can create as many patterns as we like. I agree that having every `sch:rule` in its own `sch:pattern` would have a lot of benefits, as you mentioned. I'll address it.

shorowit (approved):
This looks really good. Will let the CI do one more pass on it and will verify the CI results before merging.

shorowit:
Thanks @bpark1327!

PR #424, opened 2020-06-18
Address unmet load for partial HVAC systems

yzhou601:
## Pull Request Description

Partial HVAC systems (i.e., buildings where the sum of fraction heat load served or fraction cool load served is less than 1) result in misleading unmet load values. If a building has a room AC serving 33% of the load, for example, we shouldn't report 67% of the load as unmet. Unmet load should only include the portion of the load that the room AC should be meeting but is not.

To solve this, we use two ideal air systems -- one first in the HVAC equipment list that serves the remainder of the building load served (e.g., 67% in the room AC example) and one last in the HVAC equipment list that only meets any load after the HVAC system has run. The latter system is used to quantity unmet load.

## Checklist

Not all may apply:


yzhou601:
With updated approach, do we want to report two unmet loads? Or just the unexpected unmet load? Or the total (similar to existing unmet load)?

shorowit:
> 
> 
> With updated approach, do we want to report two unmet loads? Or just the unexpected unmet load? Or the total (similar to existing unmet load)?

Sorry, I didn't see your note. We should report just the unexpected unmet load. What are you reporting right now?

yzhou601:
> > With updated approach, do we want to report two unmet loads? Or just the unexpected unmet load? Or the total (similar to existing unmet load)?
> 
> Sorry, I didn't see your note. We should report just the unexpected unmet load. What are you reporting right now?

No worries. We're reporting the total, I will make changes towards reporting unexpected only.

SimulationOutputReport/measure.rb
```diff
@@ -764,13 +761,6 @@ def get_outputs(timeseries_frequency,
   end
 
   def check_for_errors(runner, outputs)
-    all_total = @fuels.values.map { |x| x.annual_output }.inject(:+)
```
yzhou601, line 767: There're two check/test removed because we now only retain the unexpected unmet load. Is it ok to remove them? Or any suggestion?
I think as long as we have the unexpected load available, should we add test to set up limits of unexpected unmet loads?

yzhou601: Addressed.

shorowit:
Thanks @yzhou601!

PR #430, opened 2020-06-24
Support setting daylight saving time

joseph-robertson:
## Pull Request Description

Allow daylight saving to be enabled/disabled.

Defaults to enabled. If enabled, custom dates can be specified. If custom dates are not specified, DST period will be set based on dates in the EPW header if available, otherwise will resort to hardcoded dates.

## Checklist

Not all may apply:


shorowit:
A few reactions:
- This is a good feature to allow. I'm surprised to see that we weren't already using it.
- I think daylight saving should default to on, not off, since it's used almost everywhere in the country.
- We should allow someone to specify start/end dates rather than force them to modify their weather file to specify the dates.

So, here are my initial thoughts on how the logic could work...

1. This results in DST according to user-specified dates (4/7 to 10/15):
```
<DaylightSaving>
  <Enabled>true</Enable>
  <BeginMonth>4</BeginMonth>
  <BeginDayOfMonth>7</BeginDayOfMonth>
  <EndMonth>10</EndMonth>
  <EndDayOfMonth>15</EndDayOfMonth>
</DaylightSaving>
```

2. This results in DST according to the weather file, if available, or backup values of 3/11 to 11/4 (roughly average according to [Wikipedia](https://en.wikipedia.org/wiki/Daylight_saving_time_in_the_United_States)):
```
<DaylightSaving>
  <Enabled>true</Enable>
</DaylightSaving>
```

Having no `DaylightSaving` element would do the same as above.

3. And this results in no DST:
```
<DaylightSaving>
  <Enabled>false</Enable>
</DaylightSaving>
```

joseph-robertson:
So have `<DaylightSaving>` at the same level as `<SimulationControl>`?

joseph-robertson:
Just as a note, here is the motivation in using DST dates stored in the epw header- so that the options lookup doesn't need to be updated every time a new set of epw files in a ResStock analysis. See this issue for reference: https://github.com/NREL/OpenStudio-BuildStock/issues/444.

But it does seem like your proposed logic would accommodate this.

shorowit:
I think it makes sense to be inside SimulationControl like you have it. 

Yes, let me know if the logic has any shortcomings. It's obviously a bit more complicated, but it seems like we should accommodate a few different scenarios. I couldn't see any obvious way to simplify it further.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -635,23 +636,47 @@ def check_for_errors
         end
       end
 
+      if not @dst_begin_month.nil?
```
shorowit, line 639: Rather than copying/pasting new sections of code only to make a few small changes, update the existing start/end code to also check DST. You can either loop through both or create a method to accomplish this.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -38,6 +38,7 @@ def self.apply_header(hpxml)
     hpxml.header.begin_day_of_month = 1 if hpxml.header.begin_day_of_month.nil?
     hpxml.header.end_month = 12 if hpxml.header.end_month.nil?
     hpxml.header.end_day_of_month = 31 if hpxml.header.end_day_of_month.nil?
+    hpxml.header.dst_enabled = false if hpxml.header.dst_enabled.nil?
```
shorowit, line 41: I wanted it to default to true.

HPXMLtoOpenStudio/resources/location.rb
```diff
@@ -60,17 +62,25 @@ def self.apply_year(model, epw_file)
...
+      dst_end_date = "#{month_names[@dst_end_month]} #{@dst_end_day_of_month}"
+    elsif epw_file.daylightSavingStartDate.is_initialized && epw_file.daylightSavingEndDate.is_initialized
+      dst_start_date = epw_file.daylightSavingStartDate.get
+      dst_start_date = "#{month_names[dst_start_date.monthOfYear.value - 1]} #{dst_start_date.dayOfMonth}"
+      dst_end_date = epw_file.daylightSavingEndDate.get
+      dst_end_date = "#{month_names[dst_end_date.monthOfYear.value - 1]} #{dst_end_date.dayOfMonth}"
+    else
+      dst_start_date = 'Mar 11'
+      dst_end_date = 'Nov 4'
+    end
```
shorowit, line 79: Move this code to hpxml_defaults.rb, so that the chosen start/end date can show up in the in.xml file for inspection. And then add checks in test_defaults.rb.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -153,6 +153,12 @@ The simulation run period can be optionally specified with ``BeginMonth``/``Begi
 The ``BeginMonth``/``BeginDayOfMonth`` provided must occur before ``EndMonth``/``EndDayOfMonth`` provided (e.g., a run period from 10/1 to 3/31 is invalid).
 If not provided, default values of January 1st and December 31st will be used.
 
+Whether to apply daylight saving time start/end dates can be optionally denoted with ``DaylightSaving/Enabled``.
+If daylight saving is enabled, the daylight saving period can be optionally specified with ``DaylightSaving/BeginMonth``, ``DaylightSaving/BeginDayOfMonth``, ``DaylightSaving/EndMonth``, and ``DaylightSaving/EndDayOfMonth``.
+If these are not specified, the simulation will run using the daylight saving time start/end dates specified in the header of the weather file (EPW).
+If these are not specified, the simulation will run using the daylight saving time start/end dates corresponding to March 11 and November 4, respectively.
+If ``DaylightSaving`` is not enabled, the default value of false will be used.
```
shorowit, line 160: Confusing sentence. What I want it to say is "If ``DaylightSaving`` is not provided, ``DaylightSaving/Enabled`` will default to true." or something similar.

HPXMLtoOpenStudio/resources/location.rb
```diff
@@ -60,17 +62,25 @@ def self.apply_year(model, epw_file)
...
 
-    dst = model.getRunPeriodControlDaylightSavingTime
-    dst.setStartDate(dst_start_date_month, dst_start_date_day)
-    dst.setEndDate(dst_end_date_month, dst_end_date_day)
+    run_period_control_daylight_saving_time = model.getRunPeriodControlDaylightSavingTime
+    run_period_control_daylight_saving_time.setStartDate(dst_start_date)
+    run_period_control_daylight_saving_time.setEndDate(dst_end_date)
```
shorowit, line 83: Can you add a new test_location.rb file that checks the OS model? Bonus points if you add checks for all the other stuff that location.rb does...

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -669,6 +694,20 @@ def check_for_errors
...
+
+        if (not @dst_begin_day_of_month.nil?) && (not @dst_end_day_of_month.nil?)
+          if @dst_begin_month == @dst_end_month
+            if @dst_begin_day_of_month > @dst_end_day_of_month
+              fail "Daylight Saving Begin Day of Month (#{@dst_begin_day_of_month}) cannot come after Daylight Saving End Day of Month (#{@dst_end_day_of_month}) for the same month (#{@dst_begin_month})."
+            end
+          end
+        end
+      end
+
```
shorowit, line 710: I thought you were intentionally excluding this in your first commit and almost gave it a 👍. I don't think these checks should exist because DST will wrap around the year in the southern hemisphere. I assume EnergyPlus will allow it but you can double-check.

joseph-robertson: I was intentionally excluding at first. But then I checked if you can wrap around the year- and got "Simulation Unsuccessful".

shorowit: And did the error message say?

The documentation examples show that it should be supported:
https://bigladdersoftware.com/epx/docs/9-3/input-output-reference/group-location-climate-weather-file-access.html#runperiodcontroldaylightsavingtime

HPXMLtoOpenStudio/resources/location.rb
joseph-robertson: Can you explain this further?

shorowit: When the DST values are not provided, they should be defaulted [https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hpxml_defaults.rb#L35-L41](here) similar to simulation control start/end dates.

joseph-robertson: So pass `epw_file` into the `HPXMLDefaults` class?

shorowit: Yes.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -615,6 +615,7 @@ class Header < BaseElement
              :software_program_used, :software_program_version, :eri_calculation_version,
              :eri_design, :timestep, :building_id, :event_type, :state_code,
              :begin_month, :begin_day_of_month, :end_month, :end_day_of_month,
+             :dst_enabled, :dst_begin_month, :dst_begin_day_of_month, :dst_end_month, :dst_end_day_of_month,
```
shorowit, line 618: Can you rename the existing `begin_month`, etc. variables to `sim_begin_month`, etc.? It will help distinguish from DST variables.

HPXMLtoOpenStudio/resources/hpxml.rb
joseph-robertson: I copied the example from the link above, for when dst start is after dst end, and I get an energyplys segmentation fault.

shorowit: Well, I think we should allow our code to support it since E+ documentation suggests it should work. Can you create an E+ GitHub issue on the crash and attach your IDF?

joseph-robertson: I updated the code to allow wrap around daylight saving time period (even though it appears there is currently an energyplus issue with that).

shorowit:
Thanks @joseph-robertson!

PR #448, opened 2020-07-14
Mini-split air conditioner

shorowit:
## Pull Request Description

Allows a mini-split to be specified for a HPXML `CoolingSystem`. Previously it could only be modeled by using a HPXML `HeatPump` with fraction heat load served = 0.

## Checklist

Not all may apply:


shorowit:
Verified that ducted MSAC and ducted MSHP (w/ fraction heat load served = 0) files give identical results, both when the system is autosized and hardsized.

PR #454, opened 2020-07-22
Add interior surface test

yzhou601:
## Pull Request Description

Add interior surface tests, waiting on [E+ PR](https://github.com/NREL/EnergyPlus/pull/8152).

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/materials.rb
+++ b/HPXMLtoOpenStudio/resources/materials.rb
@@ -79,6 +79,12 @@ class Material
     return self.AirFilm(rvalue)
   end
 
+  def self.AirFilmMFSpaces
+    combined_film_u = 8.3
+    outside_film_r = 1.0 / UnitConversions.convert(combined_film_u, 'W/(m^2*K)', 'Btu/(hr*ft^2*F)')
+    return self.AirFilm(outside_film_r)
+  end
+
   def self.AirFilmOutsideASHRAE140
     return self.AirFilm(0.174)
   end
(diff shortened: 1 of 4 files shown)


yzhou601:
Found some issues when adding interior surface tests, mostly related to OtherSideCoefficients object. I intended to check assembly R values (with film) to be consistent from hpxml values to E+ reported values for all those added interior surfaces. The OSC surfaces are throwing errors for not being consistent, first I found we didn't use assumed **Combined Convective/Radiative Film Coefficient {W/m2-K}** as air film u value to pick up construction sets, create materials, etc (it should, right?). So I adjusted the air film r values to be based on the OSC assumption (8.3 W/(m^2*K)). This's the reason that there're a bunch of changes related to air film r values. However, even though I addressed air film, the with-film U values are still not consistent. 
Then I calculated the air film Rvalue based on the with-film and no-film numbers on Envelop Summary Report tables, I found the different surfaces attached to the same OSC object can have different air film R values. It means E+ probably is not using that single constant film coefficient for air film calculation, or probably just a report issue of having grabbed the wrong number. If the first, checking no-film r values can be an option. If the latter, it will need to make changes to the E+ pull request.

HPXMLtoOpenStudio/resources/materials.rb
```diff
@@ -79,6 +79,12 @@ def self.AirFilmOutside
     return self.AirFilm(rvalue)
   end
 
+  def self.AirFilmMFSpaces
+    combined_film_u = 8.3
```
shorowit, line 83: We should reference this [here](https://github.com/NREL/OpenStudio-HPXML/blob/interior-surface-test/HPXMLtoOpenStudio/measure.rb#L3413) so that we don't have 8.3 hard-coded in two different places.

shorowit:
Code pulled into #495.

PR #464, opened 2020-08-10
Clothes dryer exhaust

joseph-robertson:
## Pull Request Description

Adds an optional input to describe the dryer as vented or unvented, and if vented, an optional input for the flowrate.

(Removes ClothesDryerExhaust minutely csv files. The clothes dryer exhaust schedule instead automatically mirrors the clothes dryer hot water schedule (but with 24 min on time), which itself mirrors the clothes washer hot water schedule (but with a 1 hour lag).)

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -749,6 +749,7 @@ def self.run_validator(hpxml_doc)
         'EnergyFactor | CombinedEnergyFactor' => zero_or_one,
         'EnergyFactor | CombinedEnergyFactor | ControlType' => zero_or_two,
         'extension/UsageMultiplier' => zero_or_one,
+        'extension/RatedFlowRate' => zero_or_one,
```
shorowit, line 752: It's a bit weird to call the flow rate "rated", since it's not really a rated value. Also, would be clearer if someone could describe the dryer as "vented".

So two suggestions:
1. Add an optional 'extension/IsVented' element (boolean); always default to true.
2. If the above element is provided and true, have `extension/VentedFlowRate` (instead of RatedFlowRate) as an optional element.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -758,6 +758,18 @@ def self.apply_appliances(hpxml, nbeds, eri_version)
...
+        if not clothes_dryer.combined_energy_factor.nil?
+          if clothes_dryer.combined_energy_factor < 4.5
+            clothes_dryer.rated_flow_rate = 100.0
+          end
+        end
+        if not clothes_dryer.energy_factor.nil?
+          if clothes_dryer.energy_factor < 5.175
+            clothes_dryer.rated_flow_rate = 100.0
+          end
+        end
```
shorowit, line 771: Where did this come from?

This will need to be updated to use the IsVented input.

docs/source/hpxml_to_openstudio.rst
```diff
@@ -918,6 +918,8 @@ If ``EnergyFactor`` is provided instead of ``CombinedEnergyFactor``, it will be
 
 An ``extension/UsageMultiplier`` can also be optionally provided that scales energy usage; if not provided, it is assumed to be 1.0.
 
+An ``extension/RatedFlowRate`` can also be optionally provided that indicates whether there is exhaust. If not provided, it is assumed that there is no exhaust.
```
shorowit, line 921: This is inconsistent with the code in hpxml_defaults.rb.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -524,6 +524,23 @@ def test_ventilation_fans
     _test_default_bath_fan_values(hpxml_default, 2, 50, 1, 15, 7)
   end
 
+  def test_clothes_dryer_exhaust
```
shorowit, line 527: Update for the IsVented input.

docs/source/hpxml_to_openstudio.rst
shorowit: Update for the IsVented input.

HPXMLtoOpenStudio/resources/EPvalidator.rb
joseph-robertson: @shorowit For number one, always default to true when the clothes dryer is non-electric?

shorowit: No, default to true for all fuel types. Electric dryers can (and often are) vented.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
joseph-robertson: HSP.

shorowit: Sorry, I understand 100 cfm comes from the HSP but where does the dependency on efficiency come from? A heat pump clothes dryer, which has a very high efficiency, can be either vented or unvented for example.

joseph-robertson:
> A few thoughts:
> 
> * Are you planning to update the clothes dryer _energy_ schedule to also reflect a 1-hr lag from the clothes washer?
> * We need to decide how to handle this new airflow/load in A) the component load outputs in results_annual.csv and B) the airflow cfm outputs in the results_timeseries.csv.

I was planning to handle the 1-hr lag in a separate PR.

Can you provide any guidance for the second thought?

shorowit:
The ideal solution would be new "Clothes Dryer Venting" outputs. 

For the component load (which shows up in annual/timeseries CSVs), you'd need to primarily update [this EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/12da8ffe388f77cf2e5c8fe61f6d6540f75f6240/HPXMLtoOpenStudio/measure.rb#L2931-L2943) and the [reporting measure](https://github.com/NREL/OpenStudio-HPXML/blob/12da8ffe388f77cf2e5c8fe61f6d6540f75f6240/SimulationOutputReport/measure.rb#L1983).

For airflow results (which shows up as cfm in the timeseries CSV), you'll need to primarily update the [reporting measure](https://github.com/NREL/OpenStudio-HPXML/blob/12da8ffe388f77cf2e5c8fe61f6d6540f75f6240/SimulationOutputReport/measure.rb#L2061).

You'll also want to update the outputs listed [in the documentation](https://github.com/NREL/OpenStudio-HPXML/blob/master/docs/source/simulation_output_report.rst).

And we'll need a new test case. You can use the [whole house fan (WHF) test](https://github.com/NREL/OpenStudio-HPXML/blob/master/SimulationOutputReport/tests/output_report_test.rb#L596) as an example, where we check that the average WHF cfm in the timeseries CSV is non-zero.

joseph-robertson:
Why aren't kitchen and bath exhaust separated out from infiltration (like you are proposing clothes dryer exhaust should be)?

shorowit:
Because they are being included inside mechanical ventilation, which is a reasonable grouping. But clothes dryer venting wouldn't make sense to be in there too.

joseph-robertson:
Should look into whether the exhaust schedule lags 1-hr behind the cw.

shorowit:
> Should look into whether the exhaust schedule lags 1-hr behind the cw.

@joseph-robertson If it's supposed to lag, can't we remove the ClothesDryerExhaust csv's and automatically/programmatically offset from the ClotheWasher csv's?

joseph-robertson:
> > Should look into whether the exhaust schedule lags 1-hr behind the cw.
> 
> @joseph-robertson If it's supposed to lag, can't we remove the ClothesDryerExhaust csv's and automatically/programmatically offset from the ClothesWasher csv's?

@shorowit I talked to @jmaguire1 a bit about this. He says that the idea is to have a max delay behind the clothes washer of 60 minutes, and that the clothes dryer exhaust should only happen for 24 minutes at a time. So for example, if the clothes washer was on during minutes 1-2, the clothes dryer should be on during minutes 60-84. It is unclear whether the current set of ClothesDryerExhaust files actually follows this logic.

shorowit:
> It is unclear whether the current set of ClothesDryerExhaust files actually follows this logic.

Can you look into it?

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -952,6 +952,8 @@
       <sch:assert test='count(h:EnergyFactor) + count(h:CombinedEnergyFactor) &lt;= 1'>Expected 0 or 1 element(s) for xpath: EnergyFactor | CombinedEnergyFactor</sch:assert>
       <sch:assert test='(count(h:EnergyFactor) + count(h:CombinedEnergyFactor) + count(h:ControlType) = 0) or (count(h:EnergyFactor) + count(h:CombinedEnergyFactor) + count(h:ControlType) = 2)'>Expected 0 or 2 element(s) for xpath: EnergyFactor | CombinedEnergyFactor | ControlType</sch:assert>
       <sch:assert test='count(h:extension/h:UsageMultiplier) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/UsageMultiplier</sch:assert>
+      <sch:assert test='count(h:extension/h:IsVented) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/IsVented</sch:assert>
+      <sch:assert test='count(h:extension/h:VentedFlowRate) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/VentedFlowRate</sch:assert>
```
shorowit, line 956: These ought to be nested checks (i.e., VentedFlowRate is only of interest if IsVented is true).

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1588,7 +1614,14 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_fans_mech,
       infil_program.addLine("Set Qbath = Qbath + #{UnitConversions.convert(vent_bath.rated_flow_rate * vent_bath.quantity, 'cfm', 'm^3/s').round(4)} * #{bath_sch_sensors_map[vent_bath.id].name}")
     end
 
-    infil_program.addLine('Set Qexhaust = Qrange+Qbath')
+    infil_program.addLine('Set Qdryer = 0')
+    clothes_dryers.each do |clothes_dryer|
+      next unless clothes_dryer.is_vented
```
shorowit, line 1619: Good check. Making sure dryer is vented before retrieving vent flow rate. 👍

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -820,6 +820,20 @@ def self.apply_appliances(hpxml, nbeds, eri_version)
...
+      if clothes_dryer.is_vented && clothes_dryer.vented_flow_rate.nil?
+        if not clothes_dryer.combined_energy_factor.nil?
+          if clothes_dryer.combined_energy_factor < 4.5
+            clothes_dryer.vented_flow_rate = 100.0
+          end
+        elsif not clothes_dryer.energy_factor.nil?
+          if clothes_dryer.energy_factor < HotWaterAndAppliances.calc_clothes_dryer_ef_from_cef(4.5)
+            clothes_dryer.vented_flow_rate = 100.0
+          end
+        end
```
shorowit, line 835: I previously left a comment on this, but you [marked it as resolved without addressing it](https://github.com/NREL/OpenStudio-HPXML/pull/464#discussion_r469366111). 

The code makes no sense for two reasons:
1. If there is a dryer specified as vented but without a flow rate, and its CEF > 4.5, you end up with a nil flow rate. That's not appropriate and probably results in an error in airflow.rb. If you really wanted to do this, you should set `is_vented` based on CEF.
2. But I don't think any of the defaults _should_ be set based on CEF. Heat pump clothes dryers can be vented or unvented. I suggest assuming 100 cfm whenever a dryer is vented, period.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -694,6 +694,25 @@ def test_ventilation_fans
...
+    # Test inputs not overridden by defaults
+    hpxml_name = 'base-appliances-gas.xml'
+    hpxml = HPXML.new(hpxml_path: File.join(@root_path, 'workflow', 'sample_files', hpxml_name))
+    clothes_dryer = hpxml.clothes_dryers[0]
+    clothes_dryer.is_vented = true
```
shorowit, line 702: Since `is_vented` defaults to true, the defaulting of this parameter is not actually being tested. I suggest an additional test where you set `is_vented = false` and ensure it stays false (i.e., not overridden by default and changed to true).

docs/source/hpxml_to_openstudio.rst
```diff
@@ -1025,6 +1025,9 @@ If ``EnergyFactor`` is provided instead of ``CombinedEnergyFactor``, it will be
 
 An ``extension/UsageMultiplier`` can also be optionally provided that scales energy usage; if not provided, it is assumed to be 1.0.
 
+An optional ``extension/IsVented`` element can be used to indicate whether the clothes dryer is vented. If not provided, it is assumed that the clothes dryer is vented.
+If the clothes dryer is vented, an optional ``extension/VentedFlowRate`` element can be used to specify the exhaust cfm.
```
shorowit, line 1029: Need to specify the default value for flow rate when not provided.

shorowit (changes requested):
This is looking pretty good, but there are a few changes still needed. I haven't yet looked at CI results or run any tests, I'll do that when the comments below are addressed.

Also, you still need to look into this: "It is unclear whether the current set of ClothesDryerExhaust files actually follows this logic."

HPXMLtoOpenStudio/resources/EPvalidator.xml
joseph-robertson: @shorowit do you have an example of how to do this?

shorowit: You've done this before. I'm just asking for something like:

```
 <sch:pattern name='[ClothesDryer]'>
    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Appliances/h:ClothesDryer'>
      ...
      <sch:assert test='count(h:extension/h:IsVented) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/IsVented</sch:assert> <!-- See [VentedClothesDryer] -->
    </sch:rule>
  </sch:pattern>

  <sch:pattern name='[VentedClothesDryer]'>
    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Appliances/h:ClothesDryer[h:extension/h:IsVented="true"]'>
      <sch:assert test='count(h:extension/h:VentedFlowRate) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/VentedFlowRate</sch:assert>
    </sch:rule>
  </sch:pattern>
```

joseph-robertson:
The ClothesDryerExhaust files do follow that logic. We can remove them; and "create them on-the-fly" using the ClothesWasher files.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1252,6 +1254,22 @@ def self.apply_local_ventilation(model, vent_object_array, obj_type_name)
     return obj_sch_sensors
   end
 
+  def self.apply_dryer_exhaust(model, dryer_object_array, obj_type_name)
+    obj_sch_sensors = {}
+    dryer_object_array.each_with_index do |dryer_object, index|
+      obj_name = "#{obj_type_name} #{index}"
+      days_shift = -1.0 / 24.0 # Shift by 1 hour
+      obj_sch = HotWaterSchedule.new(model, obj_type_name, @nbeds, days_shift, 24)
+      Schedule.set_schedule_type_limits(model, obj_sch.schedule, Constants.ScheduleTypeLimitsTemperature)
```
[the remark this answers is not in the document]
joseph-robertson: The clothes dryer and clothes dryer exhaust schedules are different.

Clothes dryer: `cd_schedule = HotWaterSchedule.new(model, Constants.ObjectNameClothesDryer, nbeds, days_shift)`

Clothes dryer exhaust: `obj_sch = HotWaterSchedule.new(model, obj_type_name, @nbeds, days_shift, 24)`

The difference is `exhaust_min = 24`, which is used here:
```
    if exhaust_min > 0
      items.reverse.each_with_index do |val, i|
        next unless val > 0
        place = (items.length - 1) - i
        last = place + exhaust_min
        items.fill(1, place...last)
      end
    end
```

I believe `exhaust_min` ensures that the clothes dryer exhausts for at least 24 min after the clothes dryer was last run.

And you're right, seems like the schedule type limits should be `Constants.ScheduleTypeLimitsFraction`.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -526,6 +518,15 @@ def loadMinuteDrawProfileFromFile(timestep_minutes, days_shift, weeks)
...
     end
 
+    if exhaust_min > 0
+      items.reverse.each_with_index do |val, i|
+        next unless val > 0
+        place = (items.length - 1) - i
+        last = place + exhaust_min
+        items.fill(1, place...last)
+      end
+    end
```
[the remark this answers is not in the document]
joseph-robertson: See comments in resource file.

HPXMLtoOpenStudio/resources/airflow.rb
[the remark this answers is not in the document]
shorowit: Got it. Thanks for the explanation.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1452,11 +1471,18 @@ def self.apply_infiltration_adjustment(infil_program, vent_fans_kitchen, vent_fa
       infil_program.addLine("Set Qbath = Qbath + #{UnitConversions.convert(vent_bath.rated_flow_rate * vent_bath.quantity, 'cfm', 'm^3/s').round(4)} * #{bath_sch_sensors_map[vent_bath.id].name}")
     end
 
+    infil_program.addLine('Set Qdryer = 0')
+    clothes_dryers.each do |clothes_dryer|
+      next unless clothes_dryer.is_vented
+
+      infil_program.addLine("Set Qdryer = Qdryer + #{UnitConversions.convert(clothes_dryer.vented_flow_rate, 'cfm', 'm^3/s').round(4)} * #{dryer_sch_sensors_map[clothes_dryer.id].name}")
```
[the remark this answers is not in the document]
shorowit: I think I understand now. I'll look at timeseries outputs but this seems correct.

SimulationOutputReport/measure.rb
```diff
@@ -785,7 +785,11 @@ def get_outputs(timeseries_frequency,
 
     if include_timeseries_airflows
       @airflows.each do |airflow_type, airflow|
-        airflow.timeseries_output = get_report_variable_data_timeseries(['EMS'], airflow.ems_variables.map { |var| "#{var}_timeseries_outvar" }, UnitConversions.convert(1.0, 'm^3/s', 'cfm'), 0, timeseries_frequency)
+        shift = true
+        if airflow_type == AFT::ClothesDryerExhaust
+          shift = false
+        end
```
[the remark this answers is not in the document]
shorowit: You ended up removing this code (that _undoes_ the EMS shift for dryer exhaust), right?

I tested a few different timeseries outputs (1 min timestep w/ 1 min output, 1 min timestep w/ hourly output, 60 min timestep w/ hourly output) and all of the results match the logic you described above, which is that the exhaust should operate when the dryer does, but for a minimum of 24 minutes. And the dryer itself lags behind the washer by 1 hour. So everything looks really good here. 

At this point, I'm only confused by the fact that you thought at one point we had to undo the EMS shift for dryer exhaust. Were you just confused at the time? Are we in agreement that it is working correctly now?

shorowit (approved):
I am happy with everything I've seen. Just need to wait on the latest CI run and double-check its results.

SimulationOutputReport/measure.rb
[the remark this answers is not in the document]
joseph-robertson: So when the EMS lag code was in there, it was shifting the clothes dryer exhaust up one hour (and therefore was occurring before the clothes dryer events). I excluded the shift for clothes dryer exhaust, so that it lined up with clothes dryer events. But then "Mechanical Ventilation" did not coincide with clothes dryer exhaust (because it was still be shifted up one hour). So then I just removed the shifting altogether.

shorowit:
Thanks @joseph-robertson! Really nice work.

PR #467, opened 2020-08-12
BuildResidentialHPXML measure: Shared water heater

joseph-robertson:
## Pull Request Description

Add argument `water_heater_is_shared_system`.

## Checklist

Not all may apply:


BuildResidentialHPXML/measure.rb
```diff
@@ -1594,6 +1594,12 @@ def arguments(model)
     arg.setDefaultValue(false)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeBoolArgument('water_heater_is_shared_system', true)
+    arg.setDisplayName('Water Heater: Is Shared System')
+    arg.setDescription('Whether the water heater is a shared system.')
```
shorowit, line 1599: Since you're not asking for the number of units served by the shared water heater and are assuming it serves all units in the building, which is reasonably fine, I suggest adding that to the description. Something like:
'Whether the water heater is a shared system. If true, assumed to served all the units in the building.'

BuildResidentialHPXML/measure.rb
```diff
@@ -1596,7 +1596,7 @@ def arguments(model)
 
     arg = OpenStudio::Measure::OSArgument::makeBoolArgument('water_heater_is_shared_system', true)
     arg.setDisplayName('Water Heater: Is Shared System')
-    arg.setDescription('Whether the water heater is a shared system. If true, assumed to served all the units in the building.')
+    arg.setDescription('Whether the water heater is a shared system. If true, assumed to serve all the units in the building.')
```
shorowit, line 1599: My bad. 😄

PR #471, opened 2020-08-14
Boiler & GSHP fixes/enhancements

shorowit:
## Pull Request Description

Enhancements:
- Allow shared boiler electric auxiliary energy to be entered or defaulted (in addition to option of calculating it based on shared loop power).
- Adds optional inputs for GSHP pump power and fan power.

Fixes:
- Fixes boiler & GSHP pump power to be a function of heating system PLR (i.e., based on load) instead of pump/loop PLR (i.e., based on flow rate) via EMS.

## Checklist

Not all may apply:


shorowit:
@yzhou601 Can you look into why combi systems fail the tests? Can the tests be updated so combi systems pass? Can you check if the pump energy use makes sense for combi systems? See https://github.com/NREL/OpenStudio-HPXML/pull/471/files#diff-49f44146ae0831fa6adc6dae2b09f461R869-R871

PR #474, opened 2020-08-19
Shared Mech Vent Systems

yzhou601:
## Pull Request Description

Allow shared mechanical ventilation systems (optionally with preconditioning equipment).

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -557,14 +557,48 @@ def self.run_validator(hpxml_doc)
...
+      '/HPXML/Building/BuildingDetails/Systems/MechanicalVentilation/VentilationFans/VentilationFan[UsedForWholeBuildingVentilation="true" and IsSharedSystem="false"]' => {
+        'TestedFlowRate | RatedFlowRate' => one_or_more,
         'FanPower' => one,
       },
 
       ## [MechVentType=HRV]
       '/HPXML/Building/BuildingDetails/Systems/MechanicalVentilation/VentilationFans/VentilationFan[UsedForWholeBuildingVentilation="true" and FanType="heat recovery ventilator"]' => {
-        'SensibleRecoveryEfficiency | AdjustedSensibleRecoveryEfficiency' => one,
+        'FanPower' => one,
+        'TestedFlowRate | RatedFlowRate' => one_or_more,
```
shorowit, line 601: Seems like something weird here?

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -557,14 +557,48 @@ def self.run_validator(hpxml_doc)
...
+      ## [SharedSystem=true]
+      '/HPXML/Building/BuildingDetails/Systems/MechanicalVentilation/VentilationFans/VentilationFan[UsedForWholeBuildingVentilation="true" and IsSharedSystem="true"]' => {
+        'TestedFlowRate[@scope="single unit"] | RatedFlowRate[@scope="single unit"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'TestedFlowRate[@scope="multiple units"] | RatedFlowRate[@scope="multiple units"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'FanPower[@scope="multiple units"] | FanPower[@scope="single unit"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'FanPower[not(@scope)]' => zero, # Use scope attribute to describe at which level data is provided
+        'FractionOutdoorAir | FractionRecirculation' => one_or_more,
+        '../../../../BuildingSummary/BuildingConstruction[ResidentialFacilityType[text()="single-family attached" or text()="apartment unit"]]' => one,
+        'extension/PreconditioningHeating' => zero_or_one,
+        'extension/PreconditioningCooling' => zero_or_one,
```
shorowit, line 573: Can we simply rename these to PreHeating and PreCooling?

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3138,7 +3138,13 @@ class VentilationFan < BaseElement
...
              :used_for_local_ventilation, :total_recovery_efficiency, :total_recovery_efficiency_adjusted,
              :sensible_recovery_efficiency, :sensible_recovery_efficiency_adjusted,
-             :fan_power, :quantity, :fan_location, :distribution_system_idref, :start_hour]
+             :fan_power, :quantity, :fan_location, :distribution_system_idref, :start_hour,
+             :is_shared_system, :building_rated_flow_rate, :building_tested_flow_rate,
+             :building_fan_power, :fraction_oa, :fraction_recirculation, :preconditioning_heating_fuel,
+             :preconditioning_heating_capacity, :preconditioning_heating_capacity_building,
+             :preconditioning_heating_efficiency, :preconditioning_cooling_fuel,
+             :preconditioning_cooling_capacity, :preconditioning_cooling_capacity_building,
+             :preconditioning_cooling_efficiency]
```
shorowit, line 3147: Also simplify these with :preheating_foo and :precooling_foo.

HPXMLtoOpenStudio/resources/EPvalidator.rb
```diff
@@ -557,14 +557,48 @@ def self.run_validator(hpxml_doc)
...
+        'IsSharedSystem' => one, # See [SharedSystem]
+      },
+
+      ## [SharedSystem=true]
+      '/HPXML/Building/BuildingDetails/Systems/MechanicalVentilation/VentilationFans/VentilationFan[UsedForWholeBuildingVentilation="true" and IsSharedSystem="true"]' => {
+        'TestedFlowRate[@scope="single unit"] | RatedFlowRate[@scope="single unit"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'TestedFlowRate[@scope="multiple units"] | RatedFlowRate[@scope="multiple units"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'FanPower[@scope="multiple units"] | FanPower[@scope="single unit"]' => one_or_more, # Use scope attribute to describe at which level data is provided
+        'FanPower[not(@scope)]' => zero, # Use scope attribute to describe at which level data is provided
+        'FractionOutdoorAir | FractionRecirculation' => one_or_more,
```
shorowit, line 570: I suggest just asking for one of these, not both, since they need to sum to 1 (and may be off due to rounding). No strong preference but I guess I'd choose FractionRecirculation.

shorowit:
No need to worry about the shared PV error on the CI, it will be resolved once the shared HVAC PR is merged.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3371,29 +3366,22 @@ def to_oga(doc)
       XMLHelper.add_attribute(sys_id, 'id', @id)
       XMLHelper.add_element(ventilation_fan, 'Quantity', to_integer(@quantity)) unless @quantity.nil?
       XMLHelper.add_element(ventilation_fan, 'FanType', @fan_type) unless @fan_type.nil?
-      if @is_shared_system
-        rated_fr = XMLHelper.add_element(ventilation_fan, 'RatedFlowRate', to_float(@rated_flow_rate)) unless @rated_flow_rate.nil?
-        tested_fr = XMLHelper.add_element(ventilation_fan, 'TestedFlowRate', to_float(@tested_flow_rate)) unless @tested_flow_rate.nil?
-      else
-        XMLHelper.add_element(ventilation_fan, 'RatedFlowRate', to_float(@rated_flow_rate)) unless @rated_flow_rate.nil?
-        XMLHelper.add_element(ventilation_fan, 'TestedFlowRate', to_float(@tested_flow_rate)) unless @tested_flow_rate.nil?
```
yzhou601: Oops, sorry that forgot to clean this up after removing scope attributes.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1478,104 +1668,83 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_fans_mech,
...
+    # Set up some variables for use of mech vent EMS program
+    infil_program = apply_erv_hrv_fans(model: model, program: infil_program, vent_mech_erv_hrv: vent_mech_erv_hrv_unprecond, hrv_erv_effectiveness_map: hrv_erv_effectiveness_map, fan_sens_load_actuator: fan_sens_load_actuator, fan_lat_load_actuator: fan_lat_load_actuator, preconditioned: false)
 
-    # Apply CFIS
-    infil_program.addLine('Set QWHV_cfis = 0.0')
+    # Apply unpreconditioned CFIS
     infil_program.addLine("Set #{cfis_fan_actuator.name} = 0.0")
+    infil_program = apply_cfis(infil_program, vent_mech_cfis_unprecond, cfis_fan_actuator)
+
+    # FIXME: Is shared exhaust system existing?
```
shorowit, line 1736: What is the question here?

shorowit (commented):
Some initial review comments.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2705,7 +2705,7 @@ def self.add_component_loads_output(runner, model, spaces)
       end
     end
     model.getOtherEquipments.sort.each do |o|
-      next unless o.name.to_s.start_with? Constants.ObjectNameERVHRV
+      next unless o.name.to_s.start_with? Constants.ObjectNameMechanicalVentilationHouseFan + ' sensible load'
```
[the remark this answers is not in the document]
yzhou601: It's because the load we're comparing in component loads is specific to sensible loads. The EnergyTransfer meter only reports sensible loads, but in the new infiltration + ventilation program I also calculates latent loads and had a meter for latent. So I pointed out sensible specifically here to filter out the latent load object.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -419,12 +419,19 @@ def self.create_return_air_duct_zone(model, loop_name)
     return ra_duct_zone
   end
 
-  def self.create_sens_lat_load_actuator_and_equipment(model, name, space, frac_lat, frac_lost, is_duct_load_for_report = nil)
+  def self.create_sens_lat_load_actuator_and_equipment(model:, name:, space:, frac_lat:, frac_lost:, hpxml_fuel_type: nil, end_use: nil, is_duct_load_for_report: nil)
```
yzhou601, line 422: change to the other equipment object creation to include optional fuel and end use inputs, probably better to rename it as create_other_equipment_object_and_actuator?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1243,135 +1250,154 @@ def self.apply_local_ventilation(model, vent_object_array, obj_type_name)
...
-      # Must assume an operating condition (HVI seems to use CSA 439)
-      t_sup_in = 0.0
-      w_sup_in = 0.0028
-      t_exh_in = 22.0
-      w_exh_in = 0.0065
-      cp_a = 1006.0
-      p_fan = vent_mech.average_fan_power # Watts
+  def self.calc_hrv_erv_effectiveness(vent_mech_fans)
+    # Create the mapping between mech vent instance and the effectiveness results
+    hrv_erv_effectiveness_map = {}
```
yzhou601, line 1255: Change this method to return a map, so that it can only be called once at the top level and store data for all methods related to HRV/ERV calculations.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,241 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
     return vent_mech_fan_actuator
   end
 
-  def self.apply_erv_hrv_load_to_infil_program(model, infil_program, vent_mech_cfm, vent_mech_sens_eff, vent_mech_lat_eff, erv_sens_load_actuator, erv_lat_load_actuator)
+  def self.apply_fan_load_to_infil_program(model:, infil_program:, oa_flow_rate_var_name:, tot_flow_rate_var_name:, vent_mech_sens_eff:, vent_mech_lat_eff:, fan_sens_load_actuator:, fan_lat_load_actuator:, preconditioned: true)
```
yzhou601, line 1428: Change previous ERV/HRV program to be suitable for all mech vent systems.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,241 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
+      infil_program.addLine('Set FanTotalToLv = Fan_MFR * (OASupInEnth - ZoneAirEnth)')
+      infil_program.addLine('Set FanSensToLv = Fan_MFR * ZoneCp * (OASupInTemp - ZoneTemp)')
+      infil_program.addLine('Set FanLatToLv = FanTotalToLv - FanSensToLv')
+    end
 
     # Actuator
-    infil_program.addLine("Set #{erv_sens_load_actuator.name} = #{erv_sens_load_actuator.name} + ERVSensToLv")
-    infil_program.addLine("Set #{erv_lat_load_actuator.name} = #{erv_lat_load_actuator.name} + ERVLatToLv")
-    infil_program.addLine("Set QWHV_ervhrv = QWHV_ervhrv + #{UnitConversions.convert(vent_mech_cfm, 'cfm', 'm^3/s')}")
+    if not preconditioned
```
yzhou601, line 1456: This condition is applied only if ERV/HRV is called out of the preconditioning program/subprogram. If the ERV/HRV is a shared preconditioned system, this method will be called in preconditioning method with preconditioned = true, shouldn't apply the actuator and QWHV in that case because preconditioning program will handle that later.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,241 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
-      if (air_infil.unit_of_measure == HPXML::UnitsACH) && !air_infil.house_pressure.nil?
-        living_ach = air_infil.air_leakage
-        living_ach50 = calc_air_leakage_at_diff_pressure(0.65, living_ach, air_infil.house_pressure, 50.0)
-      elsif (air_infil.unit_of_measure == HPXML::UnitsCFM) && !air_infil.house_pressure.nil?
-        living_ach = air_infil.air_leakage * 60.0 / @infil_volume # Convert CFM to ACH
-        living_ach50 = calc_air_leakage_at_diff_pressure(0.65, living_ach, air_infil.house_pressure, 50.0)
-      elsif air_infil.unit_of_measure == HPXML::UnitsACHNatural
-        if @apply_ashrae140_assumptions
-          living_const_ach = air_infil.air_leakage
+  def self.setup_mech_vent_vars_actuators(model:, program:)
```
yzhou601, line 1465: Create a method to have some top level actuators/variables set up, so that these can only be created once throughout the program.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,241 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
+    if not vent_mech_erv_hrv.empty?
+      # ERV/HRV EMS load model
+      # E+ ERV model is using standard density for MFR calculation, caused discrepancy with other system types.
+      # Therefore ERV is modeled within EMS infiltration program
+      # combine ERV/HRV to a single system with weighted effectivenesses
+      vent_mech_sens_eff = 0.0
+      vent_mech_lat_eff = 0.0
+      oa_flow_rate_sum = vent_mech_erv_hrv.map { |fan| fan.average_oa_flow_rate }.sum
+      tot_flow_rate_sum = vent_mech_erv_hrv.map { |fan| fan.average_unit_flow_rate }.sum
+      vent_mech_erv_hrv.each do |vent_fan|
```
yzhou601, line 1500: The ERV/HRV method here did two things: 1. combine ERV/HRV effectiveness to a single one. 2. call the apply_fan_load_to_infil_program for load calculations.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,241 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
+
+    vent_mech_precond.each_with_index do |vent_mech, i|
+      # Create energy use actuator per system because of different fuel type can be applied
+      # FIXME: Preconditioning equipment energy end use: mech vent or hvac category? Proceed with a new end use: mech vent preconditioning
+      clg_energy_actuator = create_sens_lat_load_actuator_and_equipment(model: model, name: "shared mech vent precooling energy #{i}", space: @living_space, frac_lat: 0.0, frac_lost: 1.0, hpxml_fuel_type: vent_mech.precooling_fuel, end_use: Constants.ObjectNameMechanicalVentilationPreconditioning)
+      htg_energy_actuator = create_sens_lat_load_actuator_and_equipment(model: model, name: "shared mech vent preheating energy #{i}", space: @living_space, frac_lat: 0.0, frac_lost: 1.0, hpxml_fuel_type: vent_mech.preheating_fuel, end_use: Constants.ObjectNameMechanicalVentilationPreconditioning)
+      infil_program.addLine("Set #{clg_energy_actuator.name} = 0.0")
+      infil_program.addLine("Set #{htg_energy_actuator.name} = 0.0")
+
+      if not [HPXML::MechVentTypeERV, HPXML::MechVentTypeHRV].include? vent_mech.fan_type
```
yzhou601, line 1535: If cfis or erv/hrv, call its method and connect the calculations with some local variables: Q_precond_oa, Q_precond_tot, etc.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: I'm excluding shared exhaust system here because I feel like it's not physically possible, just not sure whether we need to allow such combination.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1619,15 +1788,54 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_fans_mech,
...
+    apply_infiltration_to_unvented_crawlspace(model, weather)
+    apply_infiltration_to_vented_attic(model, weather, vented_attic)
+    apply_infiltration_to_unvented_attic(model, weather)
+
+    # Local ventilation
+    range_sch_sensors_map = apply_local_ventilation(model, vent_fans_kitchen, Constants.ObjectNameMechanicalVentilationRangeFan)
+    bath_sch_sensors_map = apply_local_ventilation(model, vent_fans_bath, Constants.ObjectNameMechanicalVentilationBathFan)
+
+    # Get mechanical ventilation
+    apply_mechanical_ventilation(model, vent_fans_mech, living_ach50, living_const_ach, weather, vent_fans_kitchen, vent_fans_bath, range_sch_sensors_map, bath_sch_sensors_map, has_flue_chimney)
```
yzhou601, line 1822: I combined the mech vent part into another method. It started with a feeling that it will make things more clear by separating mech vent out, but ended up with too many variables (highly intertwined with others) in this method, not sure whether it helps a lot on clearing things up.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1479,155 +1724,128 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_fans_mech,
...
-      infil_program.addLine('Set Qb = Qfan - Qu') # Balanced flow
-      infil_program.addLine('Set Qtot = (((Qu^2) + (Qinf^2)) ^ 0.5) + Qb')
-      infil_program.addLine('Set Qinf_adj = Qtot - Qu - Qb')
+    infil_program = apply_infiltration_adjustment(infil_program, vent_fans_kitchen, vent_fans_bath, sup_cfm_tot, exh_cfm_tot, bal_cfm_tot, erv_hrv_cfm_tot, infil_flow_actuator, range_sch_sensors_map, bath_sch_sensors_map)
+
+    # Address load of Qfan (Qload)
+    # Qload as variable for tracking outdoor air flow rate, excluding recirculation
+    infil_program.addLine('Set Qload = Qfan')
+    (vent_mech_sup_tot + vent_mech_bal_tot + vent_mech_erv_hrv_tot + vent_mech_cfis_tot).each do |vent_mech_supply_side|
+      # Subtract recirculation air flow rate from Qfan, can only come from supply side
```
shorowit, line 1798: I understand that you're intentionally excluding exhaust, but is there any harm in including it? Won't average_unit_flow_rate minus average_oa_flow_rate simply be zero for an exhaust system?

It would make the code a lot simpler if we could just do `vent_fans_mech.each do |vent_fan_mech| ...`.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,296 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
-    apply_infiltration_to_unvented_attic(model, weather)
+        infil_program.addLine("  Set #{clg_energy_actuator.name} = (@Max FanTotalToLv FanSensToLv) / #{f_precool.precooling_efficiency_cop}")
+      end
+      infil_program.addLine('EndIf')
+    end
 
-    # Local ventilation
-    range_sch_sensors_map = apply_local_ventilation(model, vent_fans_kitchen, Constants.ObjectNameMechanicalVentilationRangeFan)
-    bath_sch_sensors_map = apply_local_ventilation(model, vent_fans_bath, Constants.ObjectNameMechanicalVentilationBathFan)
+    return infil_program
```
shorowit, line 1698: Similar to above comment, shouldn't have to `return infil_program`. Comment probably applies to other methods too.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1479,155 +1724,128 @@ def self.apply_infiltration_and_ventilation_fans(model, weather, vent_fans_mech,
...
-
-    model.getBuilding.additionalProperties.setFeature(Constants.SizingInfoMechVentLatentEffectiveness, weighted_vent_mech_lat_eff)
-    model.getBuilding.additionalProperties.setFeature(Constants.SizingInfoMechVentApparentSensibleEffectiveness, weighted_vent_mech_apparent_sens_eff)
+    # Common variable and load actuators across multiple mech vent calculations, create only once
+    infil_program, fan_sens_load_actuator, fan_lat_load_actuator = setup_mech_vent_vars_actuators(model: model, program: infil_program)
 
     # Apply CFIS
-    infil_program.addLine('Set QWHV_cfis = 0.0')
     infil_program.addLine("Set #{cfis_fan_actuator.name} = 0.0")
+    infil_program = apply_cfis(infil_program, vent_mech_cfis_tot, cfis_fan_actuator)
```
shorowit, line 1788: No need to `infil_program = `. Comment applies to additional lines of code below.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3252,23 +3254,58 @@ def distribution_system
...
+      # Need this method for cfis
+      if not @is_shared_system
+        if not @tested_flow_rate.nil?
+          return @tested_flow_rate
+        else
+          return @rated_flow_rate
+        end
+      else
+        if @fan_type == HPXML::MechVentTypeExhaust && @fraction_recirculation > 0.0
+          fail "Exhaust fan #{@id} can not have recirculation."
```
shorowit, line 3267: 👍

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -904,12 +904,13 @@ def self.process_infiltration_ventilation(model, zone_loads, weather, infilvolum
...
 
-    q_unb, q_bal_Sens, q_bal_Lat = get_ventilation_rates(model)
+    q_unb_cfm, q_preheat, q_precool, q_recirc, q_bal_Sens, q_bal_Lat = get_ventilation_rates(model)
 
-    cfm_Heating = q_bal_Sens + (icfm_Heating**2.0 + q_unb**2.0)**0.5
+    cfm_Heating = q_bal_Sens + (icfm_Heating**2.0 + q_unb_cfm**2.0)**0.5 - q_preheat - q_recirc
 
-    cfm_Cool_Load_Sens = q_bal_Sens + (icfm_Cooling**2.0 + q_unb**2.0)**0.5
-    cfm_Cool_Load_Lat = q_bal_Lat + (icfm_Cooling**2.0 + q_unb**2.0)**0.5
+    # FIXME: Assume preconditioning doesn't introduce/handle latent loads?
```
shorowit, line 911: I would do whatever is easiest since we don't really know if the preconditioning equipment can address the latent load or not.

HPXMLtoOpenStudio/tests/test_airflow.rb
```diff
@@ -144,13 +154,14 @@ def test_mechanical_ventilation_none
...
-    assert_in_epsilon(0.0, UnitConversions.convert(program_values['QWHV_bal'].sum, 'm^3/s', 'cfm'), 0.01)
-    assert_in_epsilon(0.0, UnitConversions.convert(program_values['QWHV_cfis'].sum, 'm^3/s', 'cfm'), 0.01)
-    assert_in_epsilon(0.0, UnitConversions.convert(program_values['QWHV_ervhrv'].sum, 'm^3/s', 'cfm'), 0.01)
+    assert_in_epsilon(0.0, UnitConversions.convert(program_values['QWHV_bal_erv_hrv'].sum, 'm^3/s', 'cfm'), 0.01)
     assert_in_epsilon(1, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationHouseFan).size, 0.01)
     assert_in_epsilon(0.0, get_eed_for_ventilation(model, Constants.ObjectNameMechanicalVentilationHouseFan)[0].designLevel.get, 0.01)
     assert_in_epsilon(0.0, UnitConversions.convert(program_values['Qrange'].sum, 'm^3/s', 'cfm'), 0.01)
     assert_in_epsilon(0.0, UnitConversions.convert(program_values['Qbath'].sum, 'm^3/s', 'cfm'), 0.01)
+    # Load actuators
+    assert_in_epsilon(1, get_oed_for_ventilation(model, "#{Constants.ObjectNameMechanicalVentilationHouseFan} sensible load").size, 0.01)
```
shorowit, line 163: Why assert_in_epsilon? Just use assert_equal(1, foo.size) here and elsewhere.

shorowit (commented):
That was a nice set of code improvements! Some fairly minor comments below:

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -739,13 +732,15 @@
 
   <sch:pattern name='[MechanicalVentilationType=SharedWithPreHeating]'>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:MechanicalVentilation/h:VentilationFans/h:VentilationFan[h:UsedForWholeBuildingVentilation="true" and h:IsSharedSystem="true"]/h:extension/h:PreHeating'>
+      <sch:assert test='count(../../h:FanType[text()="exhaust only"]) = 0'>Expected 0 element(s) for xpath: ../../FanType[text()="exhaust only"]</sch:assert>
```
yzhou601, line 735: Probably not an issue in the context, but will it disable any exhaust fan (including in-unit exhaust) once there's one preconditioned fan?

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1399,77 +1425,296 @@ def self.add_ee_for_vent_fan_power(model, obj_name, frac_lost, is_cfis, pow = 0.
...
     # Actuator
-    infil_program.addLine("Set #{erv_sens_load_actuator.name} = #{erv_sens_load_actuator.name} + ERVSensToLv")
-    infil_program.addLine("Set #{erv_lat_load_actuator.name} = #{erv_lat_load_actuator.name} + ERVLatToLv")
-    infil_program.addLine("Set QWHV_ervhrv = QWHV_ervhrv + #{UnitConversions.convert(vent_mech_cfm, 'cfm', 'm^3/s')}")
+    if not preconditioned
+      infil_program.addLine("Set #{fan_sens_load_actuator.name} = #{fan_sens_load_actuator.name} + FanSensToLv")
+      infil_program.addLine("Set #{fan_lat_load_actuator.name} = #{fan_lat_load_actuator.name} + FanLatToLv")
+    else
+      # FIXME: Should latent be subtracted if heating, or if latent load is negative?
+      # FIXME: Should energy calculation later in the program uses FanTotalToLv or FanSensToLv in above situations?
```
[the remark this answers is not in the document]
yzhou601: You mean apply FanSensToLv all the time or just in such situations?

HPXMLtoOpenStudio/resources/EPvalidator.xml
shorowit: Given the context, it should only look at the fan type of a shared ventilation fan with preheating. But feel free to test it and double-check.

yzhou601: Just realized the base-mechven-shared-preconditioning.xml already did that.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -726,12 +726,13 @@
   <sch:pattern name='[MechanicalVentilationType=CFIS]'>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:MechanicalVentilation/h:VentilationFans/h:VentilationFan[h:UsedForWholeBuildingVentilation="true" and h:FanType="central fan integrated supply"]'>
       <sch:assert test='count(h:AttachedToHVACDistributionSystem) = 1'>Expected 1 element(s) for xpath: AttachedToHVACDistributionSystem</sch:assert>
+      <sch:assert test='count(h:IsSharedSystem) = 0'>Expected 0 element(s) for xpath: IsSharedSystem</sch:assert>
```
shorowit: Hmm, this seems like it wouldn't allow a HPXML file where a CFIS system has IsSharedSystem=false.

Instead, how about checking that the fan type is not CFIS inside [MechanicalVentilationType=Shared]? (This is a similar approach used to check that systems w/ preheating or precooling aren't exhaust only, see lines 744 and 752 for example.)

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -726,12 +726,13 @@
...
     </sch:rule>
   </sch:pattern>
 
   <sch:pattern name='[MechanicalVentilationType=Shared]'>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:MechanicalVentilation/h:VentilationFans/h:VentilationFan[h:UsedForWholeBuildingVentilation="true" and h:IsSharedSystem="true"]'>
-      <sch:assert test='count(h:FractionRecirculation) = 1'>Expected 1 element(s) for xpath: FractionRecirculation</sch:assert>
+      <sch:assert test='count(h:FractionRecirculation) = 1 and number(h:FractionRecirculation) &gt;=0 and number(h:FractionRecirculation) &lt;=1'>Expected 1 element(s) value from 0.0 to 1.0 for xpath: FractionRecirculation</sch:assert>
```
shorowit, line 735: I haven't tested this, but I think it could simply be:
```suggestion
      <sch:assert test='count(h:FractionRecirculation[number(text()) &gt;= 0 and number(text()) &lt;= 1]) = 1'>Expected 1 element(s) for xpath: FractionRecirculation[number(text()) >= 0 and number(text()) <= 1]</sch:assert>
```

HPXMLtoOpenStudio/resources/EPvalidator.xml
yzhou601: I've fixed this in the last commit where I checked there's no issharedsystem=true for  [MechanicalVentilationFanType=CFIS], should be equivalent to checking system type inside [MechanicalVentilationType=Shared]

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -1605,8 +1605,8 @@ def self.apply_mechanical_ventilation(model, vent_fans_mech, living_ach50, livin
     cfis_cfm_tot = vent_mech_cfis_tot.map { |vent_mech| vent_mech.average_total_unit_flow_rate }.sum(0.0)
 
     # Average preconditioned oa air cfms (only oa, recirculation will be addressed below for all shared systems)
-    oa_cfm_preheat = vent_mech_preheat.map { |vent_mech| vent_mech.average_oa_unit_flow_rate }.sum(0.0)
-    oa_cfm_precool = vent_mech_precool.map { |vent_mech| vent_mech.average_oa_unit_flow_rate }.sum(0.0)
+    oa_cfm_preheat = vent_mech_preheat.map { |vent_mech| vent_mech.average_oa_unit_flow_rate * vent_mech.preheating_fraction_load_served }.sum(0.0)
+    oa_cfm_precool = vent_mech_precool.map { |vent_mech| vent_mech.average_oa_unit_flow_rate * vent_mech.precooling_fraction_load_served }.sum(0.0)
```
yzhou601, line 1609: @shorowit Sorry for bothering again for the hvac sizing. I am a bit concerned about an ERV/HRV fan combined with a preconditioned shared system. Is it assuming that the preconditioning happens after heat/energy recovery? Then I think airflows in hvac sizing should not subtract all the `vent_mech.average_oa_unit_flow_rate` as preconditioned (assuming fraction load served is 1.0), but only the portion after recovery. It looks like there's a portion of OA where the ERV/HRV heat recovery and preconditioning have been both counted in hvac sizing. Does it make sense? Though not sure how much it will impact the results and whether it is a case worth more efforts.

shorowit: So I ran 2 files: 1) A shared HRV w/ preconditioning and 2) the same file without any mechanical ventilation. I then looked at the design loads for these two files (hint: you can see them by running workflow/run_simulation.rb with the --debug flag and then looking at the extra output in run.log). All of the design loads were identical (except for Cool Infil Lat). This seems like the correct result to me, as the infiltration should be the same between a balanced ventilation system and no ventilation system, and the ventilation outdoor air load introduced is being fully met by the preconditioning equipment.

I then ran the same exercise but with a shared supply system w/ preconditioning instead of HRV. In this case, the design loads were not identical, since the infiltration will differ between the supply system and no mech vent system (while the ventilation outdoor air load introduced is presumably still being fully met by the preconditioning equipment).

So I don't have any concerns here unless I missed your point.

shorowit: GitHub isn't letting me upload my test files for some reason so I'll send on Slack.

shorowit (approved):
This is in really good shape @yzhou601, thanks for your hard work here! I did a fair amount of testing and didn't notice any red flags. Will do a bit more looking at the results in the context of ERI scores before merging this.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: Oh I figured it out. I [only looped non-preconditioned ERV/HRV fans for combined effectiveness](https://github.com/NREL/OpenStudio-HPXML/blob/shared-mech-vent-precond-v3/HPXMLtoOpenStudio/resources/airflow.rb#L1625-L1628), so that preconditioning and heat/energy recovery won't be both fully-counted in a single system. Apparently I've thought about this issue and have avoided it.🤦

PR #483, opened 2020-09-15
Schematron, Phase 2

bpark1327:
## Pull Request Description

Some additional enhancements/improvements for the Schematron document.
- Replace not(foo) assertions with nested rules
- Add warnings via `sch:report`

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -102,8 +102,7 @@
       <sch:report role='WARN' test='count(h:Appliances/h:Dishwasher) = 0'>No dishwasher specified, the model will not include dishwasher energy use.</sch:report>
       <sch:report role='WARN' test='count(h:Appliances/h:Refrigerator) = 0'>No refrigerator specified, the model will not include refrigerator energy use.</sch:report>
       <sch:report role='WARN' test='count(h:Appliances/h:CookingRange) = 0'>No cooking range specified, the model will not include cooking range/oven energy use.</sch:report>
-      <sch:report role='WARN' test='count(h:Systems/h:WaterHeating/h:WaterHeatingSystem) = 0'>No water heater specified, the model will not include water heating energy use.</sch:report>
-    </sch:rule>
+      <sch:report role='WARN' test='count(h:Systems/h:WaterHeating/h:WaterHeatingSystem) = 0'>No water heater specified, the model will not include water heating energy use.</sch:report>    </sch:rule>
```
shorowit, line 105: Looks like an inadvertent change.

bpark1327: Oops. Thanks for catching.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -1182,6 +1146,30 @@
...
+      <sch:extends rule='appliance_location'/>
+    </sch:rule>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Appliances/h:Freezer/h:Location'>
+      <sch:extends rule='appliance_location'/>
+    </sch:rule>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Appliances/h:CookingRange/h:Location'>
+      <sch:extends rule='appliance_location'/>
+    </sch:rule>
+  </sch:pattern>
+
```
bpark1327, line 1172: @shorowit what do you think this abstract rule approach to gather multiple rules and place them in one pattern? I guess this would make future maintenance a little easier.

shorowit: I like the idea of a single pattern. Thought the abstract rule definitely adds some complexity. A different approach would be to have a single rule w/ a wildcard, e.g.:

`context='/h:HPXML/h:Building/h:BuildingDetails/h:Appliances/*/h:Location'` (or `h:*`, whatever the right syntax is)

bpark1327: You're right. I prefer the wildcard approach to the abstract rule approach. I'll make that change. Thanks!

HPXMLtoOpenStudio/resources/validator.rb
```diff
@@ -30,8 +32,24 @@ def self.run_validator(hpxml_doc, stron_path)
...
+            xpath_result = context_element.xpath(report_test)
+          rescue
+            fail "Invalid xpath: #{report_test}"
+          end
+          next unless xpath_result # check if report_test is true
+
+          warning_message = report_element.children.text # the value of sch:report
+          warnings << warning_message
+        end
+      end
```
shorowit, line 50: This is a lot of duplicated code. Can't you integrate into the above code loop?

HPXMLtoOpenStudio/resources/validator.rb
```diff
@@ -3,6 +3,8 @@
 class Validator
   def self.run_validator(hpxml_doc, stron_path)
     errors = []
+    warnings = []
+    abstract_assertions = {}
```
shorowit, line 7: Delete?

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -96,17 +96,32 @@
...
+      <sch:report role='WARN' test='count(h:Appliances/h:Dishwasher) = 0'>No dishwasher specified, the model will not include dishwasher energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:Appliances/h:Refrigerator) = 0'>No refrigerator specified, the model will not include refrigerator energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:Appliances/h:CookingRange) = 0'>No cooking range specified, the model will not include cooking range/oven energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:Systems/h:WaterHeating/h:WaterHeatingSystem) = 0'>No water heater specified, the model will not include water heating energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:MiscLoads/h:PlugLoad[h:PlugLoadType[text()="other"]]) = 0'>No 'other' plug loads specified, the model will not include misc plug load energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:MiscLoads/h:PlugLoad[h:PlugLoadType[text()="TV other"]]) = 0'>No 'TV other' plug loads specified, the model will not include television plug load energy use.</sch:report>
```
shorowit, line 107: I actually changed these two plug loads so that they are required (note that the ERROR rules above expect 1 element), but apparently forgot to remove the warnings (that can no longer be triggered). So these two can be removed.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -96,17 +96,32 @@
...
-      <sch:assert role='ERROR' test='not(h:SiteType) or count(h:SiteType[text()="urban" or text()="suburban" or text()="rural"]) = 1'>Expected 1 element(s) for xpath: [not(SiteType)] | SiteType[text()="urban" or text()="suburban" or text()="rural"]</sch:assert>
+      <sch:assert role='ERROR' test='count(h:SiteType) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SiteType</sch:assert> <!-- See [SiteType] -->
       <sch:assert role='ERROR' test='count(h:extension/h:ShelterCoefficient) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/ShelterCoefficient</sch:assert>
       <sch:assert role='ERROR' test='count(h:extension/h:Neighbors) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/Neighbors</sch:assert> <!-- See [Neighbors] -->
     </sch:rule>
   </sch:pattern>
 
+  <sch:pattern name='[SiteType]'>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:BuildingSummary/h:Site/h:SiteType'>
+      <sch:assert role='ERROR' test='contains("_urban_ _suburban_ _rural_", concat("_", text(), "_"))'>Expected "text()" for xpath: "urban", "suburban", or "rural"</sch:assert>
```
shorowit, line 121: In https://github.com/NREL/OpenStudio-HPXML/pull/489/files, you are proposing messages like "Expected the value to be greater than or equal to 0". It'd be nice if we were more consistent here. How about we change these warnings from:
Expected "text()" for xpath: ...
to:
Expected value to be: ...

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -96,17 +96,32 @@
...
+      <sch:report role='WARN' test='count(h:Systems/h:WaterHeating/h:WaterHeatingSystem) = 0'>No water heater specified, the model will not include water heating energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:MiscLoads/h:PlugLoad[h:PlugLoadType[text()="other"]]) = 0'>No 'other' plug loads specified, the model will not include misc plug load energy use.</sch:report>
+      <sch:report role='WARN' test='count(h:MiscLoads/h:PlugLoad[h:PlugLoadType[text()="TV other"]]) = 0'>No 'TV other' plug loads specified, the model will not include television plug load energy use.</sch:report>
     </sch:rule>
   </sch:pattern>  
   
   <sch:pattern name='[Site]'>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:BuildingSummary/h:Site'>
-      <sch:assert role='ERROR' test='not(h:SiteType) or count(h:SiteType[text()="urban" or text()="suburban" or text()="rural"]) = 1'>Expected 1 element(s) for xpath: [not(SiteType)] | SiteType[text()="urban" or text()="suburban" or text()="rural"]</sch:assert>
+      <sch:assert role='ERROR' test='count(h:SiteType) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SiteType</sch:assert> <!-- See [SiteType] -->
```
shorowit, line 113: So much better! 🎉

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -209,29 +230,41 @@
...
       <sch:assert role='ERROR' test='count(h:Insulation/h:SystemIdentifier) = 1'>Expected 1 element(s) for xpath: Insulation/SystemIdentifier</sch:assert>
       <sch:assert role='ERROR' test='count(h:Insulation/h:AssemblyEffectiveRValue) = 1'>Expected 1 element(s) for xpath: Insulation/AssemblyEffectiveRValue</sch:assert>
     </sch:rule>
   </sch:pattern>
 
+  <sch:pattern name='[RimJoistSiding]'>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Enclosure/h:RimJoists/h:RimJoist/h:Siding'>
+      <sch:assert role='ERROR' test='contains("_wood siding_ _vinyl siding_ _stucco_ _fiber cement siding_ _brick veneer_ _aluminum siding_", concat("_", text(), "_"))'>Expected "text()" for xpath: "wood siding", "vinyl siding", "stucco", "fiber cement siding", "brick veneer", or "aluminum siding"</sch:assert>
+    </sch:rule>
+  </sch:pattern>
```
shorowit, line 266: Should we combine [WallSiding] and [RimJoistSiding] into one pattern, as you did for appliances' location?

HPXMLtoOpenStudio/resources/EPvalidator.xml
shorowit: Right, that's the downside. I don't have a strong preference between A) using wildcards like you have to minimize the potential of inconsistent rules/enumerations between appliances, or B) having a separate rule for each appliance so that the context is more clear. I assume if we went with B, we could still have all the rules within a single pattern.

But I do prefer either of these approaches to using an abstract rule which just seems overly complicated.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -568,23 +607,35 @@
...
       <sch:assert role='ERROR' test='count(h:CoolingSensibleHeatFraction) &lt;= 1'>Expected 0 or 1 element(s) for xpath: CoolingSensibleHeatFraction</sch:assert>
-      <sch:assert role='ERROR' test='not(h:BackupSystemFuel) or count(h:BackupSystemFuel[text()="electricity" or text()="natural gas" or text()="fuel oil" or text()="fuel oil 1" or text()="fuel oil 2" or text()="fuel oil 4" or text()="fuel oil 5/6" or text()="diesel" or text()="propane" or text()="kerosene" or text()="wood" or text()="wood pellets"]) = 1'>Expected 1 element(s) for xpath: [not(BackupSystemFuel)] | BackupSystemFuel[text()="electricity" or text()="natural gas" or text()="fuel oil" or text()="fuel oil 1" or text()="fuel oil 2" or text()="fuel oil 4" or text()="fuel oil 5/6" or text()="diesel" or text()="propane" or text()="kerosene" or text()="wood" or text()="wood pellets"]</sch:assert> <!-- See [HeatPumpBackup] -->
+      <sch:assert role='ERROR' test='count(h:BackupSystemFuel) &lt;= 1'>Expected 0 or 1 element(s) for xpath: BackupSystemFuel</sch:assert> <!-- See [HeatPumpBackup] and [HeatPumpBackupSystemFuel]-->
```
shorowit, line 610: This is good. But it might appear strange to a developer/user why BackupSystemFuel is separated into nested rules but not HeatPumpFuel. I'm not really sure how to handle it. We _need_ nested rules if the element is optional, whereas nested rules _could_ (but don't have to be) used if the element is required. I guess my worry is that a developer will not realize this subtlety and will create an optional element w/o nested rules in the future, which would be incorrect.

So should we move in the direction of nesting rules as much as possible, regardless of whether the element is optional or not? I'm not really sure.

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -55,18 +56,48 @@ def before_setup
 
         if assertion.start_with?('Expected 0 element')
           # Skipping for now
-        elsif assertion.start_with?('Expected 0 or ') || assertion.partition(': ').last.start_with?('[not') # FIXME: Is there another way to do this?
+        elsif assertion.start_with?('Expected 0 or ')
```
shorowit, line 59: Yay!

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -55,18 +56,48 @@ def before_setup
...
+          key = [context_xpath.split('/')[0...-1].reject(&:empty?).join('/').chomp('/'), context_xpath.split('/')[-1]] # override the key
+          @expected_assertions_by_alteration[key] = _get_expected_error_msg(context_xpath, assertion, 'alteration')
         else
           fail "Unexpected assertion: '#{assertion}'."
         end
       end
     end
   end
 
+  def test_role_attributes
```
shorowit, line 74: Thanks!

shorowit (commented):
This is really nice progress. Some comments/questions below. Perhaps we should chat about some of my questions where I wasn't sure what direction to go.

HPXMLtoOpenStudio/resources/EPvalidator.xml
bpark1327: I agree to go in the direction of reducing the amount of code by combining these kinds of elements. I'll search for these elements and put them together.

bpark1327: I've been thinking about this. I like the wildcards approach but the more I think about it, the more I lean towards option B you described in the next comment (i.e. having a separate rule for each appliance within a single pattern). I think providing clear warning messages is more important than simplifying the code. Plus, this should also be more clear for developers.

HPXMLtoOpenStudio/resources/EPvalidator.xml
bpark1327: I agree. An abstract rule is overly complicated. I vote for Option B. I think providing clear warning messages is more important than simplifying the code. This should also be more clear for developers.

HPXMLtoOpenStudio/resources/EPvalidator.xml
bpark1327: Hmm. As you mention, there's a high chance that a developer will not realize this subtlety. I think we should use nesting rules as much as possible. It will result in some additional lines for assert tests but that way, the code will be consistent.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -139,12 +152,22 @@
       <sch:assert role='ERROR' test='count(h:ConditionedBuildingVolume) + count(h:AverageCeilingHeight) &gt;= 1'>Expected 1 or more element(s) for xpath: ConditionedBuildingVolume | AverageCeilingHeight</sch:assert>
       <sch:assert role='ERROR' test='count(h:extension/h:HasFlueOrChimney) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/HasFlueOrChimney</sch:assert>
     </sch:rule>
+    <!-- FIXME: Does the following rules need to be wrapped in separate patterns for consistency? -->
+    <!-- [ResidentialFacilityType] -->
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:BuildingSummary/h:BuildingConstruction/h:ResidentialFacilityType'>
+      <sch:assert role='ERROR' test='contains("_single-family detached_ _single-family attached_ _apartment unit_ _manufactured home_", concat("_", text(), "_"))'>Expected value to be 'single-family detached' or 'single-family attached' or 'apartment unit' or 'manufactured home'</sch:assert>
+    </sch:rule>
```
shorowit, line 159: @bpark1327 Do you think we can go the same approach we went for HPXMLvalidator.xml? E.g., remove the additional rule, and change the assert to be:
```
<sch:assert role='ERROR' test='not(h:ResidentialFacilityType) or contains("_single-family detached_ _single-family attached_ _apartment unit_ _manufactured home_", concat("_", h:ResidentialFacilityType, "_"))'>Expected ResidentialFacilityType to be 'single-family detached' or 'single-family attached' or 'apartment unit' or 'manufactured home'</sch:assert>
```

See https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/HPXMLvalidator.xml#L22 for corresponding assert in HPXMLvalidator.xml.

bpark1327: If we want to use the same approach, the test will be slightly different depending on whether an element is optional or required. 
For optional elements,
`<sch:assert role='ERROR' test='not(h:Foo) or contains("_Bar_ _Baz_", concat("_", h:Foo, "_"))'>Expected Foo to be 'Bar' or 'Baz'</sch:assert>`
For required elements,
`<sch:assert role='ERROR' test='count(h:Foo) = 1 and contains("_Bar_ _Baz_", concat("_", h:Foo, "_"))'>Expected Foo to be 'Bar' or 'Baz'</sch:assert>`
I'm not sure what would be the best way to distinguish between them in messages. Perhaps, we should keep the expected number of elements in messages (e.g., Expected 0 or 1 Foo to be 'Bar' or 'Baz').

Depending on how we want to change our messages, we may need to think about how we want to assign each assertion to different validation tests (i.e. addition, deletion, and alteration tests).

shorowit: I'm not suggesting that we recombine the two tests back into one, just that we change the new test you're creating.

So an optional element becomes:

```
<sch:assert role='ERROR' test='Foo &lt;=1'>Expected 0 or 1 element(s) for xpath: Foo</sch:assert>
<sch:assert role='ERROR' test='not(h:Foo) or contains("_Bar_ _Baz_", concat("_", h:Foo, "_"))'>Expected Foo to be 'Bar' or 'Baz'</sch:assert>
```

If they provide two Foo's, they get the first error. If they provide one Foo but it has the wrong value, they get the second error.

And a required element becomes:

```
<sch:assert role='ERROR' test='Foo =1'>Expected 1 element(s) for xpath: Foo</sch:assert>
<sch:assert role='ERROR' test='not(h:Foo) or contains("_Bar_ _Baz_", concat("_", h:Foo, "_"))'>Expected Foo to be 'Bar' or 'Baz'</sch:assert>
```

If they provide zero Foo's, they get the first error. If they provide one Foo but it has the wrong value, they get the second error.

shorowit: (In other words, the test is the same for optional and required elements.)

bpark1327: Aha, that makes sense! I'll pick this up. Thanks for the clarification.

shorowit (approved):
This looks good to go. Will merge when the CI comes back ✔️.

shorowit:
Thanks @bpark1327!

PR #489, opened 2020-09-16
Schematron schema validation

bpark1327:
## Pull Request Description

Transform schema validation into Schematron-based validation.
- Generate Schematron schema validator from scanning the sample HPXML files
- Validate an HPXML via validator.rb

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -11,6 +11,7 @@
 require_relative 'resources/constants'
 require_relative 'resources/constructions'
 require_relative 'resources/energyplus'
+require_relative 'resources/HPXMLvalidator'
```
shorowit, line 14: Doesn't seem needed? Code below only references validator.rb.

HPXMLtoOpenStudio/resources/schema_validator.xml
```diff
@@ -0,0 +1,714 @@
...
+      <sch:assert role='ERROR' test='contains("_electricity_ _renewable electricity_ _natural gas_ _renewable natural gas_ _fuel oil_ _fuel oil 1_ _fuel oil 2_ _fuel oil 4_ _fuel oil 5/6_ _district steam_ _district hot water_ _district chilled water_ _solar hot water_ _propane_ _kerosene_ _diesel_ _coal_ _anthracite coal_ _bituminous coal_ _coke_ _wood_ _wood pellets_ _combination_ _other_", concat("_", text(), "_"))'>Expected value to be: 'electricity' or 'renewable electricity' or 'natural gas' or 'renewable natural gas' or 'fuel oil' or 'fuel oil 1' or 'fuel oil 2' or 'fuel oil 4' or 'fuel oil 5/6' or 'district steam' or 'district hot water' or 'district chilled water' or 'solar hot water' or 'propane' or 'kerosene' or 'diesel' or 'coal' or 'anthracite coal' or 'bituminous coal' or 'coke' or 'wood' or 'wood pellets' or 'combination' or 'other'</sch:assert>
+    </sch:rule>
+    <sch:rule context='//h:Building/h:BuildingDetails/h:BuildingSummary/h:BuildingOccupancy/h:NumberofResidents'>
+      <sch:assert role='ERROR' test='number(.) &gt;= 0'>Expected value to be greater than or equal to 0 for xpath: </sch:assert>
```
shorowit, line 18: Why the "for xpath" ending? Can we not simply have the message be "Expected value to be greater than or equal to 0"? There are other cases where the message is simply ">Expected value to be: 'create' or 'update'".

HPXMLtoOpenStudio/resources/schema_validator.xml
```diff
@@ -0,0 +1,714 @@
...
+    </sch:rule>
+    <sch:rule context='//h:Building/h:BuildingDetails/h:Enclosure/h:Walls/h:Wall/h:Area'>
+      <sch:assert role='ERROR' test='number(.) &gt; 0'>Expected value to be greater than 0 for xpath: </sch:assert>
```
shorowit, line 142: Is there a difference between using `.` and `text()`? The enumeration checks use `text()` while the numeric checks use `.`. I'd probably use text() everywhere for consistency if we can. One less xpath peculiarity that someone would have to learn, and I feel that text() is more explicit.

HPXMLtoOpenStudio/resources/HPXMLvalidator.rb
```diff
@@ -0,0 +1,234 @@
+# frozen_string_literal: true
+
+require_relative 'xmlhelper'
+require 'oga'
+
+class HPXMLValidator
+  def self.get_elements_from_sample_files()
```
shorowit, line 7: Awesome!

HPXMLtoOpenStudio/resources/HPXMLvalidator.rb
```diff
@@ -0,0 +1,234 @@
...
+    else
+      expanded_elements = deep_copy_object(complex_type_or_group_dict[element_type])
+      expanded_elements.each do |k, v|
+        k.push(element_xpath).flatten!
+        if complex_type_or_group_dict[v].nil?
+          element_xpaths[k] = v
+          next
+        end
+
+        get_expanded_elements(element_xpaths, complex_type_or_group_dict, k, v)
```
shorowit, line 226: Recursive calls 😮 👍

shorowit (commented):
This is really great stuff, @bpark1327. I have a few comments below.

In addition, I suggest moving the HPXMLvalidator.rb code into tasks.rb and renaming schema_validator.xml to HPXMLvalidator.xml (for consistency with EPvalidator.xml).

shorowit:
Seems like we should also have at least one test that demonstrates this is working correctly. How about adding a test to test_validation.rb that reads in base.xml, removes a known required element, runs the new validator, and checks for the appropriate error message? I don't think we need anything fancier than that, at least for now.

shorowit:
> Seems like we should also have at least one test that demonstrates this is working correctly. How about adding a test to test_validation.rb that reads in base.xml, removes a known required element, runs the new validator, and checks for the appropriate error message? I don't think we need anything fancier than that, at least for now.

Or better yet, create a new invalid HPXML file via tasks.rb, and then add the expected error message to `test_invalid` in workflow/tests/hpxml_translator_test.rb.

HPXMLtoOpenStudio/resources/schema_validator.xml
bpark1327: That's a good point! I'll address this.

workflow/tests/hpxml_translator_test.rb
```diff
@@ -176,6 +176,13 @@ def test_invalid
...
+                                                               "Expected value to be: //Building/BuildingDetails/BuildingSummary/Site/SiteType: 'rural' or 'suburban' or 'urban'",
+                                                               "Expected value to be: //Building/BuildingDetails/ClimateandRiskZones/ClimateZoneIECC/Year: '2012' or '2009' or '2006' or '2003'",
+                                                               'Expected value to be: //Building/BuildingDetails/Enclosure/Roofs/Roof/Azimuth: less than 360',
+                                                               'Expected value to be: //Building/BuildingDetails/Enclosure/Roofs/Roof/RadiantBarrierGrade: less than or equal to 3',
+                                                               'Expected value to be: //Building/BuildingDetails/Appliances/Dishwasher/EnergyFactor: less than or equal to 5',
+                                                               'Expected 1 element(s) for xpath: /HPXML/Building/BuildingDetails/BuildingSummary/Site: [not(SiteType)] | SiteType[text()="urban" or text()="suburban" or text()="rural"]'],
```
shorowit, line 185: Maybe now is the right time to reorganize our messages a bit. How about this? Or open to other ideas.
- "Expected value to be 'foo' or 'bar' [context: //Whatever/...]"
- "Expected value to be greater than or equal to 0 [context: //Whatever/...]"
- "Expected 1 element(s) for xpath: 'Foo | Bar' [context: //Whatever/...]"

bpark1327: I like this idea. I can't think of any other way. I'll change messages in this PR and #483. Thanks!

shorowit: Thanks @bpark1327. I'll look this over one more time after you do, but I think it might be ready (or _very_ close) to merge.

tasks.rb
```diff
@@ -4916,6 +4933,220 @@ def download_epws
...
+
+      # Skip element xpaths not being used in sample files
+      element_xpath_with_prefix = element_xpath.reverse.map { |e| "h:#{e}" }
+      context_xpath = element_xpath_with_prefix.join('/').chomp('/')
+      next unless elements_in_sample_files.any? { |item| item.include? context_xpath }
+
+      has_complex_type_or_group_dict = complex_type_or_group_dict[element_type]
+      if has_complex_type_or_group_dict
+        # FIXME: Improve recursive call so that we only add the "final" xpath
+        get_expanded_elements(element_xpaths, complex_type_or_group_dict, element_xpath, element_type)
```
shorowit, line 5061: @bpark1327 will address this.

tasks.rb
```diff
@@ -4916,6 +4933,220 @@ def download_epws
...
+  end
+
+  # Add enumeration and min/max numeric values
+  element_xpaths.each do |element_xpath, element_type|
+    next if element_type.nil?
+
+    # exclude duplicated xpaths
+    # FIXME: Remove this check when recursive code above is updated
+    has_duplicate = element_xpaths.keys.select { |k| (element_xpath != k) && (k.each_cons(element_xpath.size).include? element_xpath) } # check if an array contains another array in particular order
+    next unless has_duplicate.empty?
```
shorowit, line 5075: And this.

HPXMLtoOpenStudio/resources/HPXMLvalidator.xml
```diff
@@ -0,0 +1,714 @@
...
+    </sch:rule>
+    <sch:rule context='//h:Building/h:BuildingDetails/h:MiscLoads/h:FuelLoad/h:FuelType'>
+      <sch:assert role='ERROR' test='contains("_electricity_ _renewable electricity_ _natural gas_ _renewable natural gas_ _fuel oil_ _fuel oil 1_ _fuel oil 2_ _fuel oil 4_ _fuel oil 5/6_ _district steam_ _district hot water_ _district chilled water_ _solar hot water_ _propane_ _kerosene_ _diesel_ _coal_ _anthracite coal_ _bituminous coal_ _coke_ _wood_ _wood pellets_ _combination_ _other_", concat("_", text(), "_"))'>Expected value to be 'electricity' or 'renewable electricity' or 'natural gas' or 'renewable natural gas' or 'fuel oil' or 'fuel oil 1' or 'fuel oil 2' or 'fuel oil 4' or 'fuel oil 5/6' or 'district steam' or 'district hot water' or 'district chilled water' or 'solar hot water' or 'propane' or 'kerosene' or 'diesel' or 'coal' or 'anthracite coal' or 'bituminous coal' or 'coke' or 'wood' or 'wood pellets' or 'combination' or 'other'</sch:assert>
+    </sch:rule>
+    <sch:rule context='//h:Enclosure/h:Windows/h:Window/h:Area'>
+      <sch:assert role='ERROR' test='number(text()) &gt; 0'>Expected value to be greater than 0</sch:assert>
+    </sch:rule>
```
shorowit, line 658: And will look into why the rule context does not start with //Building.

bpark1327:
- The finding xpaths logic is changed. Now the `Building` element is the starting point and the code finds all the lower xpaths of this `Building` element sequentially. This change allowed me to remove the check that removes duplicated xpaths (https://github.com/NREL/OpenStudio-HPXML/pull/489#discussion_r501394083). 
- I wasn't able to change the recursive function to return element_xpath and element_type. Let's discuss if needed. 
- Now most rule context starts with `//Building` except for `XMLTransactionHeaderInformation`. `XMLTransactionHeaderInformation` is an independent element and it is treated separately.

tasks.rb
```diff
@@ -4916,6 +4933,210 @@ def download_epws
...
+  name_space = XMLHelper.add_element(root, 'sch:ns')
+  XMLHelper.add_attribute(name_space, 'uri', 'http://hpxmlonline.com/2019/10')
+  XMLHelper.add_attribute(name_space, 'prefix', 'h')
+  pattern = XMLHelper.add_element(root, 'sch:pattern')
+
+  # construct complexType and group elements dictionary
+  complex_type_or_group_dict = {}
+  ['//xs:complexType', '//xs:group', '//xs:element'].each do |param|
+    base_elements_xsd_doc.xpath(param).each do |param_type|
+      next if param_type.name == 'element' && (not ['XMLTransactionHeaderInformation', 'ProjectStatus', 'SoftwareInfo'].include?(param_type.get('name')))
```
shorowit, line 5007: I wish we didn't have to hardcode paths here, but I'm going to hold up the PR for this.

tasks.rb
```diff
@@ -4916,6 +4933,210 @@ def download_epws
...
+  end
+
+  element_xpaths = {}
+  top_level_elements_of_interest = elements_in_sample_files.map { |e| e.split('/')[1].gsub('h:', '') }.uniq
+  top_level_elements_of_interest.each do |element|
+    top_level_element = []
+    top_level_element << element
+    top_level_element_type = element
+    get_element_full_xpaths(element_xpaths, complex_type_or_group_dict, top_level_element, top_level_element_type)
+  end
```
shorowit, line 5050: This is a definite improvement over the previous code. Thanks @bpark1327!

tasks.rb
```diff
@@ -4916,6 +4933,210 @@ def download_epws
...
+    hpxml_data_type = hpxml_data_types_dict[hpxml_data_type_name]
+    hpxml_data_type = hpxml_data_types_dict[element_type] if hpxml_data_type.nil? # Backup
+    if hpxml_data_type.nil?
+      fail "Could not find data type name for '#{element_type}'."
+    end
+
+    next if hpxml_data_type[:enums].empty? && hpxml_data_type[:min_inclusive].nil? && hpxml_data_type[:max_inclusive].nil? && hpxml_data_type[:min_exclusive].nil? && hpxml_data_type[:max_exclusive].nil?
+
+    element_name = context_xpath.split('/')[-1]
+    context_xpath = context_xpath.split('/')[0..-2].join('/').chomp('/').prepend('/h:HPXML/')
```
shorowit, line 5072: Note that I changed from `prepend('//')` to `prepend('/h:HPXML/')`. We should always try to avoid the former since it checks for an HPXML element _anywhere_ in the file, which is much slower than giving it an absolute xpath so that it only looks at the root node.

shorowit (approved):
I'm happy with where this ended up. It adds minimal runtime while greatly improving the robustness of our validation error-checking. I'll let the CI churn on the latest commit but otherwise I think this is ready to merge.

tasks.rb
bpark1327: I agree. If they weren't `xs:element` elements in `BaseElements.xsd`, they wouldn't have had to be hardcoded.

tasks.rb
bpark1327: 💡

tasks.rb
shorowit: Got it. Good enough for now. Marking resolved.

shorowit:
Thanks @bpark1327!

shorowit:
(And thanks @yzhou601 for some of the initial code that was leveraged!)

PR #496, opened 2020-09-24
Update BuildResHPXML to take random seed and building ID as arguments

dshen109:
## Pull Request Description

Add in random_seed and building_id arguments to the BuildResidentialHPXML measure and GenerateSchedule

## Checklist

Not all may apply:


joseph-robertson:
@dshen109,

I feel like you could achieve most of this without having to modify the BuildResHPXML measure. In the [URBANopt meta measure](https://github.com/urbanopt/urbanopt-example-geojson-project/blob/build-res-hpxml-v3/example_project/measures/BuildResidentialModel/measure.rb), you would set "Building ID" on the building object's additional properties. See [ResStock's meta measure](https://github.com/NREL/OpenStudio-BuildStock/blob/master/measures/BuildExistingModel/measure.rb#L91) as an example. You may have to pass the Building ID thru from the baseline mapper into BuildResModel, like around here somewhere: https://github.com/urbanopt/urbanopt-example-geojson-project/blob/build-res-hpxml-v3/example_project/mappers/Baseline.rb#L667.

Let me know if this doesn't make sense.

PR #520, opened 2020-10-20
BuildResidentialHPXML measure: Add arguments to set the RNG for stochastic schedules

dshen109:
## Pull Request Description

Add in schedules_random_seed as an optional argument to the BuildResidentialHPXML measure to seed the RNG in `schedules.rb`

## Checklist

Not all may apply:


joseph-robertson:
I think I like the idea of this. Outside of ResStock, stochastic schedules would always be generated with seed=1. With this change, you are able to assign a seed other than 1. And, ResStock's BuildExistingModel would not have to set its seed (i.e., Building ID) on the additional properties object anymore; it could be passed directly into the BuildResidentialHPXML measure.

joseph-robertson (approved):
👍

PR #548, opened 2020-12-05
Fix possible neighbor shading error

### LINKED ISSUE #547: Neighbor building azimuth error excludes ExteriorAdjacentTo=other housing unit
Consider the attached xml file (created from testing project of OS-BuildStock) describing an "apartment unit": 
[existing.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/5644293/existing.xml.txt)

It throws: `[10:20:21.384212 ERROR] A neighbor building has an azimuth (158) not equal to the azimuth of any wall.`

It has a neighbor with azimuth=158, yet it indeed has a wall (Surface_4) with azimuth=158.

Are walls adjacent to "other housing unit" being excluded from this check?

I think this is probably related to [this newly introduced change](https://github.com/NREL/OpenStudio-HPXML/blob/build-res-hpxml-v3/BuildResidentialHPXML/measure.rb#L4104).

Should units of single-family attached/multifamily buildings never have neighbors?
Should we not store azimuth on walls that have ExteriorAdjacentTo=other housing unit?
Should you never have a neighbor with the same azimuth as an adiabatic wall?

shorowit:
## Pull Request Description

Fixes possible "A neighbor building has an azimuth (XXX) not equal to the azimuth of any wall" error if the neighbor is in the same direction as the SFA/MF dwelling unit's adiabatic side. Closes #547.

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/measure.rb
+++ b/HPXMLtoOpenStudio/measure.rb
@@ -1637,6 +1637,7 @@ class OSModel
     floor_surface.setSpace(create_or_get_space(model, spaces, HPXML::LocationLivingSpace))
     floor_surface.setOutsideBoundaryCondition('Adiabatic')
     floor_surface.additionalProperties.setFeature('SurfaceType', 'InferredFloor')
+    floor_surface.additionalProperties.setFeature('Tilt', 0.0)
 
     # Add ceiling surface
     ceiling_surface = OpenStudio::Model::Surface.new(add_ceiling_polygon(-floor_width, -floor_length, z_origin), model)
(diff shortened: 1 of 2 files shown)


shorowit:
@joseph-robertson This should fix your error. Note however that the neighbors have no impact on simulation results for this HPXML because adiabatic surfaces are already assumed to be completely shaded. It's fine to have, but not needed. (If a SFA/MF unit has neighbors to a side where the wall is _not_ adiabatic, that would of course affect the results.)

joseph-robertson (approved):
This fixed my tests on OS-BuildStock.

PR #563, opened 2020-12-18
Bugfixes for basement adjacent surfaces

yzhou601:
## Pull Request Description

There're some bugs in eliminating basement interior surface solar absorptance:
1. Subsurfaces in conditioned basement (window modeled as door as current workaround) are not modified
2. Those basement surfaces with adjacent surfaces are not handled perfectly. The properties of adjacent surfaces are not modified correspondingly, while OpenStudio expects these two adjacent surfaces are of symmetric construction.
3. For those surfaces exterior adjacent to conditioned basement, their adjacent surfaces are supposed to be put into conditioned basement surface list instead of the surfaces themselves.

## Checklist

Not all may apply:

- [ ] EPvalidator.xml has been updated
- [ ] Tests (and test files) have been updated
- [ ] Documentation has been updated
- [ ] `openstudio tasks.rb update_measures` has been run
- [ ] No unexpected regression test changes on CI

PR #564, opened 2020-12-21
Fail on OS Forward Translator errors

shorowit:
## Pull Request Description

It is possible for the OpenStudio Forward Translator (FT) to have errors but still operate successfully. When this happens, the simulation results may not be correct. Thus we now fail the workflow when such errors are detected, so that we can identify and fix the root cause instead and prevent such errors from occurring in the first place.

## Checklist

Not all may apply:


shorowit:
@yzhou601 Can you commit a new derivative HPXML test file that causes FT errors to this PR? Then we can verify that this branch fails while your [new branch](https://github.com/NREL/OpenStudio-HPXML/pull/563) addresses the issue.

yzhou601:
> @yzhou601 Can you commit a new derivative HPXML test file that causes FT errors to this PR? Then we can verify that this branch fails while your [new branch](https://github.com/NREL/OpenStudio-HPXML/pull/563) addresses the issue.

@shorowit Do you mean a automatically configured test file or just a manual test file?

shorowit:
> @shorowit Do you mean a automatically configured test file or just a manual test file?

Let's do an automatically configured test file.

PR #565, opened 2020-12-23
Allow outputs in JSON instead of CSV format

shorowit:
## Pull Request Description

Adds the ability to specify the file format of annual output (and timeseries output, if requested): either CSV or JSON.

If using the run_simulation.rb script, it can be specified using `--output-format csv` or `--output-format json`. If this argument is not provided, it will continue to default to CSV.

If running the OpenStudio measures via an OpenStudio workflow, it can be specified using the `output_format` argument in the SimulationOutputReport reporting measure.

**Breaking change**: For CSV outputs, the first two sections in the results_annual.csv file are now prefixed with "Fuel Use:" and "End Use:", respectively.

## Checklist

Not all may apply:


nmerket (approved):
I have implemented this in https://github.com/NREL/hpxml-energyplus-api/pull/15. Seems to be working as expected.

PR #568, opened 2021-01-04
New window/skylight interior shading approach

shorowit:
## Pull Request Description

Switches from using a `WindowMaterial:Shade` to using a `Shading:Building:Detailed` (in combination with adjustments to the window view factor to address ground diffuse transmittance). 

The former object is _far_ more detailed than a simple reduction in transmittance (based on an interior shading multiplier), as it includes angular effects, re-reflected/absorbed solar, convection/radiation effects, etc., and thus was not having the desired effect. The new approach using the Shading:Building:Detailed object nearly perfectly achieves the desired result without all of these complications. The only drawback to this approach is the view factor cannot change throughout the year so we cannot perfectly adjust the summer vs winter ground diffuse (but this should be a small impact and can be addressed by a minor future EnergyPlus enhancement).

## Checklist

Not all may apply:


shorowit:
Tested by @bpark1327; compared against making solar modifications to a weather file. Results look good. 

Will do one more look over this code and CI results before merging.

PR #580, opened 2021-01-08
BuildResidentialHPXML: Single flue argument

joseph-robertson:
## Pull Request Description

Removes:
* `heating_system_has_flue_or_chimney`
* `heating_system_has_flue_or_chimney_2`
* `water_heater_has_flue_or_chimney`

Adds:
* `geometry_has_flue_or_chimney`

## Checklist

Not all may apply:


BuildResidentialHPXML/measure.rb
```diff
@@ -3368,12 +3356,7 @@ def self.set_building_construction(hpxml, runner, args)
     hpxml.building_construction.conditioned_building_volume = conditioned_building_volume
     hpxml.building_construction.average_ceiling_height = args[:geometry_wall_height]
     hpxml.building_construction.residential_facility_type = args[:geometry_unit_type]
-
-    if (args[:heating_system_type] != 'none' && args[:heating_system_has_flue_or_chimney]) ||
-       (args[:heating_system_type_2] != 'none' && args[:heating_system_has_flue_or_chimney_2]) ||
-       (args[:water_heater_type] != 'none' && args[:water_heater_has_flue_or_chimney])
-      hpxml.building_construction.has_flue_or_chimney = true
-    end
+    hpxml.building_construction.has_flue_or_chimney = args[:geometry_has_flue_or_chimney] if args[:geometry_has_flue_or_chimney]
```
shorowit, line 3359: Why are you only assigning the value when true?

joseph-robertson: If `FlueOrChimney=False` in the hpxml, then they don't align with the corresponding xml test files.

shorowit: Then you either need to skip checking that element in the tests or allow it to be auto. What you've done above is wrong, because if someone entered false, the HPXML file would not include the element so it would be defaulted in HPXMLtoOS, and it could be defaulted to true.

joseph-robertson: Oh, I thought it always defaulted to false. Seems like it was wrong before this change then, too. I will try the auto route.

shorowit: True, it does look wrong before too. 

It shouldn't matter if it was always defaulted to false or not. In general, we should only leave values out of the HPXML when they are being defaulted. If someone enters a value, that value should always be reflected in the HPXML. Anything other behavior would be unexpected.

shorowit (approved):

PR #594, opened 2021-01-21
*Multiple* dehumidifiers

shorowit:
## Pull Request Description

Previously only one dehumidifier was allowed, now multiple are allowed. Note that the RH setpoint across all dehumidifiers must be the same.

## Checklist

Not all may apply:


yzhou601 (approved):
This looks great! Checked multiple dehumidifier IDF, it showed desired diffs compared with single dehumidifier model.

PR #597, opened 2021-01-26
OS v3.2 - EMS actuator for view factor to ground

bpark1327:
## Pull Request Description

Add an EMS program using the OpenStudio SDK, replacing the existing view factor code. This allows the solar from ground reflectance to be properly adjusted for windows or skylights with interior/exterior shading.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -1853,9 +1853,24 @@ def self.apply_shading(model, window_or_skylight, shading_polygon, parent_surfac
...
-      avg_trans_value = trans_values.sum(0.0) / 12.0 # FUTURE: Create EnergyPlus actuator to adjust this
-      default_vf_to_ground = ((1.0 - Math::cos(parent_surface.tilt)) / 2.0).round(2)
-      sub_surface.setViewFactortoGround(default_vf_to_ground * avg_trans_value)
+      actuator = OpenStudio::Model::EnergyManagementSystemActuator.new(sub_surface, *EPlus::EMSActuatorSurfaceViewFactorToGround)
+      actuator.setName("#{sub_surface.name.to_s.gsub(' ', '_').gsub('-', '_')}_actuator")
+
+      # EMS to actuate schedule
+      program = OpenStudio::Model::EnergyManagementSystemProgram.new(model)
+      program.setName("#{sub_surface.name.to_s.gsub('-', '_')} View Factor To Ground Program")
+      program.addLine("IF (DayOfYear >= 1) && (DayOfYear <= 120)") # FIXME: Is there a way to parameterize DayOfYear so that I can avoid hard-coding it?
```
bpark1327, line 1862: @shorowit This EMSProgram is to overwrite the `view factor to ground` of the subsurfaces. I've hardcoded `DayOfYear` for the first pass but I'd like to avoid it. Any ideas on how to parameterize `DayOfYear`? 

If we determined to parameterize `DayOfYear`, we probably need some logic to find the nearest day that’s a multiple of 20 in order to synchronize with the default period frequency (i.e., 20 days) of the detailed diffuse algorithm.

shorowit: I suggest using the schedule value from the `trans_sch` in your EMS program instead, using an EMS sensor. The schedule already incorporates the summer vs winter seasons. If you search the code, you'll find many examples of EMS programs that use sensors of schedules.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1853,9 +1853,24 @@ def self.apply_shading(model, window_or_skylight, shading_polygon, parent_surfac
...
+      program.addLine("Set #{actuator.name} = #{((1.0 - Math::cos(sub_surface.tilt)) / 2.0).round(2)}*#{sf_winter}") 
+      program.addLine("ELSEIF (DayOfYear > 120) && (DayOfYear <= 320)") # FIXME: use 320 days (11/17) instead of 304 days (11/1) in order to synchronize with the default "period" frequency of 20 days
+      program.addLine("Set #{actuator.name} = #{((1.0 - Math::cos(sub_surface.tilt)) / 2.0).round(2)}*#{sf_summer}") 
+      program.addLine("ELSEIF (DayOfYear > 320) && (DayOfYear <= 365)") 
+      program.addLine("Set #{actuator.name} = #{((1.0 - Math::cos(sub_surface.tilt)) / 2.0).round(2)}*#{sf_winter}")
```
shorowit, line 1867: I suggest keeping the "((1.0 - Math::cos(sub_surface.tilt)) / 2.0).round(2)" in a ruby variable and just setting the value in the EMS line. That will prevent EMS from having to do the exact same calculation for every timestep.

E.g.:
```
default_vf_to_ground = ((1.0 - Math::cos(parent_surface.tilt)) / 2.0).round(2)
...
program.addLine("Set #{actuator.name} = #{default_vf_to_ground}*#{sf_winter}") 
```

shorowit:
Pull in https://github.com/NREL/OpenStudio-HPXML/pull/667 and tests should pass?

shorowit:
@bpark1327 FYI, you should only pull in the os_v320 branch here, not master, since this branch is now based against os_v320. Since os_v320 is behind master, pulling in master brought in code that now shows up as changes relative to the os_v320 base, which makes it hard to see what is your code changes vs other code changes. I will pull master into os_v320 so these other code change differences go away, but figured I would explain what's happening here.

shorowit:
Pulled master in os_v320, then pulled os_v320 into here. Now we're back down to 15 changed files (from 23 before).

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1819,10 +1825,35 @@ def self.apply_shading(model, window_or_skylight, shading_polygon, parent_surfac
       end
       shading_surface.setTransmittanceSchedule(shading_schedules[trans_values].schedule)
 
-      # Adjustment to default view factor is used to reduce ground diffuse solar
-      avg_trans_value = trans_values.sum(0.0) / 12.0 # FUTURE: Create EnergyPlus actuator to adjust this
-      default_vf_to_ground = ((1.0 - Math::cos(parent_surface.tilt)) / 2.0).round(2)
-      sub_surface.setViewFactortoGround(default_vf_to_ground * avg_trans_value)
+      # EMS to actuate view factor to ground
+      actuator = OpenStudio::Model::EnergyManagementSystemActuator.new(sub_surface, *EPlus::EMSActuatorSurfaceViewFactorToGround)
+      actuator.setName("#{sub_surface.name.to_s.gsub(' ', '_').gsub('-', '_')}_actuator")
```
shorowit, line 1830: This is fragile since it is using the subsurface name directly but only handling a subset of characters that would break EMS. I was able to generate en E+ error by changing an HPXML window to "Window+North", which resulted in:
`Set Window+North_actuator = 0.5*fixedwindow_shading_coefficient, !- Program Line 1`
and
```
   ** Severe  ** EnergyManagementSystem:Actuator="WINDOW+NORTH_ACTUATOR", Invalid variable name entered.
   **   ~~~   ** ...Name; Names used as EMS variables cannot contain "+" characters.
```
Instead I suggest just using an index: window1_actuator, window2_actuator, etc. You'll have to update the tests too.

HPXMLtoOpenStudio/tests/test_constructions.rb
```diff
@@ -13,6 +13,35 @@ def sample_files_dir
     return File.join(File.dirname(__FILE__), '..', '..', 'workflow', 'sample_files')
   end
 
+  def _get_ems_values(ems_objects, name)
```
shorowit, line 16: Since we now have the same (or similar?) methods in multiple test_foo.rb files, can you move the method to util.rb so it can be shared across test files?

shorowit (commented):
This is really good work @bpark1327. Just a couple changes needed and then this should be ready.

HPXMLtoOpenStudio/tests/test_constructions.rb
```diff
@@ -57,10 +54,11 @@ def test_windows_shading
         assert_equal(sf_winter, winter_transmittance)
       end
 
-      # Check view factor for ground diffuse
-      default_view_factor = 0.5 # Default for vertical wall
-      view_factor = default_view_factor * (sf_summer * summer_length + sf_winter * winter_length) / 12.0
-      assert_in_delta(view_factor, os_window.viewFactortoGround.get, 0.001)
+      # Check subsurface view factor to ground
+      subsurface_view_factor = 0.5
+      os_window = model.getEnergyManagementSystemActuators.select { |w| w.actuatedComponent.get.name.to_s == window.id }[0]
```
shorowit, line 59: Name should be updated, this is no longer a window (subsurface) object.

HPXMLtoOpenStudio/tests/test_constructions.rb
```diff
@@ -108,10 +103,11 @@ def test_skylights_shading
         assert_equal(sf_winter, winter_transmittance)
       end
 
-      # Check view factor for ground diffuse
-      default_view_factor = 0.05 # Based on 6:12 pitch in HPXML
-      view_factor = default_view_factor * (sf_summer * summer_length + sf_winter * winter_length) / 12.0
-      assert_in_delta(view_factor, os_window.viewFactortoGround.get, 0.001)
+      # Check subsurface view factor to ground
+      subsurface_view_factor = 0.05 # 6:12 pitch
+      os_skylight = model.getEnergyManagementSystemActuators.select { |w| w.actuatedComponent.get.name.to_s == skylight.id }[0]
```
shorowit, line 108: Here too.

HPXMLtoOpenStudio/tests/test_constructions.rb
bpark1327: Oops. Yep I will change the name.

HPXMLtoOpenStudio/resources/util.rb
```diff
@@ -321,3 +321,33 @@ def self.fetch(uri_str, outfile, agent = 'curl/7.43.0', max_attempts = 10, timeo
     raise 'Too many http redirects' if attempts == max_attempts
   end
 end
+
+class UnitTestTools
```
bpark1327, line 325: @shorowit what do you think about this class name? Any suggestions?

shorowit: To be honest, I was picturing this method in a new test.rb file under /tests, as opposed to under /resources, since I can't imagine it would ever be used outside tests. No strong preference about class name; not sure it really needs a class for organization considering it's just a single method.

bpark1327: That makes sense. Will create a util.rb under tests. I agree that this method doesn't need to be in a class.

bpark1327:
I will compare the results before it gets merged.

bpark1327:
The figures below compare the sample file simulation energy consumption results obtained from `master` and this PR. The difference is minimal. For `base-enclosure-window-none.xml`, the result is exactly the same.
image
image

shorowit (approved):
This is ready to go once #667 is merged. Thanks @bpark1327!

PR #598, opened 2021-01-28
Always produce in.xml file for inspection of HPXML defaults

shorowit:
## Pull Request Description

Always produce the `in.xml` file for inspection of HPXML defaults used. Previously only generated if the "debug" argument was used, but the debug flag turns on other stuff that may slow down the runtime, so this allows access to in.xml without impacting runtime. **Breaking change**: The HPXMLtoOpenStudio measure argument "output_dir" is now required.

## Checklist

Not all may apply:


shorowit:
@joseph-robertson Does this seem okay to you (note the breaking change)? It should allow ResStock to calculate cost multipliers off the in.xml without having to use the debug flag.

joseph-robertson (approved):

PR #602, opened 2021-01-29
BuildResidentialHPXML measure: Consolidate efficiency arguments

### LINKED ISSUE #601: BuildResidentialHPXML measure: Combine `clothes_washer` efficiency arguments into single argument
https://github.com/NREL/OpenStudio-HPXML/blob/6def3dca9f68ec87d7bc068d2d5991a5fd9c2a76/BuildResidentialHPXML/measure.rb#L2267-L2279

joseph-robertson:
## Pull Request Description

Consolidates many arguments:
- cooling_system_cooling_efficiency (seer and eer)
- heat_pump_heating_efficiency (hspf and cop)
- heat_pump_cooling_efficiency (seer and eer)
- dishwasher_efficiency (ef and kwh)
- clothes_washer_efficiency (mef and imef; closes #601)
- mech_vent_recovery_efficiency_type

## Checklist

Not all may apply:


shorowit:
@joseph-robertson This is a good change and was probably an oversight that we hadn't previously combined the arguments.

It makes me think that we should go ahead and do this across _all_ the other technologies. I think at one point we maybe had separate inputs for some some efficiency values because they might have different default values/units. But now I'm thinking that just makes things more complicated. How would you feel about combining arguments for the below list?
- cooling_system_cooling_efficiency (seer and eer)
- heat_pump_heating_efficiency (hspf and cop)
- heat_pump_cooling_efficiency (seer and eer)
- dishwasher_efficiency (ef and kwh)
- mech_vent_recovery_efficiency_type (combine total and sensible, no need to allow adjusted for one and unadjusted for the other)
If the different types being combined have different units, we can just describe the combined units as "FOO or BAR".

BuildResidentialHPXML/measure.rb
```diff
@@ -1025,18 +1029,17 @@ def arguments(model)
...
 
-    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('cooling_system_cooling_efficiency_eer', true)
-    arg.setDisplayName('Cooling System: Rated EER')
-    arg.setUnits('EER')
-    arg.setDescription("The rated efficiency value of the #{HPXML::HVACTypeRoomAirConditioner} cooling system.")
-    arg.setDefaultValue(8.5)
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('cooling_system_cooling_efficiency', true)
+    arg.setDisplayName('Cooling System: Efficiency')
+    arg.setUnits("#{HPXML::UnitsSEER} or #{HPXML::UnitsEER}")
+    arg.setDescription('The rated efficiency value of the cooling system.')
```
shorowit, line 1041: Need to retain which system types use SEER and which use EER in the argument description.

BuildResidentialHPXML/measure.rb
```diff
@@ -1104,32 +1111,30 @@ def arguments(model)
...
 
-    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heat_pump_heating_efficiency_cop', true)
-    arg.setDisplayName('Heat Pump: Rated Heating COP')
-    arg.setUnits('COP')
-    arg.setDescription("The rated heating efficiency value of the #{HPXML::HVACTypeHeatPumpGroundToAir} heat pump.")
-    arg.setDefaultValue(3.6)
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heat_pump_heating_efficiency', true)
+    arg.setDisplayName('Heat Pump: Heating Efficiency')
+    arg.setUnits("#{HPXML::UnitsHSPF} or #{HPXML::UnitsCOP}")
+    arg.setDescription('The rated heating efficiency value of the heat pump.')
```
shorowit, line 1123: Same comment as above.

BuildResidentialHPXML/measure.rb
```diff
@@ -1104,32 +1111,30 @@ def arguments(model)
...
     args << arg
 
-    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heat_pump_cooling_efficiency_eer', true)
-    arg.setDisplayName('Heat Pump: Rated Cooling EER')
-    arg.setUnits('EER')
-    arg.setDescription("The rated cooling efficiency value of the #{HPXML::HVACTypeHeatPumpGroundToAir} heat pump.")
-    arg.setDefaultValue(16.6)
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heat_pump_cooling_efficiency', true)
+    arg.setDisplayName('Heat Pump: Cooling Efficiency')
+    arg.setUnits("#{HPXML::UnitsSEER} or #{HPXML::UnitsEER}")
```
shorowit, line 1135: Same as above.

BuildResidentialHPXML/measure.rb
```diff
@@ -1418,7 +1423,7 @@ def arguments(model)
     arg.setDefaultValue(24)
     args << arg
 
-    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('mech_vent_total_recovery_efficiency_type', mech_vent_recovery_efficiency_type_choices, true)
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('mech_vent_recovery_efficiency_type', mech_vent_recovery_efficiency_type_choices, true)
```
shorowit, line 1426: GitHub won't let me comment below, but need to add to the total/sensible recovery efficiency argument descriptions which mech vent system types they apply to.

shorowit (commented):
Looks good, just a few minor comments.

BuildResidentialHPXML/measure.rb
joseph-robertson: Actually, why do we need the `cooling_system_cooling_efficiency_type` argument at all if the efficiency type corresponds to the system type?

BuildResidentialHPXML/measure.rb
joseph-robertson: Same question as above.

BuildResidentialHPXML/measure.rb
joseph-robertson: Same.

BuildResidentialHPXML/measure.rb
shorowit: I think it's safer. For a central air conditioner, if someone entered its EER instead of SEER, they would get an incorrect simulation result if we removed the argument and not even realize that the value was treated as a SEER. But as it stands now, if they entered EER instead of SEER, they would get an error that SEER is expected.

BuildResidentialHPXML/measure.rb
```diff
@@ -1025,18 +1029,17 @@ def arguments(model)
...
 
-    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('cooling_system_cooling_efficiency_eer', true)
-    arg.setDisplayName('Cooling System: Rated EER')
-    arg.setUnits('EER')
-    arg.setDescription("The rated efficiency value of the #{HPXML::HVACTypeRoomAirConditioner} cooling system.")
-    arg.setDefaultValue(8.5)
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('cooling_system_cooling_efficiency', true)
+    arg.setDisplayName('Cooling System: Efficiency')
+    arg.setUnits("#{HPXML::UnitsSEER} or #{HPXML::UnitsEER}")
+    arg.setDescription("The rated efficiency value of the cooling system. System types #{HPXML::HVACTypeCentralAirConditioner} and #{HPXML::HVACTypeMiniSplitAirConditioner} use #{HPXML::UnitsSEER}. System type #{HPXML::HVACTypeRoomAirConditioner} uses #{HPXML::UnitsEER}.")
```
shorowit, line 1041: The description of which system types use which efficiency metrics should really go with the above argument, not this one. Same comment presumably applies to some arguments below.

Also, it's probably confusing to a user that it doesn't say anything about evaporative coolers. We should say that these arguments are ignored for them. (Alternatively we could make the arguments optional, but that makes me a bit nervous.)

shorowit (approved):

PR #608, opened 2021-02-03
Switch gshp cooling curve generated from more consistent rated conditions

yzhou601:
## Pull Request Description

The GSHP cooling performance coefficients in Tang's thesis are generated without using a standard rated condition, instead a rule of thumb is adopted where the largest capacity point is taken as rated. This caused inconsistency(large diff) compared with previous implementation. 
This PR summarized the work of regenerating the curve coefficients following Tang's approach and switched our workflow to use the new curve.

## Checklist

Not all may apply:


yzhou601:
image

@shorowit I am implementing the curve rated at 80F 1200cfm. Is this the optimal one we want to use?

shorowit:
> @shorowit I am implementing the curve rated at EWT: 80F AFR: 1200cfm. Is this the optimal one we want to use?

The results are so similar between 1200cfm and 975cfm that it doesn't matter match. What you've done seems fine.

yzhou601:
CI only shows diffs on test files with ground source heat pump, and the cooling energy usage is reduced compared to previous results, which is consistent with our observation from HEScore test file.

shorowit:
I just checked HEScore. Compared to master, cooling changes by < 0.5%. (With the previous Tang implementation, cooling changed by something like 84%.)

This looks good to me. Any reason not to merge?

yzhou601:
> I just checked HEScore. Compared to master, cooling changes by < 0.5%. (With the previous Tang implementation, cooling changed by something like 84%.)
> 
> This looks good to me. Any reason not to merge?

No, go ahead! 🚀

shorowit (approved):

PR #629, opened 2021-03-02
Fix multiple vented crawlspace/attic

yzhou601:
## Pull Request Description

Not all vented crawlspaces or vented attics are defaulted with an sla, which might cause error in further calculation (eg. hvac_sizing)

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -230,40 +230,57 @@ def self.apply_infiltration(hpxml)
...
-    if vented_attic.vented_attic_sla.nil? && vented_attic.vented_attic_ach.nil?
-      vented_attic.vented_attic_sla = Airflow.get_default_vented_attic_sla()
+    vented_attics.each do |vented_attic|
+      next unless vented_attic.vented_attic_sla.nil? && vented_attic.vented_attic_ach.nil?
+      if not default_ach.nil? # ACH specified
+        vented_attic.vented_attic_ach = default_ach
+      else # Use SLA
+        vented_attic.vented_attic_sla = default_sla
+      end
       vented_attic.vented_attic_sla_isdefaulted = true
```
yzhou601, line 257: @shorowit If there's ach/sla specified in any attic, then other attics will be defaulted as the same ach/sla. Otherwise we call default sla method. In the first case, do we want the `vented_attic_sla_isdefaulted` to be true for attics except for the one that was explicitly specified?

shorowit: Hmm. I'm wondering if we should simplify these types of questions by changing EPvalidator to allow multiple ventilation rates (one per attic) to be specified, and then perform defaulting on a per-attic basis. Anywhere in our model (airflow.rb, hvac_sizing.rb), we'd simply use a floor-area-weighted average value. What do you think?

yzhou601: Sounds fair to me.

yzhou601: One question for vented attic, it seems both ach and sla inputs are allowed for vented attic (while only sla was requested in [hvac_sizing](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L2577-L2578), is this intended? Or is ach only possible in some specific cases?). So if there're multiple vented attics of which some are specifying ventilation rates in ach and some are in sla, and some are not specifying anything (defaulted). What do you think is a better way to average them? One possible approach is to convert them all into either sla or ach form then do averaging.

yzhou601: Proceed with allowing one VentilationRate

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2568,16 +2568,21 @@ def self.get_space_ua_values(space_type, weather)
...
         vented_attic = @hpxml.attics.select { |f| f.attic_type == HPXML::AtticTypeVented }[0]
-        sla = vented_attic.vented_attic_sla
+        if not vented_attic.vented_attic_sla.nil?
+          sla = vented_attic.vented_attic_sla
+        else
+          ach = vented_attic.vented_attic_ach
+        end
       end
-      ach = Airflow.get_infiltration_ACH_from_SLA(sla, 8.202, weather)
+      ach = Airflow.get_infiltration_ACH_from_SLA(sla, 8.202, weather) if ach.nil?
```
yzhou601, line 2585: @shorowit I figured out that this is actually not a bug. This method `calculate_space_design_temps`->`get_space_ua_values` will never be called when space_type = HPXML::LocationAtticVented. Its space design temperatures are calculated/defined in another methodology. For this `get_space_ua_values`, however, I am not sure whether it will be used in scope out of `calculate_space_design_temps` in future, so I retained its logic for `HPXML::LocationAtticVented` and added ach just in case.

shorowit (approved):
Looks good. Thanks!

Issue #618, opened 2021-02-12
Additional HVAC technologies


shorowit:
Allow more HVAC technologies:
- [ ] @yzhou601 PTAC/PTHP (HPXML inputs; performance map; 1, 2, variable speed; HVAC sizing)
- [ ] Electric radiant heating
- [ ] Steam boiler
- [ ] Ground-to-water heat pump (GSHP w/ hydronic distribution)
- [ ] Air-to-water heat pump
- [ ] [Absorption (e.g., gas-fired) heat pump](https://www.energy.gov/energysaver/heat-pump-systems/absorption-heat-pumps)
- [ ] Direct and direct-indirect evaporative coolers; expose effectiveness input?
- [ ] GSHP w/ horizontal slinky loop
- [ ] Commercial boiler (allow Thermal Efficiency (TE or Percent?) instead of AFUE)
- [ ] In-unit chiller (e.g., w/ baseboard or fan coil)

shorowit:
Will open separate tickets for each of these.

PR #650, opened 2021-03-10
PTAC/PTHP

yzhou601:
## Pull Request Description

Add capability to model PTAC and PTHP (#618).
Clean up hvac methods to make different system types more consistent in their modeling approach (curve, E+ object, etc).

## Checklist

Not all may apply:


yzhou601:
@shorowit I ran tests for room ac system where PTAC object is replaced by unitary system object, see the comparison:
image
The summary is:
1.  The unitary system combination is showing a very minor diff
2. Overall the minor diff is caused by a residual load being resolved by our ideal loads air system so the unitary system approach showed a slightly lower energy use. The output csv file shows this residual load existing in every hour where cooling loads are non-zero, so it looked to me that it can be caused by small convergence tolerance etc. in E+ airloop simulation.
3.  I think this minor diff is negligible, this results looks promising to our idea of combining these method and making our models more consistent.

shorowit:
@yzhou601 Great! Definitely looks good enough to continue w/ the unitary system. That's odd that there are slightly non-zero unmet loads for the unitary system, but not large enough to be concerning.

shorowit:
1. See if energy change due to different curves in @yzhou601's handful of test files is substantially caused by oversized capacity.
2. Consider creating a new branch where we change sample files to have more reasonable fixed capacities.
3. Perform window AC runs using HEScore historic buildings to understand changes.

yzhou601:
The remaining CI diffs are caused by curve variable min/max boundary changes

workflow/tests/hpxml_translator_test.rb
```diff
@@ -642,9 +643,16 @@ def _verify_outputs(rundir, hpxml_path, results, hpxml)
       if hpxml.hvac_distributions.select { |d| d.air_type.to_s == HPXML::AirTypeFanCoil }.size > 0
         next if err_line.include? 'In calculating the design coil UA for Coil:Cooling:Water' # Warning for unused cooling coil for fan coil
       end
+      if hpxml.cooling_systems.select { |c| c.cooling_system_type == HPXML::HVACTypeRoomAirConditioner }.size > 0
+        next if err_line.include? 'AirLoopHVAC:UnitarySystem = ROOM AC UNITARY SYSTEM' # Warning for unused heating coil
+      end
```
shorowit, line 648: Why are we now including a heating coil for a room AC? Is it possible to exclude it?

tasks.rb
```diff
@@ -401,6 +404,8 @@ def create_hpxmls
     'base-hvac-undersized.xml' => 'base.xml',
     'base-hvac-undersized-allow-increased-fixed-capacities.xml' => 'base-hvac-undersized.xml',
     'base-hvac-wall-furnace-elec-only.xml' => 'base.xml',
+    'base-hvac-ptac-electric-resistance.xml' => 'base.xml',
```
shorowit, line 407: I suggest splitting this up into two files:
- base-hvac-ptac.xml
- base-hvac-ptac-cooling-only.xml

Since the AHRI directory showed that the vast majority of PTACs include electric resistance heating, I'm suggesting that that be assumed in the base-hvac-ptac.xml file.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -631,8 +643,8 @@
   </sch:pattern>
 
   <sch:pattern>
-    <sch:title>[CoolingSystemType=RoomAC]</sch:title>
-    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/h:CoolingSystem[h:CoolingSystemType="room air conditioner"]'>
+    <sch:title>[CoolingSystemType=RoomAC or CoolingSystemType=PTAC]</sch:title>
```
shorowit, line 646: Let's just shorten here (and other references) to "[CoolingSystemType=RoomAC or PTAC]"

tasks.rb
shorowit: I am hoping that the PTAC with cooling only will not need a `HeatingSystem` defined in the HPXML, is that right? (In contrast to forcing the user to have to create such a system but with capacity and fraction load served set to zero.)

shorowit (commented):
I haven't gotten through the PR in detail, but here are a few initial comments while scanning it.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -446,6 +446,18 @@
...
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/h:HeatingSystem[h:HeatingSystemType/h:PackagedTerminalAirConditionerHeating]'>
+      <sch:assert role='ERROR' test='count(h:DistributionSystem) = 0'>Expected 0 element(s) for xpath: DistributionSystem</sch:assert>
+      <sch:assert role='ERROR' test='count(h:HeatingSystemFuel) = 1'>Expected 1 element(s) for xpath: HeatingSystemFuel</sch:assert>
+      <sch:assert role='ERROR' test='h:HeatingSystemFuel[text()="electricity"] or not(h:HeatingSystemFuel)'>Expected HeatingSystemFuel to be 'electricity'</sch:assert>
+      <sch:assert role='ERROR' test='count(h:HeatingCapacity) &lt;= 1'>Expected 0 or 1 element(s) for xpath: HeatingCapacity</sch:assert>
+      <sch:assert role='ERROR' test='count(h:AttachedToCoolingSystem) = 1'>Expected 1 element(s) for xpath: AttachedToCoolingSystem</sch:assert>
+      <sch:assert role='ERROR' test='count(../h:CoolingSystem/h:CoolingSystemType[text()="packaged terminal air conditioner"]) &gt;= 1'>Expected 1 or more element(s) for xpath: ../CoolingSystem/CoolingSystemType[text()="packaged terminal air conditioner"]</sch:assert>
```
shorowit, line 458: Nice check on cooling system type.

shorowit: I suggest exposing the heating efficiency (%) as an input rather than hard-coding it in hvac.rb.

HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -724,8 +736,8 @@
...
       <sch:assert role='ERROR' test='count(h:SystemIdentifier) = 1'>Expected 1 element(s) for xpath: SystemIdentifier</sch:assert>
       <sch:assert role='ERROR' test='count(../../h:HVACControl) = 1'>Expected 1 element(s) for xpath: ../../HVACControl</sch:assert> <!-- See [HVACControl] -->
-      <sch:assert role='ERROR' test='count(h:HeatPumpType) = 1'>Expected 1 element(s) for xpath: HeatPumpType</sch:assert> <!-- See [HeatPumpType=AirSource] or [HeatPumpType=MiniSplit] or [HeatPumpType=GroundSource] or [HeatPumpType=WaterLoop] -->
-      <sch:assert role='ERROR' test='h:HeatPumpType[text()="air-to-air" or text()="mini-split" or text()="ground-to-air" or text()="water-loop-to-air"] or not(h:HeatPumpType)'>Expected HeatPumpType to be 'air-to-air' or 'mini-split' or 'ground-to-air' or 'water-loop-to-air'</sch:assert>
+      <sch:assert role='ERROR' test='count(h:HeatPumpType) = 1'>Expected 1 element(s) for xpath: HeatPumpType</sch:assert> <!-- See [HeatPumpType=AirSource] or [HeatPumpType=MiniSplit] or [HeatPumpType=GroundSource] or [HeatPumpType=WaterLoop] or [HeatPumpType=WaterLoop] -->
```
shorowit, line 739: You accidentally added a second "[HeatPumpType=WaterLoop]" here instead of "[HeatPumpType=PTHP]". Also, it looks like you didn't actually add the code below that defines the PTHP inputs (e.g., EER/COP, SHR, capacity, etc.).

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -2656,13 +2659,28 @@ def distribution_system
     end
 
     def attached_cooling_system
-      return if distribution_system.nil?
+      return if distribution_system.nil? && @cooling_system_idref.nil?
+
+      # by heating system's idref
+      if not @cooling_system_idref.nil?
+        if @heating_system_type != HVACTypePTACHeating
+          fail "Heating system type: #{@heating_system_type} should not be attached to a cooling system. Only PTAC allowed."
```
shorowit, line 2667: Can you add a test file and expected error message for this?

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -562,6 +563,9 @@ def self.apply_hvac(hpxml, weather, convert_shared_systems)
       elsif heat_pump.heat_pump_type == HPXML::HVACTypeHeatPumpGroundToAir
         heat_pump.cooling_shr = 0.732
         heat_pump.cooling_shr_isdefaulted = true
+      elsif heat_pump.heat_pump_type == HPXML::HVACTypeHeatPumpPTHP
+        heat_pump.cooling_shr = 0.65 # Fixme: is it ok to use the same default as roomac and ptac?
```
shorowit, line 567: That seems reasonable to me. You can remove the comment.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1,98 +1,135 @@
...
+          obj_name = Constants.ObjectNameAirSourceHeatPump
+        elsif cooling_system.heat_pump_type == HPXML::HVACTypeHeatPumpMiniSplit
+          obj_name = Constants.ObjectNameMiniSplitHeatPump
+        elsif cooling_system.heat_pump_type == HPXML::HVACTypeHeatPumpPTHP
+          obj_name = Constants.ObjectNamePTHP
+          fan_watts_per_cfm = 0.0
+        else
+          fail "Unexpected heat pump type: #{cooling_system.heat_pump_type}."
+        end
+      else
```
shorowit, line 24: Use `elsif cooling_system.is_a? HPXML::CoolingSystem` here just to be safe.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1,98 +1,135 @@
...
+        end
+      end
+      # Cooling Coil
       clg_coil = create_dx_cooling_coil(model, obj_name, cooling_system)
       hvac_map[cooling_system.id] << clg_coil
+
+      clg_ap = cooling_system.additional_properties
+      num_speeds = clg_ap.num_speeds
+      clg_cfm = cooling_system.cooling_airflow_cfm
+    else
```
shorowit, line 54: Use `elsif heating_system.is_a? HPXML::HeatingSystem` here just to be safe.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1,98 +1,135 @@
...
+      obj_name = Constants.ObjectNameFurnace
+      num_speeds = 1
```
shorowit, line 56: Wrap this inside an `if heating_system.heating_system_type == HPXML::HVACTypeFurnace` just to be safe.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3060,7 +2905,9 @@ def self.set_heat_c_d(heating_system, num_speeds)
     htg_ap = heating_system.additional_properties
 
     # Degradation coefficient for heating
-    if num_speeds == 1
+    if (heating_system.is_a? HPXML::HeatPump) && (heating_system.heat_pump_type == HPXML::HVACTypeHeatPumpPTHP)
+      htg_ap.heat_c_d = 0.22 # Fixme: Is it correct to use the same c_d with cooling?
```
shorowit, line 2909: @jonwinkler Do you have a c_d recommendation for the heating side of a PTHP?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2946,14 +2786,17 @@ def self.create_dx_heating_coil(model, obj_name, heating_system)
...
+        else # PTHP
+          htg_coil.setRatedCOP(heating_system.heating_efficiency_cop)
+        end
+        if (htg_type != HPXML::HVACTypeHeatPumpPTHP) && (not htg_ap.crankcase_temp.nil?)
           htg_coil.setMaximumOutdoorDryBulbTemperatureforCrankcaseHeaterOperation(UnitConversions.convert(htg_ap.crankcase_temp, 'F', 'C'))
         end
         htg_coil.setRatedTotalHeatingCapacity(UnitConversions.convert(heating_system.heating_capacity, 'Btu/hr', 'W'))
         htg_coil.setRatedAirFlowRate(calc_rated_airflow(heating_system.heating_capacity, htg_ap.heat_rated_cfm_per_ton[0], 1.0))
-      else
+      else # Fixme: two speed PTHP?
```
shorowit, line 2799: What is the question?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4142,6 +4000,8 @@ def self.get_hpxml_hvac_systems(hpxml)
       heating_system = nil
       if is_central_air_conditioner_and_furnace(hpxml, cooling_system.attached_heating_system, cooling_system)
         heating_system = cooling_system.attached_heating_system
+      elsif is_ptac(hpxml, cooling_system.attached_heating_system, cooling_system)
+        heating_system = cooling_system.attached_heating_system
```
shorowit, line 4004: I'm confused as to why you don't have to skip the PTACHeating system inside the heating system loop below (like we do for AC+furnace).

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4132,6 +3979,17 @@ def self.is_central_air_conditioner_and_furnace(hpxml, heating_system, cooling_s
     return true
   end
 
+  def self.is_ptac(hpxml, heating_system, cooling_system)
```
shorowit, line 3982: Perhaps it would make more sense to rename the `is_central_air_conditioner_and_furnace` method to something generic (like `is_attached_heating_and_cooling_systems`) and just incorporate this logic into it?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -1539,8 +1541,8 @@ def self.apply_hvac_equipment_adjustments(hvac_sizing_values, weather, hvac, cfa
       end
       hvac_sizing_values.Heat_Airflow = calc_airflow_rate(hvac_sizing_values.Heat_Capacity, (hvac.SupplyAirTemp - @heat_setpoint))
 
-    elsif hvac.HeatType == HPXML::HVACTypeHeatPumpMiniSplit
-
+    elsif [HPXML::HVACTypeHeatPumpMiniSplit, HPXML::HVACTypeHeatPumpPTHP].include? hvac.HeatType
+      # Please review: should PTHP call process_heat_pump_adjustment?
```
shorowit, line 1545: Yes, pretty sure it should. Feel free to remove this comment.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -1587,11 +1589,12 @@ def self.apply_hvac_equipment_adjustments(hvac_sizing_values, weather, hvac, cfa
 
       hvac_sizing_values.Heat_Airflow = calc_airflow_rate(hvac_sizing_values.Heat_Capacity, (hvac.SupplyAirTemp - @heat_setpoint))
 
-    elsif hvac.HeatType == HPXML::HVACTypeFurnace
+    elsif [HPXML::HVACTypeFurnace, HPXML::HVACTypePTACHeating].include? hvac.HeatType
 
       hvac_sizing_values.Heat_Capacity = hvac_sizing_values.Heat_Load
       hvac_sizing_values.Heat_Capacity_Supp = 0.0
 
+      # Is fan supposed to run at the same air flow rate in cooling and heating modes? Or a different air flow rate for each mode?
```
shorowit, line 1597: Different airflow rates seem appropriate. Feel free to remove this comment.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -1998,6 +2001,8 @@ def self.process_heat_pump_adjustment(hvac_sizing_values, weather, hvac, totalCa
         hvac_sizing_values.Cool_Airflow = cfm_per_btuh * hvac_sizing_values.Cool_Capacity
       elsif hvac.HeatType == HPXML::HVACTypeHeatPumpMiniSplit
         hvac_sizing_values.Cool_Airflow = hvac.RatedCFMperTonCooling[-1] * hvac.CapacityRatioCooling[-1] * UnitConversions.convert(hvac_sizing_values.Cool_Capacity, 'Btu/hr', 'ton')
+      elsif hvac.HeatType == HPXML::HVACTypeHeatPumpPTHP
+        hvac_sizing_values.Cool_Airflow = hvac.RatedCFMperTonCooling[hvac.SizingSpeed] * UnitConversions.convert(hvac_sizing_values.Cool_Capacity, 'Btu/hr', 'ton')
```
shorowit, line 2005: Hmm. In comparing this to the MSHP code right above, A) why use `hvac.SizingSpeed` instead of `-1` and B) why not use `hvac.CapacityRatioCooling`?

workflow/sample_files/base-hvac-ptac-electric-resistance.xml
```diff
@@ -0,0 +1,518 @@
...
+              <HeatingCapacity>36000.0</HeatingCapacity>
+              <FractionHeatLoadServed>1.0</FractionHeatLoadServed>
+              <AttachedToCoolingSystem idref='CoolingPTAC'/>
+            </HeatingSystem>
+            <CoolingSystem>
+              <SystemIdentifier id='CoolingPTAC'/>
+              <CoolingSystemType>packaged terminal air conditioner</CoolingSystemType>
+              <CoolingSystemFuel>electricity</CoolingSystemFuel>
+              <CoolingCapacity>24000.0</CoolingCapacity>
+              <CompressorType>single stage</CompressorType>
```
shorowit, line 322: Is compressor type an input? It's not specified according to EPvalidator. Please ensure consistency of EPvalidator inputs, sample file inputs, inputs used by the model, and documentation.

shorowit (commented):
It's nice to see so much code removed from hvac.rb!

When reviewing the PR, I got confused as to whether Compressor Type (single vs two vs variable speed) is an intended input or not. If it is, can it be defaulted? The defaults for central AC are based on SEER, so I'm not clear how the defaults are being set based on EER here. 

Please also review the PR checklist items, this PR still needs documentation & changelog updates as well as unit tests in test_hvac.rb (and test_defaults.rb if any inputs are being defaulted).

HPXMLtoOpenStudio/resources/EPvalidator.xml
yzhou601: I think it is ended up like this because we're not sure how the ptac heating efficiency will be rated (eg. unit % or afue). Do you suggest we could simply accept %  or both (regarding to afue and % are modeled in the same way)?

yzhou601: I guess we can proceed with % since we only support electric resistance as ptac heating for now.

shorowit: > I guess we can proceed with % since we only support electric resistance as ptac heating for now.

👍 That was going to be my suggestion.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: Sorry, should be quoted as `TODO:`

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: I think it is because I didn't add a function like apply_ptac_heating() being called at `add_heating_system` so it just jumped out of the `add_heating_system` after system type checks and didn't create duplicated systems. But it's absolutely a bug because it calculates the sequential loads multiple times, so it will go wrong for multiple systems. Added to the `base-hvac-multiple.xml` and verified this bug.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
yzhou601: I was using the same equation as PTAC for PTHP, but for now these two are equivalent in results since we only model single speed systems. Changed to be the same as MSHP for simplicity.

workflow/sample_files/base-hvac-ptac-electric-resistance.xml
yzhou601: We can only allow single speed for now (default it to single speed)

HPXMLtoOpenStudio/resources/hpxml.rb
yzhou601: I added a test file with furnace pointing to the central AC cooling system. This makes me wonder whether we should link heating and cooling systems in a more consistent way. So for now, PTAC cooling and heating is connected by the cooling system idref, but central ac and furnace are connected by distribution systems. Later we might allow ptac/pthp to be ducted (though from our previous conclusion it is rare, right?), in that case, if we retain the logic of using cooling system idref, will it look confusing if we don't allow furnace+ac to be connected in this way?

Issue #617, opened 2021-02-12
Allow more defaulting


shorowit:
Allow more HPXML inputs to be optional and defaulted.

Some possibilities:
- [x] @bpark1327 HVAC/DHW efficiencies (based on age from HEScore)
- [x] @shorowit Whole-house mechanical ventilation fan power
- [x] @shorowit Wall/roof solar absorptance
- [x] @shorowit Duct areas using duct fractions
- [x] @shorowit Allow orientation instead of azimuth (will need to consider `get_default_azimuths` method during implementation)
- [x] @shorowit Radiant barrier grade
- [x] Whole-house mechanical ventilation flow rate (per ASHRAE 62.2)
- [x] Whole-house fan flow rate and fan power
- [x] Window/skylight U-factor and SHGC based on glazing type, frame type, etc.
- [ ] Air leakage (infiltration) rate
- [ ] Duct leakage

shorowit:
This is good enough for now. Will open separate issues as the need arises.

PR #674, opened 2021-03-20
HVAC/DHW default efficiencies

bpark1327:
## Pull Request Description

Allow age-based HVAC/DHW default efficiencies (taken from HEScore). (#617)

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -1078,7 +1078,8 @@
...
       <sch:assert role='ERROR' test='count(h:TankVolume) &lt;= 1'>Expected 0 or 1 element(s) for xpath: TankVolume</sch:assert>
       <sch:assert role='ERROR' test='count(h:HeatingCapacity) &lt;= 1'>Expected 0 or 1 element(s) for xpath: HeatingCapacity</sch:assert>
-      <sch:assert role='ERROR' test='count(h:UniformEnergyFactor) + count(h:EnergyFactor) = 1'>Expected 1 element(s) for xpath: UniformEnergyFactor | EnergyFactor</sch:assert>
+      <sch:assert role='ERROR' test='count(h:UniformEnergyFactor) + count(h:EnergyFactor) &lt;= 1'>Expected 0 or 1 element(s) for xpath: UniformEnergyFactor | EnergyFactor</sch:assert>
+      <sch:assert role='ERROR' test='count(h:EnergyFactor) + count(h:YearInstalled) &gt;= 1 or count(h:UniformEnergyFactor) = 1'>Expected 1 or more element(s) for xpath: (EnergyFactor | YearInstalled) | UniformEnergyFactor</sch:assert>
```
bpark1327, line 1082: Not sure if this is right. What I want to test here is:
- UniformEnergyFactor and EnergyFactor are optional.
- YearInstalled only required if both UniformEnergyFactor and EnergyFactor are not provided.
- EnergyFactor only required if both UniformEnergyFactor and YearInstalled are not provided. 
- UniformEnergyFactor is only required if both YearInstalled and EnergyFactor are not provided.

shorowit (commented):
Some initial comments.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -426,6 +426,101 @@ def self.apply_skylights(hpxml)
...
+      next unless (heating_system.heating_efficiency_afue.nil? && heating_system.heating_efficiency_percent.nil?)
+
+      year_installed = heating_system.year_installed
+      heating_system_type = heating_system.heating_system_type
+      heating_system_fuel = heating_system.heating_system_fuel
+
+      if [HPXML::HVACTypeFurnace, HPXML::HVACTypeWallFurnace].include? heating_system_type
+        if heating_system_fuel == HPXML::FuelTypeElectricity
+          heating_system.heating_efficiency_afue = 0.98
+        elsif not year_installed.nil?
```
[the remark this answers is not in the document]
bpark1327: I got your point. Yes, if heating_efficiency_afue is nil, year_installed cannot be nil. This will be caught by EPvalidator.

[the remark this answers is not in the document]
bpark1327: Instead, I guess the fuel type needs to be checked here. If a user tries the fuel type not listed in `lookup_hvac_efficiency`, a user would get an error.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -725,6 +820,7 @@ def self.apply_hvac(hpxml, weather)
 
       end
     end
+    # TODO: Default HeatingCapacity17F?
```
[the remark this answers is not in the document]
bpark1327: Hmm, this may sound weird but this is not what I added. I don't know where it came from, but I'll remove it.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -426,6 +426,101 @@ def self.apply_skylights(hpxml)
...
+      end
+      if (heat_pump.fraction_heat_load_served == 0) && heat_pump.heating_efficiency_hspf.nil? && heat_pump.heating_efficiency_cop.nil?
+        if heat_pump_type == HPXML::HVACTypeHeatPumpGroundToAir
+          heat_pump.heating_efficiency_cop = 3.6
+          heat_pump.heating_efficiency_cop_isdefaulted = true
+        else
+          heat_pump.heating_efficiency_hspf = 7.7
+          heat_pump.heating_efficiency_hspf_isdefaulted = true
+        end
+      end
```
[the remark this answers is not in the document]
bpark1327: I wanted to assign an arbitrary value for heating/cooling efficiency if the heat pump has no cooling/heating load served. I guess I can simply assign default heating/cooling efficiency regardless of heating/cooling load served. This will be removed.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -426,6 +426,101 @@ def self.apply_skylights(hpxml)
...
+        if not year_installed.nil?
+          cooling_system.cooling_efficiency_seer = HVAC.lookup_hvac_efficiency(year_installed, cooling_system_type, cooling_system_fuel, 'SEER')
+          cooling_system.cooling_efficiency_seer_isdefaulted = true
+        end
+      elsif cooling_system_type == HPXML::HVACTypeRoomAirConditioner
+        if not year_installed.nil?
+          cooling_system.cooling_efficiency_eer = HVAC.lookup_hvac_efficiency(year_installed, cooling_system_type, cooling_system_fuel, 'EER')
+          cooling_system.cooling_efficiency_eer_isdefaulted = true
+        end
+      end
```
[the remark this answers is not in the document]
bpark1327: MiniSplitAC defaulting will not be allowed because the lookup table doesn't provide default efficiencies for MiniSplitAC.

shorowit:
How should we handle situations when the equipment is newer than 2010?

bpark1327:
> 
> 
> How should we handle situations when the equipment is newer than 2010? Coordinate w/ ResStock & UrbanOpt.

There seem to be two options we could consider: (a) the DOE minimum standard for the year installed or (b) the EnergyStar fraction for the year.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -428,6 +428,97 @@ def self.apply_hvac(hpxml, weather, convert_shared_systems)
...
+        elsif [HPXML::FuelTypeNaturalGas, HPXML::FuelTypePropane, HPXML::FuelTypeOil,
+               HPXML::FuelTypeOil1, HPXML::FuelTypeOil2, HPXML::FuelTypeOil4, HPXML::FuelTypeOil5or6,
+               HPXML::FuelTypeKerosene, HPXML::FuelTypeDiesel].include? heating_system_fuel
+          heating_system.heating_efficiency_percent = 0.80 # with inserts; https://www.volunteerenergy.com/energy-efficient-fireplace/
+        elsif [HPXML::FuelTypeWoodCord, HPXML::FuelTypeWoodPellets].include? heating_system_fuel
+          heating_system.heating_efficiency_percent = 0.70 # https://www.houselogic.com/organize-maintain/home-maintenance-tips/wood-burning-fireplace-inserts-save-energy/
+        elsif [HPXML::FuelTypeCoal, HPXML::FuelTypeCoalAnthracite, HPXML::FuelTypeCoalBituminous, HPXML::FuelTypeCoke].include? heating_system_fuel
+          heating_system.heating_efficiency_percent = 0.75 # assume coal-burning stove efficiency as oil-burning fireplace efficiency
+        end
+        heating_system.heating_efficiency_percent_isdefaulted = true
```
shorowit, line 482: I'm still not comfortable with this. Let's discuss.

bpark1327: I would like to clarify that the shipment-weighted default lookup table provides efficiencies of natural gas wall furnaces. As floor furnaces are similar to wall furnaces, floor furnaces are mapped to wall furnaces. All fuel types are mapped to natural gas for wall/floor furnaces because the lookup table only provides efficiencies for natural gas wall/floor furnaces.

I did a little more research on the stoves and fireplaces. It was found that the difference in efficiency between stoves and fireplaces was not exceptionally large (~3%) for most manufacturers. I propose to use the same default efficiencies for stoves and fireplaces. 
The propane stoves/fireplaces tend to have slightly higher efficiency compared to natural gas stoves/fireplaces, but its degree is so small that it is negligible. As stoves/fireplaces that use fuel other than electricity, wood, wood pellets, propane, natural gas are quite uncommon, all other fuel types are mapped to natural gas or propane.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -930,6 +1002,10 @@ def self.apply_water_heaters(hpxml, nbeds, eri_version)
           water_heating_system.tank_volume = Waterheater.get_default_tank_volume(water_heating_system.fuel_type, nbeds, hpxml.building_construction.number_of_bathrooms)
           water_heating_system.tank_volume_isdefaulted = true
         end
+        if water_heating_system.energy_factor.nil? && water_heating_system.uniform_energy_factor.nil?
+          water_heating_system.energy_factor = Waterheater.get_default_water_heater_efficiency(water_heating_system.year_installed, water_heating_system.fuel_type)
```
shorowit, line 1006: And rename this method too for consistency.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -398,9 +398,10 @@ def test_central_air_conditioners
     hpxml.cooling_systems[0].charge_defect_ratio = -0.11
     hpxml.cooling_systems[0].airflow_defect_ratio = -0.22
     hpxml.cooling_systems[0].cooling_capacity = 12345
+    hpxml.cooling_systems[0].cooling_efficiency_seer = 12.5
```
shorowit, line 401: For robustness, how about also including the year installed here? It will verify that the year installed has no impact on the test result when the efficiency is also provided. (Make the same change to other similar tests.)

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -2329,13 +2519,24 @@ def _test_default_storage_water_heater_values(hpxml, *expected_wh_values)
...
     storage_water_heaters.each_with_index do |wh_system, idx|
-      is_shared, heating_capacity, tank_volume, recovery_efficiency, location = expected_wh_values[idx]
+      is_shared, heating_capacity, tank_volume, recovery_efficiency, location, temperature, energy_factor, first_hour_rating = expected_wh_values[idx]
 
       assert_equal(is_shared, wh_system.is_shared_system)
       assert_in_epsilon(heating_capacity, wh_system.heating_capacity, 0.01)
       assert_equal(tank_volume, wh_system.tank_volume)
       assert_in_epsilon(recovery_efficiency, wh_system.recovery_efficiency, 0.01)
       assert_equal(location, wh_system.location)
+      assert_equal(temperature, wh_system.temperature)
```
shorowit, line 2529: 👀 Nice catch!

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -2329,13 +2519,24 @@ def _test_default_storage_water_heater_values(hpxml, *expected_wh_values)
...
+      if energy_factor.nil?
+        assert_nil(wh_system.energy_factor)
+      else
+        assert_equal(energy_factor, wh_system.energy_factor)
+      end
+      if first_hour_rating.nil?
+        assert_nil(wh_system.first_hour_rating)
+      else
+        assert_equal(first_hour_rating, wh_system.first_hour_rating)
+      end
```
shorowit, line 2539: I don't think this is correct? FHR doesn't seem to ever be provided as an expected value to the `_test_default_storage_water_heater_values` method -- nor do I think it should be since it's never defaulted (and these tests are specific to checking defaulted values)?

docs/source/workflow_inputs.rst
```diff
@@ -682,12 +686,17 @@ Wall/Floor Furnace
...
-  Element                                          Type    Units  Constraints  Required  Default  Notes
-  ===============================================  ======  =====  ===========  ========  =======  ===================
-  ``AnnualHeatingEfficiency[Units="AFUE"]/Value``  double  frac   0 - 1        Yes                Rated efficiency
-  ``extension/FanPowerWatts``                      double  W      >= 0         No        0        Fan power
-  ===============================================  ======  =====  ===========  ========  =======  ===================
+  ===============================================  =======  =====  ===========  ========  ========  =================================
+  Element                                          Type     Units  Constraints  Required  Default   Notes
+  ===============================================  =======  =====  ===========  ========  ========  =================================
+  ``YearInstalled``                                integer                      See [#]_            Year installed
+  ``AnnualHeatingEfficiency[Units="AFUE"]/Value``  double   frac   0 - 1        See [#]_  See [#]_  Rated efficiency
```
shorowit, line 693: I suggest combining these into a single row with an "or" and marking it as required. For example, look at what we did for PV (SystemLossesFraction or YearModulesManufactured). And then you can combine the first too table endnotes into one.

(Make the same change for other HVAC system types.)

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -428,6 +428,78 @@ def self.apply_hvac(hpxml, weather, convert_shared_systems)
...
+        if heating_system_fuel == HPXML::FuelTypeElectricity
+          heating_system.heating_efficiency_afue = 0.98
+        else
+          heating_system.heating_efficiency_afue = HVAC.get_default_hvac_efficiency(year_installed, heating_system_type, heating_system_fuel, HPXML::UnitsAFUE)
+        end
+        heating_system.heating_efficiency_afue_isdefaulted = true
+      elsif [HPXML::HVACTypeElectricResistance].include? heating_system_type
+        next unless heating_system.heating_efficiency_percent.nil?
+
+        heating_system.heating_efficiency_percent = 0.98
```
shorowit, line 449: From my perspective, this is the only question holding up the PR. We have sample files where we assume 1.0 but here we default to 0.98 (per HEScore). We either need to update the defaults to 1.0 instead of 0.98 or we should update our sample files to use 0.98 instead, as we are currently inconsistent.

I'm not sure what to recommend yet, will think about it some more...

shorowit (commented):
This is really close. Great set of tests and I like where the defaulting has ended up. I have a few small suggestions below.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4246,4 +4246,46 @@ def self.get_dehumidifier_default_values(capacity)
 
     return { rh_setpoint: rh_setpoint, ief: ief }
   end
+
+  def self.get_default_hvac_efficiency(year, hvac_type, fuel_type, units)
+    year = 0 if year.nil?
```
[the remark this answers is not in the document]
bpark1327: I agree that it will always have a non-nil value unless a user provides efficiency. I guess we can remove it.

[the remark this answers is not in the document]
shorowit: Just to be clear, the point here is that even if year was nil, assuming zero would be a _very_ poor choice as it would end up choosing the default equipment efficiency based on the first year available in the lookup tables (so, 1970).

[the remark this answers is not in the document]
bpark1327: Ah, good point. I tried to keep the method the way it is. But I should have thought about it critically.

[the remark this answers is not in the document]
shorowit: I see. That made me concerned about the original code, but I looked into the HEScore code and it appears year is only nil when it's an energystar efficiency lookup (so basically the year doesn't matter and is set to zero just to avoid an error). So that's why that is appropriate in the original code but not here.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
shorowit: Okay, I suggest we use 0.98 for furnaces/boilers and 1.0 for electric resistance, stoves, space heaters, etc. Please update our defaults, HPXML sample files, docs, and anything else (?) to be consistent with that.

shorowit (approved):

shorowit:
Thanks @bpark1327!

PR #680, opened 2021-03-25
BuildResidentialHPXML measure: Vacancy updates

joseph-robertson:
## Pull Request Description

Adds:
* `base-schedules-stochastic-vacant.xml`
* `base-schedules-stochastic-vacant.osw`
* `BuildResidentialHPXML/tests/schedules/vacant.csv` (stochastic.csv with vacancy=1)

Removes:
* `extra-vacancy-6-months.osw`

Also, define end uses that are unaffected by vacancy (e.g., refrigerator, etc).

## Checklist

Not all may apply:


BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -87,6 +87,18 @@ def self.col_names
...
+  def self.unaffected_by_vacancy
+    return [
+      'refrigerator',
+      'extra_refrigerator',
+      'freezer',
+      'pool_pump',
+      'pool_heater',
+      'hot_tub_pump',
+      'hot_tub_heater'
+    ]
```
shorowit, line 99: Seems better to return a hash where every end use is accounted for, with true or false as appropriate. And then give an error if there's an end use not covered by the hash. That way this code would be updated if an end use name changes or a new end use is added.

joseph-robertson: Good idea. I will update.

PR #697, opened 2021-04-07
 BuildResidentialHPXML measure: Faster stochastic schedules

shorowit:
## Pull Request Description

Minimal changes that reduce the `ScheduleGenerator.create` method runtime by over 50% (7.6s -> 3.4s). Addresses the largest bottleneck that I saw in the schedule generation related to monthly offsets.

Tested against base-schedules-stochastic.osw, where an identical base-schedules-stochastic_schedules.csv is produced before and after this change.

## Checklist

Not all may apply:


rajeee:
Awesome! @shorowit  Do you think the improvement is coming from the sum vs reduce or because of the other fix involving flattening a list of arrays instead of manipulating the array in a loop? I guess it's the later, but would be good to know if sum is much more inefficient than reduce.

shorowit:
All of the improvement is the latter. I only switched to using the sum() function (new as of ruby 2.4) since it's much clearer what it does.

rajeee (approved):

joseph-robertson (approved):

PR #706, opened 2021-04-10
BuildResidentialHPXML measure: Faster stochastic schedules, second pass

shorowit:
## Pull Request Description

Some more minimal changes that reduce the ScheduleGenerator class runtime further. Now we're at ~80% reduction in speed from the original code:

  | Orig Time (s) | Prev PR Time (s) | New PR Time (s) | Tot Change
-- | -- | -- | -- | --
require yaml/json | 0.22 | 0.22 | 0.01 | -95%
ScheduleGenerator.new | 0.00 | 0.00 | 0.00 | --
ScheduleGenerator.create | 7.11 | 3.44 | 1.06 | -85%
ScheduleGenerator.export | 0.29 | 0.29 | 0.28 | --
**Sum** | **7.62** | **3.95** | **1.35** | **-82%**

Once again tested against base-schedules-stochastic.osw, where an identical base-schedules-stochastic_schedules.csv is produced before and after this change.

## Checklist

Not all may apply:


BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -418,10 +418,7 @@ def create_stochastic_schedules(args:)
           if j >= @mkc_ts_per_day then break end # break as soon as we have filled activities for the day
 
           transition_probs = transition_matrix[(j - 1) * 7...j * 7] # obtain the transition matrix for current timestep
-          transition_probs_matrix = Matrix[*transition_probs]
-          current_state_vec = Matrix.row_vector(state_vector)
-          state_prob = current_state_vec * transition_probs_matrix # Get a new state_probability array
-          state_prob = state_prob.to_a[0]
+          state_prob = transition_probs[active_state]
```
shorowit, line 421: This is obviously so much simpler. Rather than create two matrices and perform matrix multiplication, we simply pull the array of interest out of the 2d array. Reduces time by ~1.5s.

BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -500,15 +497,15 @@ def create_stochastic_schedules(args:)
     #   d. if more cluster, go to 4.
     mins_in_year = 1440 * @total_days_in_year
     mkc_steps_in_a_year = @total_days_in_year * @mkc_ts_per_day
-    sink_activtiy_probable_mins = [0] * mkc_steps_in_a_year # 0 indicates sink activity cannot happen at that time
+    sink_activity_probable_mins = [0] * mkc_steps_in_a_year # 0 indicates sink activity cannot happen at that time
```
shorowit, line 500: Just fixing a typo.

BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -524,12 +521,14 @@ def create_stochastic_schedules(args:)
...
-        todays_probable_steps = sink_activtiy_probable_mins[day * @mkc_ts_per_day...((day + 1) * @mkc_ts_per_day)]
+        todays_probable_steps = sink_activity_probable_mins[day * @mkc_ts_per_day...((day + 1) * @mkc_ts_per_day)]
         todays_probablities = todays_probable_steps.map.with_index { |p, i| p * hourly_onset_prob[i / @mkc_ts_per_hour] }
         prob_sum = todays_probablities.sum(0)
         normalized_probabilities = todays_probablities.map { |p| p * 1 / prob_sum }
         cluster_start_index = weighted_random(prng, normalized_probabilities)
-        sink_activtiy_probable_mins[cluster_start_index] = 0 # mark the 15-min interval as unavailable for another sink event
+        if sink_activity_probable_mins[cluster_start_index] != 0
+          sink_activity_probable_mins[cluster_start_index] = 0 # mark the 15-min interval as unavailable for another sink event
+        end
```
shorowit, line 531: This reduces runtime by another 0.5s.

BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -908,9 +906,9 @@ def apply_monthly_offsets(array:, weekday_monthly_shift_dict:, weekend_monthly_s
         raise "Could not find the entry for month #{month}, day #{day_of_week} and state #{@state}"
       end
 
-      new_array << array[day * 1440, 1440].rotate(lead)
+      new_array.concat(array[day * 1440, 1440].rotate(lead))
     end
-    return new_array.flatten!
+    return new_array
```
shorowit, line 911: Realized we could avoid the flatten call altogether by just building the 1d array from the start. Reduces runtime by ~0.2s.

BuildResidentialHPXML/resources/schedules.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'csv'
-require 'yaml'
+require 'json'
```
shorowit, line 4: This change also shaves 0.2s.

joseph-robertson:
@shorowit Why does ScheduleGenerator.export become faster?

shorowit:
> @shorowit Why does ScheduleGenerator.export become faster?

It doesn't become faster...?

BuildResidentialHPXML/resources/schedules.rb
rajeee: Nice!

BuildResidentialHPXML/resources/schedules.rb
rajeee: Hmm, interesting. I wonder why setting a value at an index would be that expensive that putting a conditional guard would save time?  How are you measuring saved times?

BuildResidentialHPXML/resources/schedules.rb
rajeee: Good one!  
The existing code was a result of trying to translate paper-maths into code without thinking throughly :)

BuildResidentialHPXML/resources/schedules.rb
rajeee: I really like having those comments alongside the values, so not sure if it is worth getting rid of yaml to save 0.2s.

BuildResidentialHPXML/resources/schedules.rb
shorowit: In general, the sink code did take significant time to run, but yes, I was also quite surprised to see that the above line of code was the culprit. But I reproduced it repeatedly. That said, feel free to test it yourself!

I basically used code like:
```
sum_t = 0
@total_days_in_year.times do |day|
      cluster_per_day.times do |cluster_count|
            ...
            t = Time.now
            sink_activity_probable_mins[cluster_start_index] = 0 # mark the 15-min interval as unavailable for another sink event
            sum_t += (Time.now - t)
      end
end
puts sum_t
```

BuildResidentialHPXML/resources/schedules.rb
shorowit: Yes, this is definitely the only controversial change. Indeed the lack of comments in json is unfortunate. But it felt like 0.2s is worth it, especially if A) we are running hundreds of thousands of simulations, or B) we ever want to use this for more use cases (like Home Energy Score) where we are doing everything we can to keep runtimes as fast as possible.

If we really need embedded comments, we could consider CSV or XML (both of which won't introduce overhead like YAML).

joseph-robertson (approved):

BuildResidentialHPXML/resources/schedules.rb
rajeee: Thanks! It's kinda surprising that it works this way, but yeah, can't deny hard results :)

BuildResidentialHPXML/resources/schedules.rb
rajeee: Thanks! Looks like a case of we gotta lose something to gain something :)

shorowit: Also note that I did toy with using embedded comments in the JSON like so:
https://github.com/NREL/OpenStudio-HPXML/blob/d4e34809f996891ef64f52f0f23528c023bce24a/BuildResidentialHPXML/resources/schedules_config.json

Do you think that's better than the current implementation of using a separate file?

rajeee (approved):

BuildResidentialHPXML/resources/schedules.rb
rajeee: I personally find the implementation of comments on JSON even more confusing than having a separate file.

How about adding a script to autogenerate the json from the yaml but keeping the yaml around?
The workflow would be:
1. Make modifications in the yaml.
2. Run the script to generate the JSON.
3. Commit both the json and yaml.

The risk is someone modifying the yaml but forgetting to run the script, thereby silently and mistakenly using the old config while believing they are using new one.

shorowit: Hmm. I can understand the motivation, but the programmer in me 😬s at the thought of redundant (and, as you point out, potentially out of sync) information.

rajeee: Hehe, I understand. It would have been nice if we could just convert it on the fly, but then that defeats the whole purpose and makes it even slower. I guess, it's a unique problem due to having to run this whole thing 1000s of times in succession. If there was a initial setup step in the workflow where we could do this just once before kicking off all the simulations, that would have been great.

PR #744, opened 2021-04-29
Geb HVAC - Single Speed and Two Speed Systems

yzhou601:
## Pull Request Description
Single Speed features:

1.  On-off Deadband thermostat
2. Start-up Loss

Two Speed features:

1. On-off Deadband thermostat
2. Realistic Staging

Alternative of #458 

## Checklist

Not all may apply:

- [x] EPvalidator.xml has been updated
- [x] Tests (and test files) have been updated
- [x] Documentation has been updated
- [x] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [x] No unexpected regression test changes on CI (checked comparison artifacts)

PR #754, opened 2021-05-05
Geb HVAC - Single speed cycling behavior

yzhou601:
## Pull Request Description

This PR:
- [x] Adds optional on-off thermostats/deadbands for single speed system,
- [x] Accounts for the cycling behavior when using realistic thermostats using EMS program that actuates capacity and EIR performance curves.

This is just for single speed system, two speed system thermostats will be handled in #744 .

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3077,6 +3080,50 @@ def self.set_cool_rated_shrs_gross(cooling_system)
     end
   end
 
+  def self.apply_supp_control_for_ddb_thermostat(model, supp_coil, control_zone)
```
yzhou601, line 3083: @jmaguire1 Please take a look at this. I added an EMS control to limit supplemental heating coil operation, so that it is only enabled when zone temperature is out of deadbands.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1773,27 +1777,31 @@ def self.add_residual_ideal_system(runner, model, spaces)
...
 
     living_zone = spaces[HPXML::LocationLivingSpace].thermalZone.get
     obj_name = Constants.ObjectNameIdealAirSystemResidual
 
-    if @remaining_cool_load_frac < 1.0
-      sequential_cool_load_frac = 1.0
-    else
-      sequential_cool_load_frac = 0.0 # no cooling system, don't add ideal air for cooling either
-    end
+    # only enable residual ideal system when hvac's not on on/off thermostat control
```
yzhou601, line 1785: If the on off thermostat feature is enabled, it is not compatible with ideal system approach. 
1. For residual ideal system, it will perfectly resolve all remaining loads at each timestep, resulting in very short cycle (1 timestep per cycle regardless of hvac system capacity), so the residual ideal system should be disabled. 
2. For ideal systems that created for partial hvac, I verified that the on-off thermostat will also work oddly (short cycle). Taking an example of hvac system with on-off thermostat serving 33% of loads, for the current time step, ideal system first met 67% loads, and the 33% loads are assigned to hvac. If this 33% is not met perfectly, then the loads will be carried over to the next time step, and being resolved mostly by ideal system (67%). So ideal system actually helps resolve the current loads assigned to actual hvac system in the next time step.

Considering this situation, I think our feature is more reasonable to only be enabled in fully conditioned houses. @jmaguire1 @shorowit

jmaguire1: @yzhou601: I agree, I think we should only be using this feature for homes where 100% of the load is met by a single HVAC system. I think we at least want to throw a warning (probably an error) in this scenario, with a message letting the user know why we didn't allow this and what they need to change (either remove the deadband thermostat or fully condition the space with a single HVAC system)

yzhou601 (commented):
@shorowit This PR is looking good in local testing. Temperature Capacity Multiplier is manually changed to 6 in local testing. This PR didn't implement the smart recovery discussed in https://github.com/NREL/OpenStudio-HPXML/issues/772. 
image
image
image

The backup coil behavior is never turned on with current heat pump coil capacity, so I manually cut the coil capacity, and the backup coil will be turned on when heat pump cannot resolve all the loads (temperature falls below cut-in temperature). Looks good.

yzhou601:
Question: How are we going to handle capacity multiplier? Are we going to add [Temperature Capacity Multiplier](https://bigladdersoftware.com/epx/docs/9-6/input-output-reference/group-simulation-parameters.html#field-temperature-capacity-multiplier) as an exposed input? Or default it to 6.0 (or any better number being researched) whenever ddb thermostat is enabled? @jmaguire1 @shorowit

shorowit:
Where did this question come from?

My (limited) understanding is that because we're explicitly including internal mass objects (for both furniture and partition walls) in our models, we shouldn't adjust the temperature capacity multiplier too. The temperature capacity multiplier is a much blunter instrument for making adjustments due to mass.

yzhou601:
> Where did this question come from?
> 
> My (limited) understanding is that because we're explicitly including internal mass objects (for both furniture and partition walls) in our models, we shouldn't adjust the temperature capacity multiplier too. The temperature capacity multiplier is a much blunter instrument for making adjustments due to mass.

We're running models with temperature capacity multiplier manually set to 6.0 for testing, and I thought we would like to include that into our PR. @jmaguire1 How do you think about the internal mass objects? Are we double-counting its impact in our test runs?

shorowit:
@yzhou601 Sorry, the CI errors here are my fault. Please ignore, will be fixed soon.

shorowit:
@yzhou601 If you pull in master again, that'll fix the CI.

yzhou601 (commented):
This PR is ready for an initial review for the modeling approach. It only configures one test file for now due to the long simulation time for a minute time step simulation run, and no unit test added yet. Please suggest for better test suite.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -882,10 +886,18 @@ def self.get_heating_setpoints(hvac_control, year)
     if hvac_control.weekday_heating_setpoints.nil? || hvac_control.weekend_heating_setpoints.nil?
       # Base heating setpoint
       htg_setpoint = hvac_control.heating_setpoint_temp
+      # Apply thermostat offset due to onoff control
```
shorowit, line 889: Seems like this handles only one of a few different setpoint input types? For example, it doesn't address the "# 24-hr weekday/weekend heating setpoint schedules" code below. Probably better to update the setpoints after the `if`/`else` block of code rather than inserting separate logic into the `if` and `else` statements. Like this?

```
htg_weekday_setpoints = htg_weekday_setpoints.map { |i| i.map { |j| j - offset_db / 2.0 } }
htg_weekend_setpoints = htg_weekend_setpoints.map { |i| i.map { |j| j - offset_db / 2.0 } }
```

We also allow 8760 setpoints to be entered in a CSV, which doesn't seem to be handled.

The fact that we're changing setpoints makes me a bit nervous that other code may need to be updated. For example, there might be other models that use the original setpoints (i.e., refer to the `hvac_control` inputs) rather than these modified inputs, and thus don't operate correctly. Perhaps everything fine, but I could imagine problems like this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1567,6 +1587,9 @@ def self.create_supp_heating_coil(model, obj_name, heat_pump)
     end
     htg_supp_coil.setNominalCapacity(UnitConversions.convert(capacity, 'Btu/hr', 'W'))
     htg_supp_coil.setName(obj_name + ' ' + Constants.ObjectNameBackupHeatingCoil)
+    if is_ddb_control and (not htg_coil.nil?)
+      apply_supp_control_for_ddb_thermostat(model, htg_supp_coil, control_zone, htg_coil)
+    end
```
shorowit, line 1592: Is it even possible to have a backup heating coil but no heating coil? (What would it be backup relative to?) It seems to me that `not htg_coil.nil?` is not needed.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2969,6 +3012,183 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+    supp_coil_avail_program.addLine('  If living_t > htg_sp_h')
+    supp_coil_avail_program.addLine("    Set #{supp_coil_avail_actuator.name} = 0")
+    supp_coil_avail_program.addLine('  Else')
+    supp_coil_avail_program.addLine("    Set #{supp_coil_avail_actuator.name} = 1")
+    supp_coil_avail_program.addLine('  EndIf')
+    supp_coil_avail_program.addLine('Else') # Only turn on the backup coil when temprature is below lower end of ddb.
+    r_s_a = ["#{htg_energy_trend.name} > 0"]
+    (1...5).each do |t_i|
+      r_s_a << "(@TrendValue #{htg_energy_trend.name} #{t_i}) > 0"
+    end
```
shorowit, line 3068: What is this doing? Why 5?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2969,6 +3012,183 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+      cc_out_calc << c_name + " * (#{cap_curve_var_in.name}^#{i})"
+    end
+    eir_fflow_spec.each_with_index do |coeff, i|
+      c_name = "c_#{i + 1}_eir"
+      cycling_degrad_program.addLine("Set #{c_name} = #{coeff}")
+      ec_out_calc << c_name + " * (#{eir_curve_var_in.name}^#{i})"
+    end
+    cycling_degrad_program.addLine("Set cc_out = #{cc_out_calc.join(' + ')}")
+    cycling_degrad_program.addLine("Set ec_out = #{ec_out_calc.join(' + ')}")
+    (0...number_of_timestep_logged + 1).each do |t_i|
```
shorowit, line 3153: You keep using `(0...foo+1)`, isn't it much clearer to use `(0..foo)`? I recommend avoiding use of `...` in general.

```
irb(main):002:0> (0...3+1).to_a
=> [0, 1, 2, 3]
irb(main):003:0> (0..3).to_a
=> [0, 1, 2, 3]
```

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1188,6 +1190,15 @@
     </sch:rule>
   </sch:pattern>
 
+  <sch:pattern>
+    <sch:title>[HVACControl]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACControl'>
+      <sch:assert role='ERROR' test='count(h:SystemIdentifier) = 1'>Expected 1 element(s) for xpath: SystemIdentifier</sch:assert> <!-- See [HVACControl=Heating] and/or [HVACControl=Cooling] -->
+      <sch:assert role='ERROR' test='count(h:extension/h:OnOffThermostatDeadband) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/OnOffThermostatDeadband</sch:assert> <!-- deg F, specify a positive value of deadband between setpoint and cutout temperature to model on/off thermostat behavior. The same value applies to both heating and cooling. -->
```
shorowit, line 1197: Seems like this should have `Temperature` in the name. As written, I would have though it's a boolean value.

docs/source/workflow_inputs.rst
```diff
@@ -1731,11 +1731,13 @@ If any HVAC systems are specified, a single thermostat is entered as a ``/HPXML/
   ``HeatingSeason``                                        element                       No        See [#]_   Heating season        
   ``CoolingSeason``                                        element                       No        See [#]_   Cooling season
   ``extension/CeilingFanSetpointTempCoolingSeasonOffset``  double    F      >= 0         No        0          Cooling setpoint temperature offset [#]_
+  ``extension/OnOffThermostatDeadband``                    double    F      >= 0         No        0          On-off thermostats/deadbands [#]_
```
shorowit, line 1734:
```suggestion
  ``extension/OnOffThermostatDeadband``                    double    F      >= 0         No        0          On-off thermostat deadband temperature [#]_
```

tasks.rb
```diff
@@ -3823,6 +3824,14 @@ def apply_hpxml_modification(hpxml_file, hpxml)
     hpxml.hvac_controls[0].cooling_setup_temp = 80
     hpxml.hvac_controls[0].cooling_setup_hours_per_week = 6 * 7
     hpxml.hvac_controls[0].cooling_setup_start_hour = 9 # 9am
+  elsif ['base-hvac-onoff-thermostat-deadband.xml'].include? hpxml_file
+    hpxml.header.timestep = 1
+    hpxml.header.temperature_capacitance_multiplier = 4.0
```
shorowit, line 3829: I don't recall seeing any warnings/errors related to the temperature capacitance multiplier. Should there be?

tasks.rb
```diff
@@ -3823,6 +3824,14 @@ def apply_hpxml_modification(hpxml_file, hpxml)
...
+    hpxml.hvac_controls[0].onoff_thermostat_deadband = 2.0
+    hpxml.heat_pumps[0].heating_capacity = 10000
+    hpxml.heat_pumps[0].heating_capacity_17F = 5900.0
+  elsif ['base-atticroof-cathedral.xml'].include? hpxml_file
+    hpxml.hvac_distributions[0].conditioned_floor_area_served = 2700
```
shorowit, line 3834: What's this change for?

workflow/tests/hpxml_translator_test.rb
```diff
@@ -1358,15 +1358,17 @@ def skip_utility_bill_warning(err_line)
...
-      if hpxml.total_fraction_heat_load_served == 0
-        assert_equal(0, unmet_hours_htg)
-      else
-        assert_operator(unmet_hours_htg, :<, 350)
-      end
-      if hpxml.total_fraction_cool_load_served == 0
-        assert_equal(0, unmet_hours_clg)
-      else
-        assert_operator(unmet_hours_clg, :<, 350)
+      if not hpxml_path.include? 'base-hvac-onoff-thermostat-deadband.xml'
```
shorowit, line 1361: Can we update the unmet hours EMS program to properly account for the deadband temperature?

docs/source/workflow_inputs.rst
```diff
@@ -1731,11 +1731,13 @@ If any HVAC systems are specified, a single thermostat is entered as a ``/HPXML/
...
   =======================================================  ========  =====  ===========  ========  =========  ========================================
 
   .. [#] If HeatingSeason not provided, defaults to year-round.
   .. [#] If CoolingSeason not provided, defaults to year-round.
   .. [#] CeilingFanSetpointTempCoolingSeasonOffset should only be used if there are sufficient ceiling fans present to warrant a reduced cooling setpoint.
+  .. [#] OnOffThermostatDeadband should only be used if 100% heating or cooling is provided (FractionHeatLoadServed or FractionCoolLoadServed sum to 1.0). OnOffThermostatDeadband is only enabled when simulation time step is 1 minute, otherwise simulation will ignore this feature and throw a warning.
```
shorowit, line 1740: I feel like we need a bit more description about when/why a user should use this feature. There are presumably a lot of use cases where a user should _not_ use this. For example, someone should only use this if they care about timeseries outputs and ....... ?

tasks.rb
```diff
@@ -391,6 +392,7 @@ def create_hpxmls
     'base-hvac-none.xml' => 'base.xml',
     'base-hvac-portable-heater-gas-only.xml' => 'base.xml',
     'base-hvac-programmable-thermostat.xml' => 'base.xml',
+    'base-hvac-onoff-thermostat-deadband.xml' => 'base-hvac-air-to-air-heat-pump-1-speed.xml',
```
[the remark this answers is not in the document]
shorowit: This will be a challenge indeed. Not immediately sure what to suggest. At a minimum, we need to manually run simulations for different system types, thermostat setpoint input types, etc. if that hasn't already been done.

shorowit (commented):
I haven't tried to understand the EMS programs or verify the outputs (I'm relying on the fact that you have tested it and are happy with the outputs). So my review comments are mostly about software code changes.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -4092,6 +4101,7 @@ def to_oga(doc)
       XMLHelper.add_extension(hvac_control, 'WeekendSetpointTempsHeatingSeason', @weekend_heating_setpoints, :string) unless @weekend_heating_setpoints.nil?
       XMLHelper.add_extension(hvac_control, 'WeekdaySetpointTempsCoolingSeason', @weekday_cooling_setpoints, :string) unless @weekday_cooling_setpoints.nil?
       XMLHelper.add_extension(hvac_control, 'WeekendSetpointTempsCoolingSeason', @weekend_cooling_setpoints, :string) unless @weekend_cooling_setpoints.nil?
+      XMLHelper.add_extension(hvac_control, 'OnOffThermostatDeadband', @onoff_thermostat_deadband, :float) unless @onoff_thermostat_deadband.nil?
```
shorowit, line 4104: Should we default this value so it can be seen/verified by a user (in the `in.xml`)?

E.g.:
```suggestion
      XMLHelper.add_extension(hvac_control, 'OnOffThermostatDeadband', @onoff_thermostat_deadband, :float, @onoff_thermostat_deadband_isdefaulted) unless @onoff_thermostat_deadband.nil?
```

And then update hpxml_defaults.rb and test_defaults.rb.

This would also make it so that you don't have to check for nil anywhere in the code after it has been defaulted.

Changelog.md
```diff
@@ -25,6 +25,7 @@ __New Features__
 - ReportUtilityBills measure:
   - Removes utility rate and PV related arguments in lieu of new utility bill scenarios described inside the HPXML file.
 - Allows shared dishwasher/clothes washer to be attached to a hot water distribution system instead of a single water heater.
+- Allows modeling on/off thermostat deadband with start-up degredation.
```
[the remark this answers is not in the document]
yzhou601: It does includes all the DX coil based systems. @jmaguire1 Correct me if I'm wrong. I used to have a test file for central AC systems. I will call it out in the docs after checking with @jmaguire1 .

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: This is assuming that after 5 mins of unmet loads, the back up coils are turned on. There's a similar logic for realistic speed control program for 2 speed coils too. That is keep the lower speed coils operating for a few minutes (5 mins), if it's not meeting the loads, then the higher speed is kicked on. 5 is an assumed value, and @jmaguire1 and I discussed that it is probably better to use the assumption instead of exposing the inputs, regarding to the fact that these kind of information are pretty hard to get from users.

tasks.rb
yzhou601: I don't think I made this change, must be something from resolving conflicts. Will remove on next commit.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: Good point. So it's probably fine when other code references the hvac_control setpoints. Maybe the concern is when other parts of the code reference the E+ setpoint directly then. For example, the natural ventilation and whole house fan EMS program uses sensors on the thermostat setpoints.

I can't help but feel that E+ should treat the entered setpoints as the _midpoint_ of the deadband, so that a user doesn't have to modify their entered setpoints when providing this deadband temperature. What do you think?

Regarding the 8760 setpoint schedule, I recall that we are already having to modify it (to convert from F to C), so perhaps you can modify the code [here](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/schedules.rb#L1431).

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: I see. Please add a comment to the code that explains it. (In general, you did a good job at explaining hard-coded numbers, but this one has no explanation.)

tasks.rb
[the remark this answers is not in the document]
shorowit: Let's just go with two test files -- one for simple thermostat setpoints (which you already have) and one with 8760 scheduled setpoints.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: "Regarding the 8760 setpoint schedule, I recall that we are already having to modify it (to convert from F to C), so perhaps you can modify the code [here](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/schedules.rb#L1431)." I am not sure if I followed, I modified so that these two lines applied to both cases (after if-else work):

    htg_weekday_setpoints = htg_weekday_setpoints.map { |i| i.map { |j| j - offset_db / 2.0 } }
    htg_weekend_setpoints = htg_weekend_setpoints.map { |i| i.map { |j| j - offset_db / 2.0 } }

Or is there anything else we need to change?

workflow/tests/hpxml_translator_test.rb
yzhou601: My concern is that the E+ approach of modifying setpoint during the on/off thermostat operations may make unmet loads challenging to make sense, eg. when hvac is off, the thermostat setpoint is set to be the lower end of deadband and if the hvac is on, the thermostat setpoint is set to be the higher end of deadband. It could make every cycle an unmet loads. This is from what I realized that what happened in E+, need to double check the `Zone Heating/Cooling Setpoint Not Met Time` outputs to confirm.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: The if/else does not apply when using 8760 setpoint schedules, so I assume additional changes are still needed, yes. If you run the `base-schedules-detailed-setpoints.xml`, for example, it should be obvious if additional changes are needed.

workflow/tests/hpxml_translator_test.rb
shorowit: Right. The idea is that we can zero out unmet loads in our [EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/4c519231ee9604d0baaa8ee0a840b190559b8ee8/HPXMLtoOpenStudio/measure.rb#L1917-L1941) for times when the indoor temperature is actually meeting the nominal (i.e., not adjusted for deadband) setpoint, so that it won't be confusing to the user.

shorowit: (In the same way that we currently zero out unmet loads when they occur outside the heating/cooling season.)

docs/source/workflow_inputs.rst
jmaguire1: I'll add that description, but it's mostly when looking at advanced controls or building to grid cosim.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: Got it. Yes the 8760 setpoint schedules are not handled here. I looked at the convert_setpoints method at `schedules.rb`, there's some challenge reading in the hpxml onoff thermostat temperature inputs into that method. It looks like a preprocessing on schedule files we get from users (and storing the information in the schedule class if I understood it correctly) and it looks pretty separate from other measures. Do you think we should introduce a hpxml argument of on/off thermostat temperature into the schedule handler? Or is it better to modify it later in the hvac.rb or HPXMLtoOpenStudio/measure.rb? I personally would think the latter one might be better since we read in and handle the inputs in the same place.

yzhou601: @shorowit I took a second look at the preheating/precooling EMS programs and the natural ventilation program. 
1. The preheating/precooling EMS program I felt like it mimics the real HVAC systems, so it's fine to use the deadband setpoint to calculate loads (so instead of meeting a constant setpoint, it will operate based on the on/off thermostat setpoints, starting to report preheating when zone temperature meets lower bound and stopping preheating when zone temperature meets higher bound). 
2. And the natural ventilation program I might need some extra suggestions. So currently we're shifting heating down by half deadband degree F, and shifting cooling up by half deadband degree F. So that means if there's no heating/cooling operations, the average of these two(which is used in natural ventilation program), is not changed, eg. if the heating setpoint is 70F, cooling setpoint is 76F, then the average will be 73F, with deadband, heating setpoint is 69F. cooling setpoint is 77F, average is still 73F. But if there's heating/cooling operation, the setpoint schedule values will be adjusted by deadband (going up or down by the thermostat deadband, eg. if heating setpoint is 71F then the schedule value will be at 70F or 72F based on HVAC), it will make the average off by half of the deadband. Is this still a fair assumption? If not, adjusting it back will need some extra EMS efforts to detect zone temperature and see which value the current schedule is at.

yzhou601:
@shorowit I looked at the detailed E+ output results, the degradation curve is working correctly, the unmet hours are counting correctly, the backup coil is operating correctly. This PR is ready for review again.

Changelog.md
```diff
@@ -35,6 +35,7 @@ __New Features__
 - ReportUtilityBills measure:
   - Removes utility rate and PV related arguments in lieu of new utility bill scenarios described inside the HPXML file.
   - Adds detailed calculations for tiered, time-of-use, or real-time pricing electric rates using OpenEI tariff files.
+- Allows modeling on/off thermostat deadband with start-up degredation.
```
shorowit, line 38: Still need to call out that this applies only to DX systems -- both in the changelog and the documentation.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1625,6 +1636,13 @@ def self.add_setpoints(runner, model, weather, spaces)
     hvac_control = @hpxml.hvac_controls[0]
     living_zone = spaces[HPXML::LocationLivingSpace].thermalZone.get
     has_ceiling_fan = (@hpxml.ceiling_fans.size > 0)
+    timestep = @hpxml.header.timestep
+    # Do not apply on off thermostat if timestep is >= 2
+    # Only availabe with 1 min time step
+    if timestep >= 2 && hvac_control.onoff_thermostat_deadband > 0.0
+      hvac_control.onoff_thermostat_deadband = 0.0
+      runner.registerWarning('Time step too large to enable on-off thermostat deadband. Continute without on-off control.')
```
shorowit, line 1644: It looks like you've added this same warning to EPvalidator.xml, so it can be removed from here?

shorowit: Also, how about moving this code to hpxml_defaults.rb so that the deadband gets reset to zero as early in the measure as possible? It would also then indicate in the `in.xml` whether we have overridden the value or not.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1186,6 +1188,16 @@
...
+      <sch:assert role='ERROR' test='number(h:extension/h:OnOffThermostatDeadbandTemperature) &gt;= 0 or not(h:extension/h:OnOffThermostatDeadbandTemperature)'>Expected extension/OnOffThermostatDeadbandTemperature to be greater than or equal to 0</sch:assert>
+      <sch:report role='WARN' test='number(../../../../../h:SoftwareInfo/h:extension/h:SimulationControl/h:Timestep) &gt; 1 and (h:extension/h:OnOffThermostatDeadbandTemperature)'>Timestep should be 1 when extension/OnOffThermostatDeadbandTemperature exists</sch:report>
+      <sch:report role='WARN' test='number(../../../../../h:SoftwareInfo/h:extension/h:SimulationControl/h:TemperatureCapacitanceMultiplier) &lt;= 1 and (h:extension/h:OnOffThermostatDeadbandTemperature)'>TemperatureCapacitanceMultiplier should typically be greater than 1 when extension/OnOffThermostatDeadbandTemperature exists</sch:report>
```
shorowit, line 1197: These conditional checks here (and the two above) seem far more complicated than they need to be. And these aren't checking if the deadband temperature is non-zero, only if the element exists.

I suggest moving all checks based on there being a non-zero deadband temperature (these and the two earlier) into a new conditional block:

```
<sch:pattern>
  <sch:title>[HVACControl=DeadbandControl]</sch:title>
  <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACControl[number(h:extension/h:OnOffThermostatDeadbandTemperature) &gt; 0]'>
    <sch:assert role='ERROR' test='sum(../h:HVACPlant/*/h:FractionHeatLoadServed) &gt;= 0.99 or count(../h:HVACPlant/*/h:FractionHeatLoadServed) = 0>Expected sum(FractionHeatLoadServed) to be equal to 1</sch:assert>
    <sch:assert role='ERROR' test='sum(.../h:HVACPlant/*/h:FractionCoolLoadServed) &gt;= 0.99 or count(.../h:HVACPlant/*/h:FractionCoolLoadServed) = 0>Expected sum(FractionCoolLoadServed) to be equal to 1</sch:assert>
    <sch:report role='WARN' test='number(../../../../../h:SoftwareInfo/h:extension/h:SimulationControl/h:Timestep) &gt; 1'>Timestep should be 1; simulation will continue without deadband control.</sch:report>
    <sch:report role='WARN' test='number(../../../../../h:SoftwareInfo/h:extension/h:SimulationControl/h:TemperatureCapacitanceMultiplier) &lt;= 1'>TemperatureCapacitanceMultiplier should typically be greater than 1.</sch:report>
  </sch:rule>
</sch:pattern>
```

Something like that. It's untested. But the asserts are simpler and it should be easier for a user to understand.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1397,18 +1397,24 @@ def calc_peak_flow_from_daily_gpm(col_name:,
...
-    if File.exist? tmp_schedules_path
-      @external_file = OpenStudio::Model::ExternalFile::getExternalFile(@model, tmp_schedules_path)
+  def get_external_file
+    if File.exist? @tmp_schedules_path
+      if @external_file.nil?
+        @external_file = OpenStudio::Model::ExternalFile::getExternalFile(@model, @tmp_schedules_path)
+      else
+        # Re-read file contents
+        @external_file.remove
+        @external_file = OpenStudio::Model::ExternalFile::getExternalFile(@model, @tmp_schedules_path)
```
shorowit, line 1407: If I'm understanding correctly, you are creating the csv file, reading it back in, deleting it, and then recreating it? 🤢 Why is it so complicated? Is it not possible to just add the deadband offset logic into the existing `convert_setpoints` method?

HPXMLtoOpenStudio/resources/unit_conversions.rb
```diff
@@ -51,6 +51,7 @@ class UnitConversions
 
     # Temperature
     ['k', 'r'] => 1.8,
+    ['deltac', 'deltaf'] => 1.8,
```
shorowit, line 54: 👍 

Can we remove the line above and update any code that still references 'k' <-> 'c' conversions? They are all almost certainly doing deltaC <-> deltaF conversions, it will be less confusing to developers. I am still seeing references in hvac.rb, waterheater.rb, and weather.rb.

docs/source/workflow_inputs.rst
```diff
@@ -1736,18 +1736,20 @@ HPXML HVAC Control
...
-  =======================================================  ========  =====  ===========  ========  =========  ========================================
+  ``extension/OnOffThermostatDeadbandTemperature``         double    F      >= 0         No        0          On-off thermostats/deadbands temperature[#]_
+  =======================================================  ========  =====  ===========  ========  =========  ============================================
 
   .. [#] If HeatingSeason not provided, defaults to year-round.
   .. [#] If CoolingSeason not provided, defaults to year-round.
   .. [#] CeilingFanSetpointTempCoolingSeasonOffset should only be used if there are sufficient ceiling fans present to warrant a reduced cooling setpoint.
+  .. [#] OnOffThermostatDeadband should only be used if 100% heating or cooling is provided (FractionHeatLoadServed or FractionCoolLoadServed sum to 1.0). OnOffThermostatDeadband is only enabled when simulation time step is 1 minute, otherwise simulation will ignore this feature and throw a warning.
```
shorowit, line 1752: We still need some updated guidance here on when a user should use this. And we should probably call it a research or advanced feature.

shorowit: We should also define deadband here and make it clear that this deadband is _not_ the difference between the heating and cooling setpoints. There is a lot of confusion around the term deadband, not everyone uses the same definition.

One such example of everyone using different definitions: https://hvac-talk.com/vbb/threads/2198021-thermostat-deadband-vs-temperature-differential

shorowit (commented):
Nice job addressing my comments. This is getting there. There are still a few things to clean up and address.

HPXMLtoOpenStudio/resources/schedules.rb
yzhou601: This is because the convert_setpoints is called at the creation of schedulefile instance. I thought the deadband temperature inputs are hard to pass at that point, so I try to modify the created schedule file when we’re handling setpoints in the measure. Maybe it’s not true. I will give it a second thought and try to simplify the logic.

shorowit: It's hard for me to imagine why it might be hard to do it the more straightforward way, but it's possible. Let @joseph-robertson or me know if you need help here.

shorowit: @yzhou601 Looks like you figured it out! Much better. 🎉

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1420,34 +1420,31 @@ def set_vacancy
...
-      end
-      if offset_db.to_f > 0.0
-        SchedulesFile.HVACSetpointColumnNames.each do |hvac_setpoint_col_name|
-          next unless col_names.include?(hvac_setpoint_col_name)
-          if hvac_setpoint_col_name == ColumnHeatingSetpoint
-            @tmp_schedules[hvac_setpoint_col_name][i] = @tmp_schedules[hvac_setpoint_col_name][i] - UnitConversions.convert(offset_db / 2.0, 'deltaF', 'deltaC')
-          elsif hvac_setpoint_col_name == ColumnCoolingSetpoint
-            @tmp_schedules[hvac_setpoint_col_name][i] = @tmp_schedules[hvac_setpoint_col_name][i] + UnitConversions.convert(offset_db / 2.0, 'deltaF', 'deltaC')
-          end
+        if setpoint_col_name == ColumnHeatingSetpoint
```
yzhou601, line 1444: 👍

yzhou601:
Remaining work here:
1. Conclusion from discussion with Jon: We'll only allow this feature to be applied to single speed central ac and ashp for now. Will throw a warning and reset the onoffthermostattemperature to 0.0 for other systems @yzhou601 .  In the future we might want to explore a methodology to find the relationship between c_d and the time to reach steady state, and update other systems' start-up capacity degradation approach based on this.
2. Jeff to update docs. @jmaguire1

yzhou601:
@shorowit @jmaguire1 So I have been addressing the use case to be:
1. Only apply the On/off thermostat + degradation curve + supplemental coil control logic to central ac and ASHP, otherwise throw an error from EPvalidator
2. Only allow one cooling system to be specified and enable this feature, to filter out some fringe cases.
3. Only apply the  On/off thermostat + degradation curve + supplemental coil control logic to single speed system, otherwise throw an warning and ignore the inputs from hpxml_defaults
4. If central AC has a heating system, the on/off thermostat will apply to both heating and cooling

yzhou601:
@shorowit  This pr is ready for review

yzhou601:
@shorowit Do you have any idea what may cause the CI failure? It doesn't seem to be anything related to this PR.
image

shorowit:
@yzhou601 I told the CI to re-run that test and it seems to be working. Must have been a weird hiccup.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -168,12 +168,15 @@ def self.create(hpxml, runner, model, hpxml_path, epw_path, weather, output_dir,
 
     # Init
     check_file_references(hpxml_path)
+    # Fixme: do we need to check system number of speed?
```
yzhou601, line 171: the schedule modification happens at the top of the measure, and number of speeds is an optional user input so we don't guarantee to have that info at this point, that's why I put a fixme here.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1895,7 +1916,13 @@ def self.add_unmet_hours_output(model, spaces)
       end
       line += " && (#{@hvac_availability_sensor.name} == 1)" if not @hvac_availability_sensor.nil?
       program.addLine(line)
-      program.addLine("  Set #{htg_hrs} = #{htg_hrs} + #{htg_sensor.name}")
+      if is_ddb_control
+        program.addLine("  If #{zone_air_temp_sensor.name} < (#{sensor_htg_spt.name} - #{UnitConversions.convert(hvac_control.onoff_thermostat_deadband, 'deltaF', 'deltaC')})")
```
yzhou601, line 1920: Modify the unmet hour EMS to exclude time steps when temperature is within deadband.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1537,6 +1537,28 @@ def self.apply_hvac_control(hpxml, schedules_file)
         hvac_control.cooling_setup_start_hour_isdefaulted = true
       end
 
+      if hvac_control.onoff_thermostat_deadband.nil?
+        hvac_control.onoff_thermostat_deadband = 0.0
+        hvac_control.onoff_thermostat_deadband_isdefaulted = true
+      end
+
+      # Do not apply on off thermostat if timestep is >= 2
+      # Only availabe with 1 min time step
```
yzhou601, line 1546: This is because our EMS program to ramp up capacity happens within 2 mins, if the time step is over 1 min, then the impact is not captured.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1537,6 +1537,28 @@ def self.apply_hvac_control(hpxml, schedules_file)
...
+      if hvac_control.onoff_thermostat_deadband > 0.0 && hpxml.header.timestep >= 2
+        hvac_control.onoff_thermostat_deadband = 0.0
+        hvac_control.onoff_thermostat_deadband_isdefaulted = true
+      end
+
+      # Only apply for single speed systems for now
```
yzhou601, line 1552: This will change in the two speed PR.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -865,6 +865,7 @@ def self.apply_setpoints(model, runner, weather, hvac_control, living_zone, has_
     thermostat_setpoint.setName("#{living_zone.name} temperature setpoint")
     thermostat_setpoint.setHeatingSetpointTemperatureSchedule(heating_sch)
     thermostat_setpoint.setCoolingSetpointTemperatureSchedule(cooling_sch)
+    thermostat_setpoint.setTemperatureDifferenceBetweenCutoutAndSetpoint(UnitConversions.convert(hvac_control.onoff_thermostat_deadband, 'deltaF', 'deltaC'))
```
yzhou601, line 868: Add on-off thermostat here.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -939,6 +940,10 @@ def self.get_heating_setpoints(hvac_control, year)
       htg_weekend_setpoints = hvac_control.weekend_heating_setpoints.split(',').map { |i| Float(i) }
       htg_weekend_setpoints = [htg_weekend_setpoints] * num_days
     end
+    # Apply thermostat offset due to onoff control
+    offset_db = hvac_control.onoff_thermostat_deadband
+    htg_weekday_setpoints = htg_weekday_setpoints.map { |i| i.map { |j| j - offset_db / 2.0 } }
```
yzhou601, line 945: Modify setpoint due to the nature of E+ on-off thermostat taking setpoint as cut in temperature.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1646,6 +1655,9 @@ def self.create_supp_heating_coil(model, obj_name, heat_pump)
     end
     htg_supp_coil.setNominalCapacity(UnitConversions.convert(capacity, 'Btu/hr', 'W'))
     htg_supp_coil.setName(obj_name + ' ' + Constants.ObjectNameBackupHeatingCoil)
+    if is_ddb_control
+      apply_supp_control_for_ddb_thermostat(model, htg_supp_coil, control_zone, htg_coil)
```
yzhou601, line 1659: This EMS is to disable supp coil when temperature is within deadband.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3019,6 +3051,194 @@ def self.set_cool_rated_shrs_gross(runner, cooling_system)
...
+
+  def self.calc_time_to_full_cap(c_d)
+    # assuming a linear relationship between points we have data for: 2 minutes at 0.08 and 5 minutes at 0.23
+    time = (20.0 * c_d + 0.4).round
+    time = [time, 2].max
+    time = [time, 5].min
+    return time
+  end
+
+  def self.apply_capacity_degradation_EMS(model, system_ap, coil_name, is_cooling, cap_fff_curve, eir_fff_curve)
```
yzhou601, line 3136: EMS program for capacity ramp up for each cycle.

yzhou601:
Will be combined with #744

PR #774, opened 2021-06-04
GSHP installation quality, take 2

yzhou601:
## Pull Request Description

Try to fix an issue where quadlinear/quintlinear curve is not reaching 1 for rated airflow in EMS program.

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/hvac.rb
+++ b/HPXMLtoOpenStudio/resources/hvac.rb
@@ -3816,13 +3816,17 @@ class HVAC
           fault_program.addLine("Set a1_AF_EIR_c = #{cool_eir_fff_curve.coefficient1Constant}")
           fault_program.addLine("Set a2_AF_EIR_c = #{cool_eir_fff_curve.coefficient2x}")
           fault_program.addLine("Set a3_AF_EIR_c = #{cool_eir_fff_curve.coefficient3xPOW2}")
+          fault_program.addLine("Set AF_Qgr_residual = 0")
+          fault_program.addLine("Set AF_EIR_residual = 0")
         else
-          fault_program.addLine("Set a1_AF_Qgr_c = #{cool_cap_fff_curve.coefficient1Constant} + (#{cool_cap_fff_curve.coefficient2w}*#{var1_sensor.name}) + (#{cool_cap_fff_curve.coefficient3x}*#{var2_sensor.name}) + (#{cool_cap_fff_curve.coefficient5z}*#{var4_sensor.name})")
+          fault_program.addLine("Set a1_AF_Qgr_c = 1 - #{cool_cap_fff_curve.coefficient4y}")
           fault_program.addLine("Set a2_AF_Qgr_c = #{cool_cap_fff_curve.coefficient4y}")
           fault_program.addLine('Set a3_AF_Qgr_c = 0')
-          fault_program.addLine("Set a1_AF_EIR_c = #{cool_eir_fff_curve.coefficient1Constant} + (#{cool_eir_fff_curve.coefficient2w}*#{var1_sensor.name}) + (#{cool_eir_fff_curve.coefficient3x}*#{var2_sensor.name}) + (#{cool_eir_fff_curve.coefficient5z}*#{var4_sensor.name})")
+          fault_program.addLine("Set a1_AF_EIR_c = 1 - #{cool_eir_fff_curve.coefficient4y}")
           fault_program.addLine("Set a2_AF_EIR_c = #{cool_eir_fff_curve.coefficient4y}")
           fault_program.addLine('Set a3_AF_EIR_c = 0')
+          fault_program.addLine("Set AF_Qgr_residual = #{cool_cap_fff_curve.coefficient1Constant} + (#{cool_cap_fff_curve.coefficient2w}*#{var1_sensor.name}) + (#{cool_cap_fff_curve.coefficient3x}*#{var2_sensor.name}) + (#{cool_cap_fff_curve.coefficient5z}*#{var4_sensor.name}) - 1") # residual from other terms
+          fault_program.addLine("Set AF_EIR_residual = #{cool_eir_fff_curve.coefficient1Constant} + (#{cool_eir_fff_curve.coefficient2w}*#{var1_sensor.name}) + (#{cool_eir_fff_curve.coefficient3x}*#{var2_sensor.name}) + (#{cool_eir_fff_curve.coefficient5z}*#{var4_sensor.name}) - 1") # residual from other terms
         end
 
         qgr_values, p_values, ff_chg_values = get_installation_quality_cooling_coeff(f_chg)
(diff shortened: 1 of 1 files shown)


yzhou601:
The GSHP(where charge defect ratio = 0) test result running on this branch is now consistent with removing the EMS program. I ran the ERI rated home hpxml (the master one @shorowit sent me, because the iq test file had bugs in sizing) locally on this branch, and then removed EMS program for comparison, the energy diffs are as small as 0.02GJ (42.80GJ vs 42.78GJ). This is a good sign.
However, I think there's a bug in [setting rated airflow](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac.rb#L413) for GSHP coil. This line takes ACTUAL airflow after applying defect ratio instead of before applying defect ratio. This will cause the actual always equal to rated, and the curve calculation will always have a 1.0 for airflow variable, which is not true. So we should use clg_cfm/airflow_defect_ratio here as rated. In the local test, I changed the idf manually to be correct rated airflow, and reran the above test, the energy number is also very close (46.73GJ vs 46.75GJ). 
So this approach is now looking good in the 0 charge defect ratio case. Can be a candidate implementation pending for discussion.

yzhou601:
06/21 meeting with Jon and Scott:
We will proceed Implementing the IQ equations with Cutler curve coefficients, calculating the adjustment factor and then applying it to the GSHP curve output. (proposed approach for better consistency with other systems)
Might also need to change two-speed system and variable speed system to use Cutler curve coefficients in the curve adjustment calculation.

yzhou601:
@shorowit I fixed an issue in adjustment calculation in the EMS program, it used to be calculating the adjustment ratio by: 
  
    Cutler adjustment ratio = IQ airflow curve value (airflow + charge defect applied) / Cutler airflow curve value(calculated with defect airflow), 
    Curve value after adjustment = current curve value calculated with defect airflow * Cutler adjustment ratio

after the fix, it's now:

    Cutler adjustment ratio = IQ airflow curve value (airflow + charge defect applied) / Cutler airflow curve value(calculated with no-defect airflow), 
    Curve value after adjustment = curve value calculated with no-defect airflow * Cutler adjustment ratio

The previous equation actually didn't apply cutler coefficients in airflow defect impact calculations. So this change is supposed to make airflow defect impact more consistent across different system types. I ran those test files with latest changes (I didn't exclude fan powers), and see below the results:

Charge defect ratio:
image

Airflow defect ratio:
image

However, it didn't make GSHP results closer with other systems, but even in opposite direction. I dig down to the time step outputs and found GSHP's runtime fraction change after applying airflow defect is the cause. 

    GSHP power consumption = Rated power * power curve value * RTF
    ASHP power consumption = Curve adjusted capacity * Curve adjusted EIR * RTF

So the power curve value is now adjusted by a similar ratio compared with ASHP (EIR curve adj * CAP curve adj), but the RTF is increased 34% in average for heating, while ASHP only increased 6.7% in average for heating. As far as I know, RTF calculation for ASHP is determined by PLR/PLF, where PLF is the curve value, while the RTF for gshp is still a bit vague to me. Considering the charge defect impact is pretty consistent across system types, I guess the RTF calculation must of a different relationship to airflow rate in these two system types. So at this point I am not sure how should we resolve this discrepancy, further investigation on RTF might be a bigger effort than we thought.

shorowit:
Related to https://github.com/NREL/EnergyPlus/issues/8955

yzhou601:
@shorowit It's not able to directly run workflow with customized E+ build on 8955 PR, need latest OS to resolve those new breaking changes in E+ 9.6. Is it ok to run workflow to generate gshp idf and manually change it to be compatible with latest E+, then just check hvac energy? The reporting measure is not able to apply in this case.

shorowit:
@yzhou601 Oh, right. Yes, please try that. I don't think there's any other way to test it.

yzhou601:
@shorowit New GSHP runs with latest E+(bugfix merged today! 🎉 )

image

The results are making more sense now. Heating impact is basically consistent with other systems. The energy reduce on cooling is mostly caused by fan energy reduce. I checked pump and coil electricity, they both went up with defects.

Changelog.md
```diff
@@ -32,6 +32,7 @@ __New Features__
 - Revises shared mechanical ventilation preconditioning control logic to operate less often.
 - Removes error-check for number of bedrooms based on conditioned floor area, per RESNET guidance.
 - Updates the reporting measure to register all outputs from the annual CSV with the OS runner (for use in, e.g., PAT).
+- Use constant Cutler curve coefficients in installation quality calculations instead of coil curve coefficients for two-speed, variable-speed systems and ground source heat pump.
```
shorowit, line 35: Is there anything else from this PR that should go in the changelog? The PR description says "To fix an issue where quadlinear/quintlinear curve is not reaching 1 for rated airflow in EMS program."

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3712,18 +3714,186 @@ def self.get_installation_quality_cooling_coeff(f_chg)
...
+    ff_chg_values = [0.0, 8.33] # Add a zero term to combine cooling and heating claculation
     return qgr_values, p_values, ff_chg_values
   end
 
+  def self.get_airflow_fault_cooling_coeff()
+    # Cutler curve coefficients for single speed
+    cool_cap_fflow_spec = [0.718664047, 0.41797409, -0.136638137]
+    cool_eir_fflow_spec = [1.143487507, -0.13943972, -0.004047787]
+    return cool_cap_fflow_spec, cool_eir_fflow_spec
+  end
```
shorowit, line 3734: Can you call this method from `set_cool_curves_ashp` so that we don't have the same coefficients two different places?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3712,18 +3714,186 @@ def self.get_installation_quality_cooling_coeff(f_chg)
...
+
+  def self.get_airflow_fault_heating_coeff()
+    # Cutler curve coefficients for single speed
+    heat_cap_fflow_spec = [0.694045465, 0.474207981, -0.168253446]
+    heat_eir_fflow_spec = [2.185418751, -1.942827919, 0.757409168]
+    return heat_cap_fflow_spec, heat_eir_fflow_spec
+  end
```
shorowit, line 3741: Same as above.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3712,18 +3714,186 @@ def self.get_installation_quality_cooling_coeff(f_chg)
...
+
+  def self.add_install_quality_calculations(fault_program, tin_sensor, tout_sensor, airflow_rated_defect_ratio, clg_or_htg_coil, model, f_chg, obj_name, mode, defect_ratio)
```
shorowit, line 3743: The code was collapsed into a single method that can be used for heating or cooling. Nice. 👍

shorowit (approved):
This looks good. Just some minor things to address and then this can get merged in.

Changelog.md
yzhou601: That language is from using ground source heat pump curve coefficients in IQ calculation, I think it's no longer applicable, I should have removed it from description.

shorowit:
This is basically ready to go, but we'll wait on pulling in EnergyPlus v9.6 first just to confirm results.

shorowit:
@yzhou601 Do you want to pull in the `os330` branch into here and do final verification of results? If you do that, you'll also want to change the base branch here to os330. If we're still happy, we can then merge this in.

yzhou601:
CI only shows diffs on base-hvac-install-quality-foo test files where two speed, variable speed, and gshp systems are installed. This is expected due to the replacement of curve coefficients in IQ calculations. 
And I compared the CI between before and after pulling in os330, only gshp test files showed a reduce of energy consumption (this is the expected behavior of new E+, which should show a smaller impact on energy by airflow defect), other test files are showing equivalent results to E+9.5.
Is this ready to go? Any other tests to run? @shorowit

PR #805, opened 2021-07-14
BuildResidentialHPXML measure: 2-story SFA home bugfix

shorowit:
## Pull Request Description

Fixes a bug where the adiabatic walls were not translated correctly to HPXML for a 2-story SFA home. Spotted by @aspeake.

Verified that all existing test files are 100% identical. Added a new 2-story SFA home test file where the HPXML in this branch has the two additional (adiabatic) walls relative to the base branch.

## Checklist

Not all may apply:


shorowit:
The fix was simply to remove the `get_adiabatic_adjacent_surface` method, which returned nil for every test file we have. However, it was (_incorrectly_) not returning nil for this new test file. I don't know why this method exists, but it wasn't doing whatever it was intending to do. Maybe this is not the appropriate fix, but we have no test files that indicate otherwise.

shorowit:
The CI is going to fail because the window areas in this new test file end up differing by 1 sqft. Need to figure out why that is.

shorowit:
@joseph-robertson @aspeake Can either of you explain why this method existed? You both have fingerprints in it (see below). If it turns out that there is indeed a reason for it, then we need to add a test that _actually_ exercises the code.

image

aspeake:
> 
> 
> @joseph-robertson @aspeake Can either of you explain why this method existed? You both have fingerprints in it (see below). If it turns out that there is indeed a reason for it, then we need to add a test that _actually_ exercises the code.
> 
> image

My edits just added the `sort` for the coordinate values, since there were cases where we had an adjacent surface but it wasn't returned by this method. I can't speak on the origin of the method, but it does get used in various places, I made my changes to try and ensure that doors were added to MF buildings with corridors here: https://github.com/NREL/OpenStudio-HPXML/blob/build-res-hpxml-v3/BuildResidentialHPXML/measure.rb#L3974-L3996

shorowit:
Note that the `sort` lines were actually the cause of the bug, because when you have a 2-story building, the 2nd story wall has the same coordinates as the 1st story wall (relative to their respective space origins). So the method returned an incorrect non-nil value when it should have returned nil.

More importantly, as far as I can tell, the single-unit model approach **never** produces an adjacent surface/space in the model. So this method should always returns nil. Right? Is this method just left over from the days before the single-unit model approach was implemented?

shorowit:
After discussing further with @aspeake, I see now that this code did serve a purpose, but none of our test files exercised it. Will add test files that exercise it and then properly address the issue.

BuildResidentialHPXML/resources/geometry.rb
```diff
@@ -712,9 +712,10 @@ def self.create_single_family_detached(runner:,
   end
 
   def self.has_same_vertices(surface1, surface2)
-    if getSurfaceXValues([surface1]) == getSurfaceXValues([surface2]) &&
-       getSurfaceYValues([surface1]) == getSurfaceYValues([surface2]) &&
-       getSurfaceZValues([surface1]) == getSurfaceZValues([surface2])
+    if getSurfaceXValues([surface1]).sort == getSurfaceXValues([surface2]).sort &&
+       getSurfaceYValues([surface1]).sort == getSurfaceYValues([surface2]).sort &&
+       getSurfaceZValues([surface1]).sort == getSurfaceZValues([surface2]).sort &&
+       surface1.space.get.zOrigin.round(5) == surface2.space.get.zOrigin.round(5)
```
shorowit, line 718: Adds z coordinate matching of the surfaces' spaces (to prevent 2nd story walls from being matched with 1st story walls) and incorporates the `.sort` here.

BuildResidentialHPXML/tests/base-bldgtype-single-family-attached-2stories.osw
```diff
@@ -0,0 +1,339 @@
+{
```
shorowit, line 1: Adds a test file that differs between here and the base branch. All other test files show identical results.

BuildResidentialHPXML/measure.rb
```diff
@@ -3529,7 +3529,7 @@ def self.set_roofs(hpxml, runner, model, args)
       hpxml.roofs.add(id: valid_attr(surface.name),
                       interior_adjacent_to: get_adjacent_to(surface),
                       azimuth: azimuth,
-                      area: UnitConversions.convert(surface.grossArea, 'm^2', 'ft^2').round(1),
+                      area: UnitConversions.convert(surface.grossArea, 'm^2', 'ft^2').round(2),
```
shorowit, line 3532: Adds a second decimal to all areas so that the window areas perfectly match in the new test file.

aspeake (approved):

craxrubra:
Hi, I am from Mexico. Does this application runs independently from Sketchup for residential geometry? Does climate information can be add?

BuildResidentialHPXML/resources/geometry.rb
joseph-robertson: @shorowit Is this the actual bugfix?

shorowit: Yes

joseph-robertson (approved):

PR #817, opened 2021-08-05
BuildResidentialScheduleFile

### LINKED ISSUE #814: Create `BuildResidentialScheduleFile` measure
Pull code out of the BuildResidentialHPXML measure and make its own measure. That will allow the schedule generator to be used in additional workflows (i.e., that do not include the BuildResidentialHPXML measure).

### BuildResidentialScheduleFile measure
- Creates schedule csv file and adds path to HPXML file.
- Arguments:
  - `hpxml_path` (used to determine state, timestep, is_leap_year)
  - `schedules_type` ("smooth" or "stochastic")
  - `vacancy_period` (current argument description in BuildResHPXML measure says it only applies if `schedules_type` is "stochastic", but doesn't it also apply if "smooth"?)
  - `random_seed`
  - `output_csv_path`
- (For ResStock-HPXML it might make sense to include these arguments in the `BuildExistingModel` measure, except for `schedules_type` and `vacancy_period` since they are sampled.)

### BuildResidentialHPXML measure
- Largely unchanged other than modifications to a few arguments.
- Arguments:
  - add `state` (optional input; if defaulted, uses EPW state as currently the case. EPW may not be in the same state as the house location, that's why we should add an explicit input); resstock develop passes sampled `state` to generator measure
  - remove `schedules_type`
  - remove `schedules_path`
  - remove `schedules_vacancy_period`
  - remove `schedules_random_seed`

## Workflows

This would support multiple workflows that want to use more detailed and/or stochastic schedules:
1. BuildResidentialHPXML  -> BuildResidentialScheduleFile  -> HPXMLtoOpenStudio
2. third-party software tool -> HPXML file -> BuildResidentialScheduleFile  -> HPXMLtoOpenStudio
3. third-party software tool -> HPXML file & schedule CSV -> HPXMLtoOpenStudio

Questions:
* Should we have BuildResidentialScheduleFile update the `SchedulesFilePath` field of the hpxml file? Otherwise, you could run into an issue where your hpxml file doesn't point to the schedule file that you generate. The alternative might be to have a `output_csv_path` argument to HPXMLtoOpenStudio.
* Results are not identical between when not applying BuildResidentialScheduleFile and when applying it with `schedules_type=smooth`; should they be?

joseph-robertson:
## Pull Request Description

Closes #814.

## Checklist

Not all may apply:


workflow/run_simulation.rb
```diff
@@ -11,11 +11,21 @@
 basedir = File.expand_path(File.dirname(__FILE__))
 
 def run_workflow(basedir, rundir, hpxml, debug, timeseries_output_freq, timeseries_outputs, skip_validation, add_comp_loads,
-                 output_format, building_id, ep_input_format)
+                 output_format, building_id, ep_input_format, schedules_type)
```
joseph-robertson, line 14: Do we want more than just the `schedules_type` argument?

BuildResidentialScheduleFile/measure.rb
```diff
@@ -0,0 +1,182 @@
...
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeIntegerArgument('schedules_random_seed', false)
+    arg.setDisplayName('Schedules: Random Seed')
+    arg.setUnits('#')
+    arg.setDescription("This numeric field is the seed for the random number generator. Only applies if the schedules type is 'stochastic'.")
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeStringArgument('output_csv_path', true)
+    arg.setDisplayName('Schedules: Path')
```
shorowit, line 67:
```suggestion
    arg.setDisplayName('Schedules: Output CSV Path')
```

BuildResidentialScheduleFile/measure.rb
```diff
@@ -0,0 +1,182 @@
...
+    epw_file = OpenStudio::EpwFile.new(epw_path)
+
+    success = create_schedules(runner, hpxml, model, epw_file, args)
+    return false if not success
+
+    # modify the hpxml with the schedules path
+    hpxml.header.schedules_filepath = args[:output_csv_path]
+    hpxml_doc = hpxml.to_oga()
+    XMLHelper.write_file(hpxml_doc, hpxml_path)
+    runner.registerInfo("Wrote file: #{hpxml_path}")
```
shorowit, line 115: We shouldn't use the HPXML class to update the HPXML file, as the class is not comprehensive and does not cover every possible HPXML element. If someone had created their own custom HPXML file (i.e., not through the HPXML class) and then you read in and write the file using the HPXML class, information will be lost. Rather, you should use the XMLHelper or Oga methods directly to insert/update the schedules path.

(It is fine to read in the HPXML file using the HPXML class in order to retrieve the state code, etc. We just can't use it to update/write the file.)

BuildResidentialScheduleFile/measure.rb
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  # human readable description
+  def description
+    return 'Builds a residential schedule file.'
+  end
+
+  # human readable description of modeling approach
+  def modeler_description
+    return ''
```
shorowit, line 33: Add some description about how it creates a CSV file and inserts the CSV schedule file path into the HPXML file.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -0,0 +1,1048 @@
...
+    end
+    # shape of all_simulated_values is [2, 35040, 7] i.e. (geometry_num_occupants, period_in_a_year, number_of_states)
+    plugload_sch = schedule_config['plugload']
+    lighting_sch = schedule_config['lighting']
+    ceiling_fan_sch = schedule_config['ceiling_fan']
+
+    monthly_lighting_schedule = schedule_config['lighting']['monthly_multiplier']
+    holiday_lighting_schedule = schedule_config['lighting']['holiday_sch']
+
+    sch = Lighting.get_schedule(@model, @epw_file)
```
shorowit, line 363: It looks like Lighting.get_schedule doesn't actually use the `model`. Can you remove this argument and update all references?

BuildResidentialScheduleFile/measure.rb
```diff
@@ -0,0 +1,182 @@
...
+    if epw_file.startDateActualYear.is_initialized # AMY
+      year_description.setCalendarYear(epw_file.startDateActualYear.get)
+    end
+
+    # set the timestep
+    timestep = model.getTimestep
+    timestep.setNumberOfTimestepsPerHour(1)
+    unless hpxml.header.timestep.nil?
+      timestep.setNumberOfTimestepsPerHour(60 / hpxml.header.timestep)
+    end
```
shorowit, line 138: I'm a bit concerned about updating the model in this measure. How about being safe and creating a new OpenStudio::Model::Model that you make these changes to and then pass into the ScheduleGenerator? This will ensure the active model is unchanged.

BuildResidentialScheduleFile/tests/build_residential_schedule_file_test.rb
```diff
@@ -0,0 +1,459 @@
...
+    assert(!info_msgs.any? { |info_msg| info_msg.include?('RandomSeed') })
+    assert(info_msgs.any? { |info_msg| info_msg.include?('GeometryNumOccupants=3.0') })
+    assert(info_msgs.any? { |info_msg| info_msg.include?('VacancyPeriod=Dec 1 - Jan 31') })
+
+    sf = SchedulesFile.new(model: model, schedules_path: @args_hash['output_csv_path'], col_names: Constants.ScheduleColNames.keys)
+
+    vacancy_hrs = 31 * 2 * 24
+    occupied_ratio = (1 - vacancy_hrs / 8760)
+
+    assert_in_epsilon(4997 * occupied_ratio, sf.annual_equivalent_full_load_hrs(col_name: 'occupants', schedules: sf.tmp_schedules), 0.1)
```
shorowit, line 97: I'm probably just not thinking straight, but why is the 4997 number here different than the 6020 number from `test_smooth`? I thought you get the same schedules in the CSV either way, except for a vacancy column that will now have 1s in it.

workflow/run_simulation.rb
shorowit: Hmm. I will have to think more about what we want to do with the run_simulation.rb script.

workflow/template2.osw
```diff
@@ -0,0 +1,47 @@
+{
```
shorowit, line 1: I'm not really sure what you're going for here. If someone tries to run this OSW, they will get errors and be confused. If this is actually meant to be another example template OSW file, then A) we should use more descriptive names to distinguish the multiple template OSWs and B) we should construct the OSWs so that they can run as is.

shorowit (commented):
This is a good start. Some comments below. We'll also need to:
- [ ] update the documentation
- [ ] update the `create_release_zips` task in tasks.rb
- [ ] update the changelog
- [x] determine why results are not identical between when not applying BuildResidentialScheduleFile and when applying it with `schedules_type=smooth` (EDIT: Fixed by https://github.com/NREL/OpenStudio-HPXML/pull/829)

BuildResidentialScheduleFile/measure.rb
shorowit: Even better would be to not pass in the model (or even epw_file?) to the ScheduleGenerator and just pass in the minimal information needed (timestep, leap year, etc.). Then you wouldn't have to make any changes to _any_ model object in this code here. But I don't know how easy that would be.

BuildResidentialScheduleFile/tests/build_residential_schedule_file_test.rb
joseph-robertson: Good catch. Because, e.g., `1488/8760 = 0` in ruby.

BuildResidentialScheduleFile/measure.rb
joseph-robertson: @shorowit You mentioned "[we] might need to add an argument to the write_file method to bypass a lot of the stuff it's doing". Can you implement that here?

workflow/template2.osw
joseph-robertson: A) Yes, we could rename them if there are 2+ template OSWs.

B) The problem with referencing an actual sample file is that it would overwrite it. Maybe that's ok? Suggestions?

BuildResidentialScheduleFile/measure.rb
joseph-robertson: I think we still need to pass in `model` (e.g., `set_vacancy` needs it), but we could get away with not updating `model`.

shorowit: It looks like we are really close to not needing the `model` object at all. I'm going to simplify/refactor some methods so we can avoid it.

BuildResidentialScheduleFile/measure.rb
```diff
@@ -0,0 +1,199 @@
...
+    doc = XMLHelper.parse_file(hpxml_path)
+    software_info = XMLHelper.get_element(doc, '/HPXML/SoftwareInfo')
+    extension = XMLHelper.create_elements_as_needed(software_info, ['extension'])
+    schedules_filepath = XMLHelper.get_value(extension, 'SchedulesFilePath', :string)
+    if !schedules_filepath.nil?
+      XMLHelper.delete_element(extension, 'SchedulesFilePath')
+    end
+    XMLHelper.add_element(extension, 'SchedulesFilePath', args[:output_csv_path], :string)
+    XMLHelper.write_file(doc, hpxml_path)
+    runner.registerInfo("Wrote file: #{hpxml_path}")
```
shorowit, line 122: @joseph-robertson Can you add a warning if we are overwriting an existing SchedulesFilePath element?

joseph-robertson:
How do you suppose we should update the documentation for this? Add a new section "Schedules" at the Workflow Inputs and Workflow Outputs level? Or add a new subsection "Schedules" under Workflow Inputs, just before the Using HPXML subsection?

PR #820, opened 2021-08-10
BuildResidentalHPXML measure: Generate HES compatible xmls

aspeake:
## Pull Request Description

Updates inputs and xml creation so that ResStock-generated xmls are compatible with the hescore-hpxml translator (NREL/hescore-hpxml#151)

## Checklist

Not all may apply:


BuildResidentialHPXML/measure.rb
```diff
@@ -3318,20 +3345,63 @@ def self.set_air_infiltration_measurements(hpxml, runner, args)
...
 
+    surf_ids = { 'slabs' => { 'surfaces' => hpxml.slabs, 'ids' => [] },
+                 'frame_floors' => { 'surfaces' => hpxml.frame_floors, 'ids' => [] },
+                 'foundation_walls' => { 'surfaces' => hpxml.foundation_walls, 'ids' => [] },
+                 'walls' => { 'surfaces' => hpxml.walls, 'ids' => [] },
+                 'rim_joists' => { 'surfaces' => hpxml.rim_joists, 'ids' => [] }, }
+
+    foundation_locations = [HPXML::LocationBasementConditioned, HPXML::LocationBasementUnconditioned,
+                            HPXML::LocationCrawlspaceUnvented, HPXML::LocationCrawlspaceVented,
+                            HPXML::FoundationTypeAmbient]
```
shorowit, line 3385: Since this is a HPXML::Foundation enum and not a HPXML::Location enum, I have a hard time believing this works correctly for ambient foundations. Have you double-checked this?

BuildResidentialHPXML/measure.rb
```diff
@@ -3318,20 +3345,63 @@ def self.set_air_infiltration_measurements(hpxml, runner, args)
...
+
+    surf_ids.each do |surf_type, surf_hash|
+      surf_hash['surfaces'].each do |surface|
+        next if (not foundation_locations.include? surface.interior_adjacent_to) &&
+                (not foundation_locations.include? surface.exterior_adjacent_to) &&
+                (surf_type != 'slabs') &&
+                (surf_type != 'foundation_walls')
```
shorowit, line 3392: Why is this code here? Doesn't that mean you won't get the surfaces adjacent to, e.g., a frame floor or rim joist?

BuildResidentialHPXML/measure.rb
```diff
@@ -3452,13 +3522,22 @@ def self.set_walls(hpxml, runner, model, args)
 
       next if exterior_adjacent_to == HPXML::LocationLivingSpace # already captured these surfaces
 
+      attic_wall_type = nil
+      if ([HPXML::LocationAtticUnvented, HPXML::LocationAtticVented].include? interior_adjacent_to) && (exterior_adjacent_to == HPXML::LocationOutside)
```
shorowit, line 3526: Should probably include `HPXML::LocationAtticUnconditioned` (as you did above) just to be comprehensive.

BuildResidentialHPXML/measure.rb
```diff
@@ -3452,13 +3522,22 @@ def self.set_walls(hpxml, runner, model, args)
...
+
       wall_type = args[:wall_type]
       if [HPXML::LocationAtticUnvented, HPXML::LocationAtticVented].include? interior_adjacent_to
         wall_type = HPXML::WallTypeWoodStud
       end
 
       if exterior_adjacent_to == HPXML::LocationOutside && args[:wall_siding_type].is_initialized
-        siding = args[:wall_siding_type].get
+        if ([HPXML::LocationAtticUnvented, HPXML::LocationAtticVented].include? interior_adjacent_to) && (args[:wall_siding_type].get == HPXML::SidingTypeNone)
+          siding = HPXML::SidingTypeVinyl
```
shorowit, line 3537: This is not an appropriate change. If the user doesn't choose a siding, we need the resulting HPXML to reflect that that input wasn't provided. (That doesn't mean the building will have no siding, just that it will be defaulted downstream in the HPXMLtoOpenStudio measure.) 

I'm not sure what you're trying to accomplish here. Can you explain?

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -1915,6 +2013,7 @@ def from_oga(wall)
       @id = HPXML::get_id(wall)
       @exterior_adjacent_to = XMLHelper.get_value(wall, 'ExteriorAdjacentTo', :string)
       @interior_adjacent_to = XMLHelper.get_value(wall, 'InteriorAdjacentTo', :string)
+      @attic_wall_type = XMLHelper.get_value(wall, 'AtticWallType', :string)
```
shorowit, line 2016: 👍 Great job updating the HPXML class. It all looks good.

tasks.rb
```diff
@@ -3265,12 +3265,15 @@ def set_hpxml_site(hpxml_file, hpxml)
   if ['base.xml'].include? hpxml_file
     hpxml.site.fuels = [HPXML::FuelTypeElectricity, HPXML::FuelTypeNaturalGas]
     hpxml.site.site_type = HPXML::SiteTypeSuburban
+    hpxml.site.azimuth_of_front_of_home = 180.0
```
shorowit, line 3268: Since none of these new BuildResHPXML inputs are used by the HPXMLtoOpenStudio measure, I'd prefer that the sample files not add this extra information. It might make users creating HPXML files think that this information is required/used.

My guess is that you added this so that the BuildResHPXML measure tests pass? If that's the case, I would revert the changes to tasks.rb and instead simply [exclude these elements from being compared](https://github.com/NREL/OpenStudio-HPXML/blob/85ff0bf2565cefa16284a7fc123a7af4cc4a04ff/BuildResidentialHPXML/tests/build_residential_hpxml_test.rb#L211) in the BuildResHPXML measure tests.

shorowit (commented):
This is really good work, @aspeake. I have just a few comments.

BuildResidentialHPXML/measure.rb
aspeake: I added because of this comment: https://github.com/NREL/resstock/pull/668#discussion_r681331384, but what surfaces are ever attached to ambient foundations? Should I remove the foundation type or is there an ambient location to check?

When I run a building with ambient there are no surfaces adjacent to the foundation

BuildResidentialHPXML/measure.rb
aspeake: This ensures that we are always including slabs and foundation walls to surfaces attached to foundations, however I don't think `foundation_walls` is necessary, since those will always be adjacent to the foundation space which is already checked.

BuildResidentialHPXML/measure.rb
aspeake: Addressed in a few places

BuildResidentialHPXML/measure.rb
aspeake: This is meant to prevent wood stud gable walls from having `None` siding. This occurs, for example, when brick/CMU wall siding is set to `None`, but gable walls are forced to be wood stud in this measure. It is supposed to mimic what is done in `resstock`: https://github.com/NREL/resstock/blob/develop/resources/measures/ResidentialConstructionsWallsCMU/measure.rb#L191-L196  

So instead of assigning the siding here, should it get assigned here instead? https://github.com/NREL/OpenStudio-HPXML/blob/build-res-hpxml-v3-hes/HPXMLtoOpenStudio/measure.rb#L667

BuildResidentialHPXML/measure.rb
shorowit: I'm still confused. Slabs and foundation walls will have the same InteriorAdjacentTo; if it's not necessary for foundation walls, it shouldn't be necessary for slabs. Or am I missing something?

BuildResidentialHPXML/measure.rb
shorowit: I guess my recommendation would be to use `siding = nil` here. That will create an HPXML file where the gable walls have no siding specified -- which will then get defaulted downstream in the HPXMLtoOpenStudio measure to wood siding. It's not vinyl siding, but it's close enough.

BuildResidentialHPXML/measure.rb
shorowit: We might have to get @nmerket to weigh in here. As far as I can tell, ambient foundations [are treated as crawlspaces](https://hescore-hpxml.readthedocs.io/en/latest/translation/zone_floor.html#foundation-type) in hescore-hpxml, which means that it's going to look for foundation walls and a floor above. We have a floor above, which you should be able to update this code to handle, but we don't have any foundation walls -- I don't know what the hpxml-hescore translator will do if it finds this scenario.

tasks.rb
aspeake: Ah was not aware of that, I reverted tasks.rb and excluded the 3 fields.

BuildResidentialHPXML/measure.rb
aspeake: The slabs check is only needed when there is a slab foundation, which is adjacent to the ground and the living space.

BuildResidentialHPXML/measure.rb
aspeake: Okay that makes sense, however this will end up causing an error with the hescore-hpxml translator, which is what prompted the change originally. Should I set this value in hpxml_defaults so that it is valid in `in.xml`?

shorowit: I thought hescore-hpxml completely ignores gable walls?

BuildResidentialHPXML/measure.rb
shorowit: As written, won't the logic end up including e.g. a garage slab as attached to the foundation?

BuildResidentialHPXML/measure.rb
aspeake: If I try to run a building with a wood stud wall and no siding (gable wall or otherwise) it will throw a translation error

shorowit: Okay, but the code above is only addressing gable walls.

aspeake: Yes, I get a translation error when running with a wood stud gable wall and no siding

BuildResidentialHPXML/measure.rb
aspeake: Yes it would, should it only include living space slabs?

BuildResidentialHPXML/measure.rb
shorowit: It sounds like hescore-hpxml needs to be fixed to ignore gable walls then.

BuildResidentialHPXML/measure.rb
shorowit: Yes

aspeake: Addressed in https://github.com/NREL/OpenStudio-HPXML/pull/820/commits/e8407f1b3579308e4ba42a95f5b69bc2e8631931 and re-wrote the logic to make it easier to understand

shorowit (approved):
💯

Issue #823, opened 2021-08-16
Add cost multipliers


shorowit:
Integrate cost multipliers (e.g., [from ResStock](https://github.com/NREL/resstock/blob/c96a7012d6306a6596da04b501e02accafb9d1f5/measures/UpgradeCosts/measure.rb#L169-L284)) into the SimulationOutputReport measure so they can be used by additional software tools. Values could be added to the CSV as well as registered with the OS runner.

And perhaps we should call it a more generic "Building Summary" report instead of a "Cost Multipliers" report?

PR #824, opened 2021-08-16
ReportHPXMLOutput reporting measure

joseph-robertson:
## Pull Request Description

Closes https://github.com/NREL/OpenStudio-HPXML/issues/823.

Generates an output file called results_hpxml.csv (or results_hpxml.json), co-located with the EnergyPlus output. Unlike the other output files that are based on EnergyPlus simulation results, this file contains outputs summarizing the HPXML file (including defaults like auto-sized HVAC systems). The file includes the following sections of output:

Enclosure: Wall Area Thermal Boundary (ft^2)
Enclosure: Wall Area Exterior (ft^2)
Enclosure: Foundation Wall Area Exterior   (ft^2)
Enclosure: Floor Area Conditioned (ft^2)
Enclosure: Floor Area Lighting (ft^2)
Enclosure: Ceiling Area Thermal Boundary   (ft^2)
Enclosure: Roof Area (ft^2)
Enclosure: Window Area (ft^2)
Enclosure: Door Area (ft^2)
Enclosure: Duct Area Unconditioned (ft^2)
Enclosure: Rim Joist Area (ft^2)
Enclosure: Slab Exposed Perimeter Thermal   Boundary (ft)
 
Systems: Cooling Capacity (kBtu/h)
Systems: Heating Capacity (kBtu/h)
Systems: Heat Pump Backup Capacity   (kBtu/h)
Systems: Water Heater Tank Volume (gal)
Systems: Mechanical Ventilation Flow Rate   (cfm)

If the HPXML file has `Systems/HVAC/HVACPlant/PrimarySystems` populated, then additional system outputs will be provided:

Primary Systems:   Cooling Capacity (kBtu/h)
Primary Systems: Heating Capacity   (kBtu/h)
Primary Systems: Heat Pump Backup   Capacity (kBtu/h)
 
Secondary Systems: Cooling Capacity   (kBtu/h)
Secondary Systems: Heating Capacity   (kBtu/h)
Secondary Systems: Heat Pump Backup   Capacity (kBtu/h)

## Checklist

Not all may apply:


joseph-robertson:
@shorowit I think one argument for having this code in SimulationOutputReport, and not a new separate measure, is to maintain a simpler workflow which minimizes the number of measures called.

shorowit:
On the other hand, this measure is called SimulationOutputReport and these are not _simulation_ outputs.

joseph-robertson:
We've decided to incorporate this code not into SimulationOutputReport, but instead into its own measure. Something like HPXMLOutputReport.

joseph-robertson:
@shorowit This is ready for your review.

joseph-robertson:
@shorowit I could use some pointers/guidance on the `PrimarySystems` stuff. I'm trying to pass a `primary_system` argument into, e.g., the `HeatingSystem` object and have it populate `PrimarySystems/PrimaryHeatingSystem idref=foo`. Can you take a look? Is this a good approach?

shorowit:
> @shorowit I could use some pointers/guidance on the `PrimarySystems` stuff. I'm trying to pass a `primary_system` argument into, e.g., the `HeatingSystem` object and have it populate `PrimarySystems/PrimaryHeatingSystem idref=foo`. Can you take a look? Is this a good approach?

I like that you created a `primary_system` boolean in the individual HVAC (HeatingSystem, CoolingSystem, HeatPump) classes. This is similar to, e.g., `primary_indicator` in the Refrigerator class. I suggest two changes though:
1. Remove the [PrimarySystems class](https://github.com/NREL/OpenStudio-HPXML/blob/57f61b022a9a0f3d5307db8b8c434bbcfe86844c/HPXMLtoOpenStudio/resources/hpxml.rb#L2741), which you only use to return an array of HPXML HVAC objects. This seems overly complicated. Instead, just create a top-level HPXML class method (say, above the [total_fraction_cool_load_served method](https://github.com/NREL/OpenStudio-HPXML/blob/57f61b022a9a0f3d5307db8b8c434bbcfe86844c/HPXMLtoOpenStudio/resources/hpxml.rb#L444)) that returns the array of HPXML HVAC objects. It just needs a one-liner like `return hvac_systems.select{ |h| h.primary_system }`.
2. Only [create the PrimarySystems element](https://github.com/NREL/OpenStudio-HPXML/blob/57f61b022a9a0f3d5307db8b8c434bbcfe86844c/HPXMLtoOpenStudio/resources/hpxml.rb#L2842) if there is an HVAC system specified as primary to avoid creating empty `<PrimarySystem/>` elements in every sample HPXML file.

joseph-robertson:
For the second item above, I struggled with this because I was finding that the `PrimarySystems` block could be inserted NOT as the first `HVACPlant` entry...

shorowit:
If you have a top-level method called, say, `primary_hvac_systems`, you could just update the code from
`primary_systems = XMLHelper.create_elements_as_needed(doc, ['HPXML', 'Building', 'BuildingDetails', 'Systems', 'HVAC', 'HVACPlant', 'PrimarySystems'])`
to
`primary_systems = XMLHelper.create_elements_as_needed(doc, ['HPXML', 'Building', 'BuildingDetails', 'Systems', 'HVAC', 'HVACPlant', 'PrimarySystems']) unless primary_hvac_systems.empty?`

HPXMLOutputReport/measure.rb
```diff
@@ -0,0 +1,418 @@
...
+    return cost_mult
+  end
+
+  def assign_primary_and_secondary(hpxml, cost_multipliers)
+    has_primary_heating_system = false
+    hpxml.heating_systems.each do |heating_system|
+      next if !heating_system.primary_system
+
+      cost_mult = BaseOutput.new
+      cost_mult.output = cost_multipliers["#{BS::SizeHeatingSystem}: #{heating_system.id}"].output
```
shorowit, line 305: I've realized we are being pretty inconsistent between HVAC systems (reported by system) vs water heaters or mechanical ventilation (summed across systems).

I know I had suggested otherwise, but for this first cut, let's just stick with simple results until it's requested that we provide more detail. So let's just report A) the sum of HVAC systems for heating/cooling sizes (i.e., not outputs by system), and B) the primary/secondary sizes when applicable.

HPXMLOutputReport/measure.rb
```diff
@@ -0,0 +1,418 @@
...
+
+    # Report results
+    cost_multipliers.each do |cost_mult_type, cost_mult|
+      cost_mult_type_str = OpenStudio::toUnderscoreCase("#{cost_mult_type} #{cost_mult.units}")
+      cost_mult = cost_mult.output.round(2)
+      runner.registerValue(cost_mult_type_str, cost_mult)
+    end
+
+    # Write results
+    write_output(runner, cost_multipliers, output_format, output_path)
```
shorowit, line 141: Can you exclude the "Fixed" output? That is weird to see in the output results.

HPXMLOutputReport/measure.rb
```diff
@@ -0,0 +1,418 @@
...
+    end
+
+    # Primary and Secondary
+    if hpxml.primary_hvac_systems.size > 0
+      assign_primary_and_secondary(hpxml, cost_multipliers)
+    end
+
+    # Units
+    cost_multipliers.each do |cost_mult_type, cost_mult|
+      cost_mult.name = "Building Summary: #{cost_mult_type}"
```
shorowit, line 129: Rather than prefix every row with "Building Summary", I suggest segmenting by:
- "Surface Area: " (you can then remove "area" from the individual names)
- "Sizes: " (you can then remove "size" from the individual names
- "Other: " (mech vent flow rate and slab perimeter)

And please add blank lines between the sections (like we do in the results_annual.csv output).

HPXMLOutputReport/tests/hpxml_output_report_test.rb
```diff
@@ -0,0 +1,123 @@
...
+    assert_includes(actual_rows.sort, 'Building Summary: Size Heating System: HeatingSystem2 (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Cooling System: CoolingSystem2 (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Heating System: Primary (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Heating System: Secondary (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Cooling System: Primary (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Cooling System: Secondary (kBtu/h)')
+    assert_includes(actual_rows.sort, 'Building Summary: Size Heat Pump Backup: Secondary (kBtu/h)')
+  end
+
+  def test_base_xml
```
shorowit, line 54: Let's add another test like this but for a heat pump.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -3107,6 +3133,12 @@ def is_dual_fuel
       return true
     end
 
+    def primary_system
+      return true if @primary_heating_system || @primary_cooling_system
+
+      return false
+    end
```
shorowit, line 3140: Nice use of a helper method to simplify code elsewhere.

HPXMLtoOpenStudio/resources/xmlhelper.rb
```diff
@@ -25,6 +25,30 @@ def self.add_element(parent, element_name, value = nil, datatype = nil, defaulte
     return added
   end
 
+  # Inserts the child element with 'element_name' and sets its value. Returns the
+  # child element.
+  def self.insert_element(parent, element_name, index = 0, value = nil, datatype = nil, defaulted = false)
```
shorowit, line 30: This has copy/pasted a lot of code. To prevent redundant code, how about updating the `add_element` method to call this method w/ index=-1? I assume that would work.

tasks.rb
```diff
@@ -3145,6 +3146,9 @@ def set_hpxml_heating_systems(hpxml_file, hpxml)
     hpxml.heating_systems << hpxml.heating_systems[0].dup
     hpxml.heating_systems[1].id = 'HeatingSystem2'
     hpxml.heating_systems[1].distribution_system_idref = 'HVACDistribution2'
+  elsif ['base-hvac-multiple2.xml'].include? hpxml_file
+    hpxml.heating_systems[1].primary_system = true
+    hpxml.cooling_systems[-1].primary_system = true
```
shorowit, line 3151: I'd prefer that you just update the original base-hvac-multiple.xml file -- or did you need it to be a different file?

workflow/run_simulation.rb
```diff
@@ -53,6 +53,12 @@ def run_workflow(basedir, rundir, hpxml, debug, timeseries_output_freq, timeseri
   args['include_timeseries_weather'] = timeseries_outputs.include? 'weather'
   update_args_hash(measures, measure_subdir, args)
 
+  # Add hpxml output measure to workflow
+  measure_subdir = 'HPXMLOutputReport'
+  args = {}
+  args['output_format'] = output_format
+  update_args_hash(measures, measure_subdir, args)
```
shorowit, line 60: 👍

HPXMLOutputReport/measure.rb
```diff
@@ -0,0 +1,418 @@
...
+      return false
+    end
+    model.setSqlFile(sqlFile)
+
+    hpxml_defaults_path = model.getBuilding.additionalProperties.getFeatureAsString('hpxml_defaults_path').get
+    hpxml = HPXML.new(hpxml_path: hpxml_defaults_path)
+
+    # Set paths
+    output_dir = File.dirname(sqlFile.path.to_s)
+    output_path = File.join(output_dir, "hpxml_output.#{output_format}")
```
shorowit, line 77: Let's call it "results_hpxml" instead of "hpxml_output", as that will be consistent with the existing "results_annual" and "results_timeseries" output files.

shorowit (commented):
This is good stuff. I do suggest some changes that I think will improve this.

tasks.rb
joseph-robertson: There is no real need for `multiple2.xml`.

shorowit (approved):
This looks great to me. Good to merge if the CI comes back happy.

shorowit:
@joseph-robertson Note that I updated all the output names (sorry), so you'll need to update the ResStock column mapping. And the reporting measures are now ReportFoo instead of FooReport.

joseph-robertson:
> @joseph-robertson Note that I updated all the output names (sorry), so you'll need to update the ResStock column mapping. And the reporting measures are now ReportFoo instead of FooReport.

I won't need to update the ResStock column mapping because from UpgradeCosts I register outputs using their original names.

PR #842, opened 2021-08-31
Meta measure resource updates

joseph-robertson:
## Pull Request Description

Companion PR: https://github.com/NREL/resstock/pull/696

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/meta_measure.rb
+++ b/HPXMLtoOpenStudio/resources/meta_measure.rb
@@ -163,7 +163,7 @@ def run_hpxml_workflow(rundir, measures, measures_dir, debug: false, output_vars
   return { success: true, runner: runner, sim_time: sim_time }
 end
 
-def apply_measures(measures_dir, measures, runner, model, show_measure_calls = true, measure_type = 'OpenStudio::Measure::ModelMeasure', osw_out = nil, parent_measure_runner = {})
+def apply_measures(measures_dir, measures, runner, model, show_measure_calls = true, measure_type = 'OpenStudio::Measure::ModelMeasure', osw_out = nil)
   if not osw_out.nil?
     # Create a workflow based on the measures we're going to call. Convenient for debugging.
     workflowJSON = OpenStudio::WorkflowJSON.new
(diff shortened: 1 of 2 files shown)


HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -163,7 +163,29 @@ def run_hpxml_workflow(rundir, measures, measures_dir, debug: false, output_vars
...
 end
 
-def apply_measures(measures_dir, measures, runner, model, show_measure_calls = true, measure_type = 'OpenStudio::Measure::ModelMeasure')
+def apply_measures(measures_dir, measures, runner, model, show_measure_calls = true, measure_type = 'OpenStudio::Measure::ModelMeasure', osw_out = nil)
+  if not osw_out.nil?
+    # Create a workflow based on the measures we're going to call. Convenient for debugging.
+    workflowJSON = OpenStudio::WorkflowJSON.new
+    workflowJSON.setOswPath(File.expand_path("../#{osw_out}"))
+    workflowJSON.addMeasurePath('measures')
+    workflowJSON.addMeasurePath('resources/hpxml-measures')
```
joseph-robertson, line 172: @shorowit Could/should we make these paths into arguments?

HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -332,6 +367,11 @@ def run_measure(model, measure, argument_map, runner)
       runner.registerFinalCondition(result_child.finalCondition.get.logMessage)
     end
 
+    # re-register runner child registered values on the parent runner
+    result_child.stepValues.each do |step_value|
+      runner.registerValue(step_value.name, get_value_from_workflow_step_value(step_value))
+    end
+
```
joseph-robertson, line 374: @shorowit This is needed for ResStockArguments to work. It doesn't break anything in the OS-HPXML, but should we make it optional anyway?

HPXMLtoOpenStudio/resources/meta_measure.rb
shorowit: I don't think it needs to be since ResStock is the only consumer of this code. If we ever have another consumer in the future that wants different paths, we can consider adding an argument at that time.

HPXMLtoOpenStudio/resources/meta_measure.rb
shorowit: No, I don't think it needs to be optional.

shorowit (approved):

PR #845, opened 2021-09-01
Water Heater GEB

### LINKED ISSUE #635: Water heater GEB
Add new GEB features such as:

- Allowing tanks to use either a mixed or stratified tank model (additional user argument)
- Allowing setpoint to follow an hourly (8760) schedule rather than only fixed values.
- Allowing HPWH operating mode to be scheduled (either "HP Only" or "standard"). This allows the elements to be disabled during peak periods.
- Adding unmet shower (an unmet load metric for WHs) calculations to quantify any unmet WH loads.

joseph-robertson:
## Pull Request Description

Closes #635.

Adds `BuildResidentialHPXML` arguments:
* `water_heater_tank_model_type` (optional)
* `water_heater_operating_mode` (optional)

Adds 5 test xml files:
* base-dhw-tank-model-type-stratified.xml (`water_heater_tank_model_type=stratified`)
* base-dhw-tank-heat-pump-operating-mode-heat-pump-only.xml (`water_heater_operating_mode=heat pump only`)
* base-dhw-tank-detailed-setpoints.xml (`schedules_filepaths=water-heater-setpoints.csv`)
* base-dhw-tankless-detailed-setpoints.xml ('schedules_filepaths=water-heater-setpoints.csv`)
* base-dhw-tank-heat-pump-schedules.xml (`schedules_filepaths=water-heater-setpoints.csv, water-heater-operating-modes.csv`)
* base-dhw-tank-model-type-stratified-detailed-occupancy-stochastic.xml (`water_heater_tank_model_type=stratified, schedules_filepaths=occupancy-stochastic.csv`)

Also updates base-schedules-detailed-all-10-mins.csv to include `water-heater-setpoints-10-mins.csv`.

TODO:
- [x] Handle warnings:
  - [x] `Recovery Efficiency and Energy Factor could not be calculated during the test for standard ratings` for stratified tank model type. (I see this in resstock, as well.)
  - [x] `Water heater tank set point temperature is greater than or equal to the cut-in temperature of the heat pump water heater.` for heat pump water heaters. Update: I get this for base-dhw-tank-heat-pump-outside.xml.
- [x] Use `control_logic=AOSmith` for:
  - [x] `setup_hpwh_stratified_tank`
  - [x] `setup_hpwh_wrapped_condenser`
  - [x] `add_hpwh_control_program`
- [x] Update `add_hpwh_control_program` to use:
  - [x] `op_mode`
  - [x] `op_mode_schedule_file`
- [x] Implement `offset` argument in `create_schedule_file` (NA, dealt with through EMS)
- [ ] Think about "unmet showers" (To be dealt with in future PR)

## Checklist

Not all may apply:


jmaguire1:
@shorowit: Wanted to discuss a part of this with you. The GE control logic is a fair bit more complicated than the AO Smith control logic, and GE left the market several years ago. I'd preserved the GE control logic for continuity in ResStock, but I'm wondering if it's something we want to continue to preserve (and default to in certain situations)? The alternative would be to just always assume an AO Smith HPWH, we have enough recent test results from the newer products that it's much easier to keep those up to date and drop the GE.

shorowit:
@jmaguire1I would be fine with just assuming AO Smith control logic.

jmaguire1:
@shorowit: OK, I'm going to just keep the AO Smith control logic going forward and remove anything that only represents the 1st gen GE products.

jmaguire1:
Since this pull request allows you to specify whether you want to use the mixed or stratified tank model, I did some runs of both models so we could compare the runtime and energy impacts of the WH model fidelity. I ran simulations of an electric resistance tank in either a low use (1 occupant) or high use (5 occupant) single family detached home at either a 10 minute or 60 minute timestep. In terms of results, I was primarily looking at overall run time, annual energy consumption, and unmet shower energy (a relatively new discomfort metric I've been pushing that's included in this PR). Here's a summary of the results:

image

I want to make a couple points here:

1. In terms of run time, the stratified tank does take longer, which isn't terribly surprising. The run time impacts are larger if you have more hot water usage since the WH needs to make a lot more control decisions under higher hot water usage, so the adaptive timestep code Noel added to the stratified tank model is subdividing the timesteps much more frequently for high usage homes. In low usage homes the runtime impact might only be 10-20%, but for high usage homes it's a pretty substantial increase.
2. Energy consumption predictions are pretty consistent between the two models, it's ~1% difference so not particularly concerned there.
3. Comfort predictions, on the other hand, are quite a bit different between the models. At a 10 minute timestep, the discomfort even in a low usage home is surprisingly high. Looking at the draw profile, there are times where the flow rate is very high and as a result the tank does get pretty substantially depleted. With the stratified tank, you have a top element that's trying to heat a smaller volume of hot water so you see faster recovery after a large draw when you're modeling with a stratified tank instead of a mixed tank. This to me seems like the biggest upside of the stratified tank model, that we're going to be able to more realistically capture discomfort for an electric resistance water heater.

shorowit (commented):
Nice work, this looks to be in good shape. I have a fair number of comments but they are all pretty minor.

Also, are there warnings given if, e.g., both detailed and simple water heater setpoints (or HPWH operating modes) are provided? I assume we should do something [like this](https://github.com/NREL/OpenStudio-HPXML/blob/3a8479e91742ca16297709eafd6794c2d76f2bf0/HPXMLtoOpenStudio/resources/hvac.rb#L776-L778) and add tests [like this](https://github.com/NREL/OpenStudio-HPXML/blob/3a8479e91742ca16297709eafd6794c2d76f2bf0/HPXMLtoOpenStudio/tests/test_validation.rb#L1058-L1060). We should also do this for HVAC setpoints if it wasn't already done.

BuildResidentialHPXML/measure.rb
```diff
@@ -1821,6 +1821,28 @@ def arguments(model)
     arg.setDescription("Requires that the dwelling unit has a #{HPXML::HVACTypeHeatPumpAirToAir}, #{HPXML::HVACTypeHeatPumpMiniSplit}, or #{HPXML::HVACTypeHeatPumpGroundToAir} heat pump or a #{HPXML::HVACTypeCentralAirConditioner} or #{HPXML::HVACTypeMiniSplitAirConditioner} air conditioner.")
     args << arg
 
+    water_heater_tank_model_type_choices = OpenStudio::StringVector.new
+    water_heater_tank_model_type_choices << Constants.Auto
+    water_heater_tank_model_type_choices << HPXML::WaterHeaterTankModelTypeMixed
+    water_heater_tank_model_type_choices << HPXML::WaterHeaterTankModelTypeStratified
```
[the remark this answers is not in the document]
joseph-robertson: Not really other than your OSW might sort of imply that this argument applies to all water heater types (and not just storage).

Changelog.md
```diff
@@ -3,6 +3,8 @@ __New Features__
 - Updates to OpenStudio 3.4.0/EnergyPlus 22.1.
 - Expanded capabilities for scheduling:
   - Allows modeling detailed HVAC setpoints via a schedule CSV file.
+  - Allows modeling detailed tank and heat pump water heater setpoints via a schedule CSV file.
```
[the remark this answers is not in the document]
joseph-robertson: Doesn't apply for tankless water heaters.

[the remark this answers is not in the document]
shorowit: Are you confusing this with tank mode? Surely you can use detailed water heater setpoints for any type of water heater.

BuildResidentialHPXML/measure.rb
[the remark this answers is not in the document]
shorowit: I see. That's why making it optional also makes the most sense.

Changelog.md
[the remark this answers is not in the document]
joseph-robertson: No, I'm not. We're passing `@schedules-file` only into `apply_tank` and `apply_heat_pump`: https://github.com/NREL/OpenStudio-HPXML/blob/wh-geb/HPXMLtoOpenStudio/measure.rb#L1400-L1406.

I also confirmed this with @jmaguire1: "there is really no reason to allow [scheduled setpoints] for [tankless water heaters], you don't get any load shifting benefit from scheduling a tankless."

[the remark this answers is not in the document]
shorowit: I see. I understand the rationale, but suggest we allow detailed setpoints for any water heater type anyway. It's simpler, it's a small effort, and it causes no harm. You never know what someone might want to use it for. 

(If there _was_ a strong justification to not allow it, then at a minimum we have to document it better because there's nothing in the documentation right now saying that it only applies to certain water heater types.)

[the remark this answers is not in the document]
joseph-robertson: For combi, too?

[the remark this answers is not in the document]
shorowit: Yes. Any water heating system that allows a [simple setpoint temperature](https://github.com/NREL/OpenStudio-HPXML/blob/f44aec2615d402c80afff028bdd0317c7362fb98/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml#L1413) should be allowed a detailed setpoint schedule.

tasks.rb
```diff
@@ -401,6 +402,8 @@ def create_hpxmls
 
   hpxml_docs = {}
   hpxmls_files.each_with_index do |(hpxml_file, orig_parent), i|
+    next if hpxml_file != 'base-dhw-tankless-detailed-setpoints.xml'
+
```
shorowit, line 406: Need to revert this

joseph-robertson:
> Nice work, this looks to be in good shape. I have a fair number of comments but they are all pretty minor.
> 
> Also, are there warnings given if, e.g., both detailed and simple water heater setpoints (or HPWH operating modes) are provided? I assume we should do something [like this](https://github.com/NREL/OpenStudio-HPXML/blob/3a8479e91742ca16297709eafd6794c2d76f2bf0/HPXMLtoOpenStudio/resources/hvac.rb#L776-L778) and add tests [like this](https://github.com/NREL/OpenStudio-HPXML/blob/3a8479e91742ca16297709eafd6794c2d76f2bf0/HPXMLtoOpenStudio/tests/test_validation.rb#L1058-L1060). We should also do this for HVAC setpoints if it wasn't already done.

I think we'll need to make changes to EPvalidator.xml. Currently, simple HVAC setpoints are required.

joseph-robertson:
There is a problem [here](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/waterheater.rb#L1640) with having only detailed water heater setpoints and not a simple one.

HPXMLtoOpenStudio/resources/hpxml_schematron/HPXMLvalidator.xml
```diff
@@ -287,7 +287,7 @@
...
+      <sch:assert role='ERROR' test='h:HeatPumpType[text()="water-to-air" or text()="water-to-water" or text()="air-to-air" or text()="air-to-water" or text()="mini-split" or text()="ground-to-air" or text()="ground-to-water" or text()="packaged terminal heat pump" or text()="water-loop-to-air" or text()="variable refrigerant flow"] or not(h:HeatPumpType)'>Expected HeatPumpType to be 'water-to-air' or 'water-to-water' or 'air-to-air' or 'air-to-water' or 'mini-split' or 'ground-to-air' or 'ground-to-water' or 'packaged terminal heat pump' or 'water-loop-to-air' or 'variable refrigerant flow'</sch:assert>
```
joseph-robertson, line 290: Not sure why this file changed.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1109,8 +1109,8 @@
...
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACControl[sum(../h:HVACPlant/*/h:FractionHeatLoadServed) > 0]'>
-      <sch:assert role='ERROR' test='count(h:SetpointTempHeatingSeason) + count(h:extension/h:WeekdaySetpointTempsHeatingSeason) = 1'>Expected 1 element(s) for xpath: SetpointTempHeatingSeason | extension/WeekdaySetpointTempsHeatingSeason</sch:assert>
-      <sch:assert role='ERROR' test='count(h:SetpointTempHeatingSeason) + count(h:extension/h:WeekendSetpointTempsHeatingSeason) = 1'>Expected 1 element(s) for xpath: SetpointTempHeatingSeason | extension/WeekendSetpointTempsHeatingSeason</sch:assert>
+      <sch:assert role='ERROR' test='count(h:SetpointTempHeatingSeason) + count(h:extension/h:WeekdaySetpointTempsHeatingSeason) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SetpointTempHeatingSeason | extension/WeekdaySetpointTempsHeatingSeason</sch:assert>
+      <sch:assert role='ERROR' test='count(h:SetpointTempHeatingSeason) + count(h:extension/h:WeekendSetpointTempsHeatingSeason) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SetpointTempHeatingSeason | extension/WeekendSetpointTempsHeatingSeason</sch:assert>
```
shorowit, line 1113: Need to update the documentation to specify the default values that will be used.

HPXMLtoOpenStudio/resources/hotwater_appliances.rb
```diff
@@ -226,7 +226,9 @@ def self.apply(model, runner, hpxml, weather, spaces, hot_water_distribution,
       t_mix = 105.0 # F, Temperature of mixed water at fixtures
       avg_setpoint_temp = 0.0 # WH Setpoint: Weighted average by fraction DHW load served
       hpxml.water_heating_systems.each do |water_heating_system|
-        avg_setpoint_temp += water_heating_system.temperature * water_heating_system.fraction_dhw_load_served
+        wh_setpoint = water_heating_system.temperature
+        wh_setpoint = Constants.WaterHeaterSetpointAverage if wh_setpoint.nil? # using detailed schedules
```
joseph-robertson, line 230: Use `Waterheater.get_default_hot_water_temperature(eri_version)` instead?

shorowit: I was going to leave that comment. Yes, let's remove the new Constant and reference the default value instead.

HPXMLtoOpenStudio/resources/hpxml_schematron/HPXMLvalidator.xml
shorowit: It's safe to ignore.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1220,6 +1222,12 @@ def validate_schedules(year:)
           end
         end
 
+        if only_zeros_and_ones[col_name]
+          if values.any? { |v| v > 0 && v < 1 }
```
shorowit, line 1226: Doesn't this not catch values like -0.5 or 2?

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1487,10 +1517,22 @@ def min_value_zero
...
     end
     return min_value_zero
   end
+
+  def only_zeros_and_ones
+    only_zeros_and_ones = {}
+    column_names = SchedulesFile.ColumnNames
+    column_names.each do |column_name|
+      only_zeros_and_ones[column_name] = false
+      if SchedulesFile.OperatingModeColumnNames.include?(column_name)
```
shorowit, line 1532: Should the `vacancy` column be included here? The documentation says only 0s or 1s are allowed for it too.

workflow/tests/base_results/results.csv
```diff
@@ -61,7 +61,7 @@ base-dhw-combi-tankless-outside.xml,50.86,50.86,21.419,21.419,29.441,0.0,0.0,0.0
...
+base-dhw-desuperheater-hpwh.xml,53.481,53.481,28.795,28.795,24.686,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.407,3.629,0.658,2.824,0.0,0.0,4.51,0.0,0.334,0.0,0.0,0.0,0.0,2.22,0.0,0.0,0.319,0.365,1.513,1.528,0.0,2.114,8.374,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,24.686,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,23.117,11.051,9.25,1.807,2.503,0.0,0.0,0.0,1878.1,3056.1,22.763,18.475,0.0,3.542,3.643,0.513,7.11,0.629,-2.118,0.0,0.0,6.682,-0.1,4.798,0.0,0.727,0.0,5.241,-7.901,0.0,0.045,-0.376,-0.039,3.064,-0.006,10.029,0.0,0.0,-5.4,-0.042,-1.072,-6.622,-0.148,0.0,2.744,9.227,1354.7,997.5,11327.9,2599.4,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
```
shorowit, line 64: Why did some existing files change slightly?

joseph-robertson: I removed a few uses of `round()`. I think maybe because of that.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
joseph-robertson: I took a shot at it.

shorowit (approved):
Looks good to me. Thanks @joseph-robertson and @jmaguire1!

PR #846, opened 2021-09-02
Replaces Unmet Load outputs with Unmet Hours

### LINKED ISSUE #840: Reconsider unmet load output
Reporting unmet load in units of MBtu (instead of hours) is really convenient and works well in the vast majority of situations. It allows a user to compare the unmet load to the total load to understand, e.g., if an HVAC system is undersized.

Unfortunately, there are scenarios where the implementation (reporting the energy use provided by a backup ideal air system) can affect the desired behavior of the model and the simulation results. We may need to revert to using a more conventional metric like hours (or degree-hours?).

shorowit:
## Pull Request Description

Closes #840.

## Checklist

Not all may apply:


shorowit:
FYI @aspeake @joseph-robertson

shorowit:
@yzhou601 Can you investigate the sum of component loads vs total reported load for the base-hvac-undersized.xml file? (Might also move ducts into living space.)

docs/source/workflow_outputs.rst
```diff
@@ -162,29 +162,26 @@ Current annual building loads are listed below.
    Load: Hot Water: Solar Thermal (MBtu) Load served by the solar thermal system.
    ===================================== ==================================================================
 
-Annual Unmet Building Loads
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Note that heating and cooling loads represent the delivered energy by the HVAC system; if the HVAC system is significantly undersized, there will be unmet load not reflected by these values.
```
yzhou601, line 165: I noticed that heating and cooling peak loads are also reduced due to the unmet loads. I haven't kept track of how peak loads are calculated and reported, but they also seem to be related to the hvac delivered energy instead of load calculation.

yzhou601 (approved):
Changes look good to me. I've run tests with duct placing to the living space. The results are:
image
Master branch shows consistent loads between undersize and base test files, unmet hour branch shows reduction in load reporting caused by unmet loads. The energy increase on unmet hour branch is expected. 
I have run 3 undersized test files on unmet_hours branch, they all show consistency between total loads and individual load sums. I have kind of persuaded myself, this can be caused by zone air temperature was overall kept in a reasonable range, so energy is still balanced.

docs/source/workflow_outputs.rst
shorowit: Thanks for looking over this PR in such detail!

Good point, I will update the documentation for Peak Loads too. This also makes me think that it's probably a good idea to rename these outputs from, e.g., "Load: Heating" to "Load: Heating: Delivered". We already have a "Load: Hot Water: Delivered". (This reminds me that we should also have some language in the documentation about unmet hot water loads.) I hate to make breaking changes to output names, but it seems important. I will make these changes.

Issue #811, opened 2021-07-27
Alternative EC_acj implementation for water heating


shorowit:
In order to apply the hot water energy use adjustment from ANSI 301-2019 (Equation 4.2.2.5.2.11.2), we currently use [an EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/39a3a467649cc614982b73a6104c0d10a52e2f13/HPXMLtoOpenStudio/resources/waterheater.rb#L1263-L1370) to increase/decrease water heating energy consumption for every type of water heater object, as well as [reporting measure code](https://github.com/NREL/OpenStudio-HPXML/blob/39a3a467649cc614982b73a6104c0d10a52e2f13/SimulationOutputReport/measure.rb#L754-L772) to shift the energy use from the Equipment end use to the Water Heating end use. This all works, but A) it adds a lot of complication, B) it can cause raw EnergyPlus outputs to be misinterpreted, C) it causes a 1-hour lag due to use of EMS, and D) probably some other things (e.g., runtime performance?).

After discussing with @jmaguire1 and @yzhou601, we've come up with two alternative implementations:
1. Adjust the hot water draw flow rates and tank standby losses to produce the appropriate water heating energy uses. I verified this works perfectly for a conventional storage tank water heater. Unfortunately, we would still need a small EMS program (that actuates an `OtherEquipment` object) to counterbalance the change to standby losses so that the water heater's internal gains to the zone are not changed.
2. Adjust the water heater's efficiencies/COPs to produce the appropriate water heating energy uses. I verified this works perfectly for a conventional storage tank water heater. Unfortunately there are two issues: A) EnergyPlus would need to be updated to allow burner efficiencies greater than 1 (and presumably throw a warning if so), and B) our combi boiler model would need to be separated into separate boiler and water heater objects, since there is no way to use different combi boiler efficiencies for space heating vs water heating. That said, there could be good reasons to split our combi boiler model, such as workflow code simplification and robustness of results.

shorowit:
@yzhou601 will start looking at splitting our combi boiler model.

PR #868, opened 2021-09-30
Two boiler approach for combi system

yzhou601:
## Pull Request Description

Addresses https://github.com/NREL/OpenStudio-HPXML/issues/811

## Checklist

Not all may apply:


shorowit:
@yzhou601 Ping. Pull in the latest code.

shorowit:
With this approach, can we remove the IsCombiBoiler flag ([here](https://github.com/NREL/OpenStudio-HPXML/blob/0084f11f0018b8003f43b218d3ef47bd20646c37/HPXMLtoOpenStudio/resources/waterheater.rb#L140) and [here](https://github.com/NREL/OpenStudio-HPXML/blob/0084f11f0018b8003f43b218d3ef47bd20646c37/HPXMLtoOpenStudio/resources/waterheater.rb#L1589-L1590)) and all references?

shorowit:
And does this allow us to remove the [combi boiler EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/328e761256a55630eb433e920d4739b4cc883b6f/HPXMLtoOpenStudio/resources/waterheater.rb#L221)? 🙏

yzhou601:
> And does this allow us to remove the [combi boiler EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/328e761256a55630eb433e920d4739b4cc883b6f/HPXMLtoOpenStudio/resources/waterheater.rb#L221)? 🙏

If I remembered correctly, E+ is not modeling various tank source water flow , so we use this EMS program to feed in proper amount of hot water. This will still be the case with current proposed changes (remove unnecessary objects and plant loops).

shorowit:
I haven't fully thought through the idea, but wanted to write it down before I forget it. Is it worth modeling the piece of equipment that serves the hot water as an actual water heater object in EnergyPlus instead of a boiler? I assume it would eliminate the need for the combi boiler EMS program as well as address the need for an [Energy Factor for ERI calculations](https://github.com/NREL/OpenStudio-HPXML/blob/d203d4742f105090d3ddbfe697cae5f831a68559/ReportSimulationOutput/measure.rb#L1190-L1196).

yzhou601:
> I haven't fully thought through the idea, but wanted to write it down before I forget it. Is it worth modeling the piece of equipment that serves the hot water as an actual water heater object in EnergyPlus instead of a boiler? I assume it would eliminate the need for the combi boiler EMS program as well as address the need for an [Energy Factor for ERI calculations](https://github.com/NREL/OpenStudio-HPXML/blob/d203d4742f105090d3ddbfe697cae5f831a68559/ReportSimulationOutput/measure.rb#L1190-L1196).

I didn't fully understand the approach of `modeling the piece of equipment that serves the hot water as an actual water heater object in EnergyPlus instead of a boiler`. Currently there is a water heater being modeled, but its source side is connected to a boiler, so it's basically just a tank, do you mean model it as a water heater (add heating element)?

shorowit:
@yzhou601 Will this approach (or the boiler + water heater approach) allow us to remove [this code](https://github.com/NREL/OpenStudio-HPXML/blob/2224e6cfcf913ac7815b8f4688b250ee5c86f924/HPXMLtoOpenStudio/resources/waterheater.rb#L142-L144)? I assume we would see some runtime improvement if we can remove that code.

shorowit:
The results here are looking really good. I think this could be merged in if we:
1. Do a couple before vs after comparison tests with stochastic occupancy schedules.
2. Address [this comment](https://github.com/NREL/OpenStudio-HPXML/pull/868#issuecomment-1031744194), assuming it is acceptable.

yzhou601:
@shorowit stochastic schedule tests were finished. I used the same [stochastic schedule](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/schedule_files/stochastic.csv) to ensure consistency between before and after comparison. The test files run:
image
See below the results. 
image
A few takeaways:
1. The results of water heating and space heating natural gas consumptions are promising. 
2. As expected, the pump powers are underestimated with the new approach, it existed in both stochastic testing and current ci testing. 
3. The stochastic schedule testing reveals a difference in hot water consumption (gal). I looked at the time step tank temperature outputs, the new approach is showing a slightly higher average temperature 51.622C (before) vs 51.808 (after). Neither approach showed unreasonable temperature throughout the year (all within 51c - 52.5c), so I am not too worrying about this.

shorowit:
@yzhou601 Thanks! That all sounds great. The pump energy use is so small that it doesn't concern me. I've pulled this branch into OS-ERI just to run a few more tests.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -188,50 +189,32 @@ def self.apply_combi(model, runner, loc_space, loc_schedule, water_heating_syste
...
-    hx_stp_sch.setName("#{obj_name_combi} HX Spt")
+    # Create setpoint schedule to specify source side temperature
+    source_stp_sch = OpenStudio::Model::ScheduleConstant.new(model)
+    source_stp_sch.setName("#{obj_name_combi} Source Spt")
     boiler_spt_mngr = model.getSetpointManagerScheduleds.select { |spt_mngr| spt_mngr.setpointNode.get == boiler_plant_loop.loopTemperatureSetpointNode }[0]
-    boiler_spt = boiler_spt_mngr.to_SetpointManagerScheduled.get.schedule.to_ScheduleConstant.get.value
-    hx_temp = (UnitConversions.convert(water_heating_system.temperature, 'F', 'C') + deadband(water_heating_system.water_heater_type) / 2.0 + boiler_spt) / 2.0 # tank source side inlet temperature, degree C
-    hx_stp_sch.setValue(hx_temp)
+    boiler_heating_spt = boiler_spt_mngr.to_SetpointManagerScheduled.get.schedule.to_ScheduleConstant.get.value
+    source_temp = (UnitConversions.convert(water_heating_system.temperature, 'F', 'C') + deadband(water_heating_system.water_heater_type) / 2.0 + boiler_heating_spt) / 2.0 # tank source side inlet temperature, degree C
```
shorowit, line 197: @yzhou601 We are averaging the water heater setpoint and the boiler setpoint? Can you explain why we are doing that?

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -979,16 +962,26 @@ def self.add_hpwh_control_program(model, obj_name_hpwh, amb_temp_sensor, hpwh_bo
...
+      plant_loop_hw = plant_loop.clone(model).to_PlantLoop.get
+      # pump power is zero, fixme: is this expected? what about no heating scenarios when only water heating boiler is running but no power consumption is calculated?
+      plant_loop_hw.supplyComponents.each do |comp|
+        if comp.to_BoilerHotWater.is_initialized
+          boiler_hw = comp.to_BoilerHotWater.get
+        end
+        next unless comp.to_PumpVariableSpeed.is_initialized
+
+        pump_hw = comp.to_PumpVariableSpeed.get
+        pump_hw.setRatedPowerConsumption(0.0)
```
shorowit, line 981: Can you explain this too? I don't understand the reason for this code nor your comment above it.

shorowit (commented):
This is looking really close! A couple questions.

HPXMLtoOpenStudio/resources/waterheater.rb
yzhou601: This is the assumption we make for water heating boiler that pump power is not set (0 for now). This will underestimate pump energy consumption especially when in summer if only water heating is needed.

HPXMLtoOpenStudio/resources/waterheater.rb
yzhou601: This is a good question, this is the tank source water temperature. Ideally it should be equal to boiler setpoint, but in old approach we're using heat exchanger object, and setting that temperature will cause problems. Let me try if it works in the new two boiler approach.

yzhou601: Using boiler temperature gave a similar result. Except a higher bias on gallons for combi tankless.
image
But hourly tank temperature still is not bad.
image

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: There shouldn't be any pump energy associated with water heating, the water coming from the mains is pressurized and you just need to open the tap for it to come out. There is pump energy associated with space heating because it's a closed loop system and the pump is used to circulate the hot water to your radiators/baseboards.

shorowit (approved):
This can drop in once the CI comes back happy. Thanks @yzhou601!

PR #873, opened 2021-10-07
OS 3.3.0-Infiltration reporting

yzhou601:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


yzhou601:
@shorowit CI shows identical results with new infiltration output variables. Could you review this PR?

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2208,14 +2208,27 @@ def self.add_component_loads_output(runner, model, living_zone, liv_load_sensors
...
+        next unless i.space.get.thermalZone.get.name.to_s == living_zone.name.to_s
+
+        { 'Infiltration Sensible Heat Gain Energy' => ' - ',
+          'Infiltration Sensible Heat Loss Energy' => ' + ' }.each do |var, sign|
+          name = var.split(' ')[-2].downcase
+          sensor_name = prefix.gsub(' ', '_') + '_' + name
+          airflow_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, var)
+          airflow_sensor.setName(sensor_name)
+          airflow_sensor.setKeyName(i.name.to_s)
+          sensor_s += (sign + airflow_sensor.name.get)
```
yzhou601, line 2227: Did string manipulation here instead of passing sensors to avoid complex struct to differentiate infil/natvent/whf and gain/loss sensors. Let me know if there's a better method.

shorowit: I have to admit that it definitely feels more complicated than I expected and it's pretty inconsistent with how other code is written (for e.g., ducts and mechvent). How about:

```
infils_sensors = []
natvents_sensors = []
whfs_sensors = []
{ Constants.ObjectNameInfiltration => infils_sensors,
  Constants.ObjectNameNaturalVentilation => natvents_sensors,
  Constants.ObjectNameWholeHouseFan => whfs_sensors }.each do |prefix, all_sensors|
  ...
  # Follow same approach as e.g. https://github.com/NREL/OpenStudio-HPXML/blob/bc22b8b1d7f3c9eafd449ffa728c5971b99eb0e3/HPXMLtoOpenStudio/measure.rb#L2296-L2303
end
```

And then you can just drop these sensors into the code here: https://github.com/NREL/OpenStudio-HPXML/blob/bc22b8b1d7f3c9eafd449ffa728c5971b99eb0e3/HPXMLtoOpenStudio/measure.rb#L2437-L2438

shorowit: So for ducts and mechvent, we have `conv` and `rad` terms while you have `gain` and `loss` terms instead. Seems like you could just handle whether it's + or - by looking at the name of the sensor inside here https://github.com/NREL/OpenStudio-HPXML/blob/bc22b8b1d7f3c9eafd449ffa728c5971b99eb0e3/HPXMLtoOpenStudio/measure.rb#L2442-L2444

yzhou601: I'll try this

yzhou601:
@shorowit codes are now structured more consistent with other parts.

PR #910, opened 2021-11-10
Makes BuildResidentialHPXML measure deterministic, try 2

joseph-robertson:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


shorowit:
What if, instead, we assign an incrementing index to each surface's `additionalProperties` during creation of surfaces, and then use that for sorting the surfaces during `model.getSurfaces`? That should be cleaner and simpler, though it requires our creation of surfaces to be deterministic (which it should be, with the possible exception of `fromFloorPrint`).

PR #913, opened 2021-11-15
Return from ReportSimOutput `energyPlusOutputRequests` if runner is halted

joseph-robertson:
## Pull Request Description

ResStock can halt workflow from both `BuildExistingModel` and `ApplyUpgrade`.

## Checklist

Not all may apply:

### DIFF
--- a/ReportSimulationOutput/measure.rb
+++ b/ReportSimulationOutput/measure.rb
@@ -137,6 +137,7 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     super(runner, user_arguments)
 
     result = OpenStudio::IdfObjectVector.new
+    return result if runner.halted
 
     model = runner.lastOpenStudioModel
     if model.empty?
--- a/ReportSimulationOutput/measure.xml
+++ b/ReportSimulationOutput/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.0</schema_version>
   <name>report_simulation_output</name>
   <uid>df9d170c-c21a-4130-866d-0d46b06073fd</uid>
-  <version_id>82fbc69f-20ec-4aff-8eb4-c35bb68bc873</version_id>
-  <version_modified>20211107T033242Z</version_modified>
+  <version_id>dea2f975-2214-40e5-b60b-2954b392df2b</version_id>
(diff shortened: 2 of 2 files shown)


ReportSimulationOutput/measure.rb
```diff
@@ -137,6 +137,7 @@ def energyPlusOutputRequests(runner, user_arguments)
     super(runner, user_arguments)
 
     result = OpenStudio::IdfObjectVector.new
+    return result if runner.halted
```
shorowit, line 140: This doesn't seem like the right place for this code. It shouldn't be a measure's responsibility to check for runner.halted; have you ever seen any other measure with this check here?

If I had to guess, I assume the check should be inside meta_measure.rb? Maybe in [here](https://github.com/NREL/OpenStudio-HPXML/blob/27ca8d6323aef936ae312d64179cc61fe8c70128/HPXMLtoOpenStudio/resources/meta_measure.rb#L189) and [here](https://github.com/NREL/OpenStudio-HPXML/blob/27ca8d6323aef936ae312d64179cc61fe8c70128/HPXMLtoOpenStudio/resources/meta_measure.rb#L213)?

joseph-robertson: SimulationOutputReport on resstock develop has this check: https://github.com/NREL/resstock/blob/develop/measures/SimulationOutputReport/measure.rb#L54.

For bsb where the workflow is generated up front, I'm not sure how we would remove SimulationOutputReport from the workflow after the workflow has already begun? In reality, this may be an issue in the OS workflow gem (`run` methods get skipped but `energyPlusOutputRequests` methods do not.

shorowit: Okay, I see the issue you're facing. I guess there's no harm in adding this code to the measure, but yes, it seems like a bug should be filed against the OS workflow gem.

shorowit (approved):

ReportSimulationOutput/measure.rb
joseph-robertson: I created an issue on OpenStudio-workflow-gem: https://github.com/NREL/OpenStudio-workflow-gem/issues/143

PR #927, opened 2021-11-30
Allow schedule CSV columns to be optional

joseph-robertson:
## Pull Request Description

Checklist:
- [x] Remove error checks from EPvalidator.xml for when both weekday/weekend fractions and monthly multipliers are found in the HPXML file along with a `SchedulesFilePath`
- [x] Allow schedule CSV file to have a subset of all the possible columns
- [x] Pull `validate_schedules` out of init method so that we can pass `schedules_file` into the defaulting method
- [x] Throw warning when both weekday/weekend fractions and monthly multipliers are found in the HPXML file along with a `SchedulesFilePath`; in these instances, ignore the weekday/weekend fractions and monthly multipliers

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/EPvalidator.xml
```diff
@@ -166,9 +166,10 @@
...
-      <sch:assert role='ERROR' test='count(h:extension/h:WeekendScheduleFractions) + count(/h:HPXML/h:SoftwareInfo/h:extension/h:SchedulesFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/WeekendScheduleFractions | /HPXML/SoftwareInfo/extension/SchedulesFilePath</sch:assert>
-      <sch:assert role='ERROR' test='count(h:extension/h:MonthlyScheduleMultipliers) + count(/h:HPXML/h:SoftwareInfo/h:extension/h:SchedulesFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/MonthlyScheduleMultipliers | /HPXML/SoftwareInfo/extension/SchedulesFilePath</sch:assert>
+      <!-- Warnings -->
+      <sch:assert role='WARN' test='count(h:extension/h:WeekdayScheduleFractions) + count(/h:HPXML/h:SoftwareInfo/h:extension/h:SchedulesFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/WeekdayScheduleFractions | /HPXML/SoftwareInfo/extension/SchedulesFilePath</sch:assert>
```
shorowit, line 170: I think we should:
1. Move the warnings into the HPXMLtoOS measure (i.e., remove them from EPvalidator).
2. Only issue the warnings if the relevant column exists in the schedule CSV (as opposed to these warnings that would appear even if the user correctly, e.g., had no `occupants` column in the CSV).
3. Switch to more user-friendly warning messages.

joseph-robertson: On the second item, do we want to lift the requirement that all columns in `Constants.ScheduleColNames.keys` are present in the CSV?

shorowit: Yes

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -526,16 +526,20 @@ def self.apply_occupants(model, hpxml, num_occ, cfa, space, schedules_file)
     occ_lost = 1 - occ_lat - occ_conv - occ_rad
 
     # Create schedule
+    people_sch = false
```
shorowit, line 529: Please don't mix data types. This will result in it either being a boolean (false) or an object, which is confusing.

Use `people_sch = nil` and then use `if not people_sch.nil?` below. (Though in ruby, I don't think you even need to set `people_sch` to a value, it will default to nil.)

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -526,16 +526,20 @@ def self.apply_occupants(model, hpxml, num_occ, cfa, space, schedules_file)
...
+    if not people_sch
       weekday_sch = hpxml.building_occupancy.weekday_fractions.split(',').map(&:to_f)
       weekday_sch = weekday_sch.map { |v| v / weekday_sch.max }.join(',')
       weekend_sch = hpxml.building_occupancy.weekend_fractions.split(',').map(&:to_f)
       weekend_sch = weekend_sch.map { |v| v / weekend_sch.max }.join(',')
       monthly_sch = hpxml.building_occupancy.monthly_multipliers
       people_sch = MonthWeekdayWeekendSchedule.new(model, Constants.ObjectNameOccupants + ' schedule', weekday_sch, weekend_sch, monthly_sch, Constants.ScheduleTypeLimitsFraction)
       people_sch = people_sch.schedule
+    elsif !hpxml.building_occupancy.weekday_fractions.nil? && !hpxml.building_occupancy.weekend_fractions.nil? && !hpxml.building_occupancy.monthly_multipliers.nil?
+      puts "Warning: Using 'occupants' schedule even though weekday/weekend fractions and monthly multipliers are provided."
```
shorowit, line 542: You need to use `runner.registerWarning`, not `puts`, otherwise the warning won't show up as appropriately. This also suggests to me that you don't have a test to check for the warning because the test would have failed. Can you add a `test_measure_warning_messages` in test_validation.rb?

I also suggest making the warnings more nuanced. How about:

```
  else
      runner.registerWarning("Both 'occupants' schedule file and weekday fractions provided; the latter will be ignore.") if !hpxml.building_occupancy.weekday_fractions.nil?
      runner.registerWarning("Both 'occupants' schedule file and weekend fractions provided; the latter will be ignore.") if !hpxml.building_occupancy.weekend_fractions.nil?
      runner.registerWarning("Both 'occupants' schedule file and monthly multipliers provided; the latter will be ignore.") if !hpxml.building_occupancy.monthly_multipliers.nil?
```

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -211,20 +211,26 @@ def self.apply_neighbor_buildings(hpxml)
...
       hpxml.building_occupancy.number_of_residents = Geometry.get_occupancy_default_num(nbeds)
       hpxml.building_occupancy.number_of_residents_isdefaulted = true
     end
-    if hpxml.building_occupancy.weekday_fractions.nil? && hpxml.header.schedules_filepath.nil?
+    schedules_file_includes_occupants = false
+    if not schedules_file.nil?
+      if schedules_file.schedules.keys.include?('occupants')
+        schedules_file_includes_occupants = true
+      end
+    end
```
shorowit, line 224: Make this a method with a column_name argument so that you don't copy/paste 5 lines of code for every column.

Also, should we convert the column names to constants and update any references?

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1203,7 +1205,7 @@ def create_schedule_file(col_name:,
     end
 
     if @schedules[col_name].nil?
-      @runner.registerError("Could not find the '#{col_name}' schedule.")
+      puts "Could not find the '#{col_name}' schedule."
```
shorowit, line 1208: If we hit this, does it suggest that we have a bug in our code? If so, just convert to `fail` so that we see the bug and fix it.

shorowit (commented):
I like the direction you're heading 👍. I do have a few comments on the implementation.

BuildResidentialScheduleFile/tests/build_residential_schedule_file_test.rb
```diff
@@ -43,7 +43,8 @@ def test_smooth
     assert(!info_msgs.any? { |info_msg| info_msg.include?('VacancyPeriod') })
     assert(!warn_msgs.any? { |warn_msg| warn_msg.include?('Overwriting') })
 
-    sf = SchedulesFile.new(model: model, year: 2007, schedules_path: @args_hash['output_csv_path'])
+    sf = SchedulesFile.new(model: model, schedules_path: @args_hash['output_csv_path'])
+    sf.validate_schedules(year: 2007)
```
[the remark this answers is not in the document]
joseph-robertson: We want to pass an instance of `SchedulesFile` into `hpxml_defaults`, but the year must first be available.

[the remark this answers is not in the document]
shorowit: Ah, I see what you're saying. Okay, what you have is fine then.

HPXMLtoOpenStudio/resources/schedules.rb
joseph-robertson: No. I changed this because you can now specify a `SchedulesFilePath` but you don't need to have every column in it.

joseph-robertson: Perhaps we issue a warning?

shorowit: I assumed that, going forward, this method should only be called on columns that exist -- and if it accidentally got called on a column that didn't exist, that would represent a bug in our code. But you're saying that's not the case?

joseph-robertson: So should we check for the column's existence outside of the `create_schedule_file` call? Or how else do we know whether we can call `create_schedule_file` for a given column?

shorowit: Yes, I would assume so. We'll need to know whether to call this method and create a schedule file, or call other methods dealing with weekday/weekend/multiplier, no?

BuildResidentialScheduleFile/tests/build_residential_schedule_file_test.rb
```diff
@@ -43,9 +43,10 @@ def test_smooth
...
 
-    assert_in_epsilon(6020, sf.annual_equivalent_full_load_hrs(col_name: 'occupants', schedules: sf.tmp_schedules), 0.1)
+    assert_in_epsilon(6020, sf.annual_equivalent_full_load_hrs(col_name: Constants.Occupants, schedules: sf.tmp_schedules), 0.1)
     assert_in_epsilon(3321, sf.annual_equivalent_full_load_hrs(col_name: 'lighting_interior', schedules: sf.tmp_schedules), 0.1)
     assert_in_epsilon(2763, sf.annual_equivalent_full_load_hrs(col_name: 'lighting_exterior', schedules: sf.tmp_schedules), 0.1)
     assert_in_epsilon(2763, sf.annual_equivalent_full_load_hrs(col_name: 'lighting_garage', schedules: sf.tmp_schedules), 0.1)
```
shorowit, line 52: Looks like more columns need to be changed to constants in this file.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -846,7 +846,7 @@ def read_monthly_shift_minutes(resources_path:, daytype:)
   end
 
   def read_appliance_power_dist(resources_path:)
-    activity_names = ['clothes_washer', 'dishwasher', 'clothes_dryer', 'cooking']
+    activity_names = [Constants.ClothesWasher, Constants.Dishwasher, Constants.ClothesDryer, 'cooking']
```
shorowit, line 849: Why are some of these constants and some aren't?

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -880,7 +880,7 @@ def sample_appliance_duration_power(prng, power_dist_map, appliance_name)
   end
 
   def read_activity_cluster_size_probs(resources_path:)
-    activity_names = ['hot_water_clothes_washer', 'hot_water_dishwasher', 'shower']
+    activity_names = [Constants.HotWaterClothesWasher, Constants.HotWaterDishwasher, 'shower']
```
shorowit, line 883: Here too.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -892,7 +892,7 @@ def read_activity_cluster_size_probs(resources_path:)
   end
 
   def read_event_duration_probs(resources_path:)
-    activity_names = ['hot_water_clothes_washer', 'hot_water_dishwasher', 'shower']
+    activity_names = [Constants.HotWaterClothesWasher, Constants.HotWaterDishwasher, 'shower']
```
shorowit, line 895: And here.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -753,34 +753,34 @@ def create_stochastic_schedules(args:)
...
-    @schedules['dishwasher'] = dw_power_sch.map { |power| power / dw_peak_power }
+    @schedules[Constants.Dishwasher] = dw_power_sch.map { |power| power / dw_peak_power }
 
-    @schedules['occupants'] = away_schedule.map { |i| 1.0 - i }
+    @schedules[Constants.Occupants] = away_schedule.map { |i| 1.0 - i }
 
-    @schedules['hot_water_fixtures'] = [@schedules['showers'], @schedules['sinks'], @schedules['baths']].transpose.map { |flow| flow.reduce(:+) }
-    fixtures_peak_flow = @schedules['hot_water_fixtures'].max
-    @schedules['hot_water_fixtures'] = @schedules['hot_water_fixtures'].map { |flow| flow / fixtures_peak_flow }
+    @schedules[Constants.HotWaterFixtures] = [@schedules['showers'], @schedules['sinks'], @schedules['baths']].transpose.map { |flow| flow.reduce(:+) }
```
shorowit, line 781: And here.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -211,20 +211,21 @@ def self.apply_neighbor_buildings(hpxml)
...
       hpxml.building_occupancy.weekday_fractions = Schedule.OccupantsWeekdayFractions
       hpxml.building_occupancy.weekday_fractions_isdefaulted = true
     end
-    if hpxml.building_occupancy.weekend_fractions.nil? && hpxml.header.schedules_filepath.nil?
+    if hpxml.building_occupancy.weekend_fractions.nil? && !schedules_file_includes_occupants
       hpxml.building_occupancy.weekend_fractions = Schedule.OccupantsWeekendFractions
       hpxml.building_occupancy.weekend_fractions_isdefaulted = true
     end
-    if hpxml.building_occupancy.monthly_multipliers.nil? && hpxml.header.schedules_filepath.nil?
+    if hpxml.building_occupancy.monthly_multipliers.nil? && !schedules_file_includes_occupants
```
shorowit, line 228: Aren't there a whole bunch more in this file that need to be updated?

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -992,7 +950,60 @@ def test_measure_error_messages
...
+                                                                                  "Both 'lighting_exterior' schedule file and weekday fractions provided; the latter will be ignored.",
+                                                                                  "Both 'lighting_exterior' schedule file and weekend fractions provided; the latter will be ignored.",
+                                                                                  "Both 'lighting_exterior' schedule file and monthly multipliers provided; the latter will be ignored."] }
+
+    all_expected_warnings.each_with_index do |(warning_case, expected_warnings), i|
+      puts "[#{i + 1}/#{all_expected_warnings.size}] Testing #{warning_case}..."
+      # Create HPXML object
+      if ['schedule-file-and-weekday-weekend-multipliers'].include? warning_case
+        hpxml = HPXML.new(hpxml_path: File.join(@sample_files_path, 'base-schedules-simple.xml'))
+        hpxml.header.schedules_filepath = 'HPXMLtoOpenStudio/resources/schedule_files/smooth.csv'
```
shorowit, line 998: Nice test.

BuildResidentialScheduleFile/resources/schedules.rb
joseph-robertson: Good catch. These variables are internal to the algorithm (correspond to schedule resource files) and shouldn't be touched.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
joseph-robertson: Oops. I think I forgot to finish the "first cut".

How do we handle interior lighting? It doesn't appear to get defaulted in this file.

joseph-robertson: Looks like it pulls it from the epw. So, nothing to do I suppose.

shorowit (approved):

shorowit:
Thanks @joseph-robertson!

PR #938, opened 2021-12-15
Allow applying defaults to the HXPML file produced by BuildResHPXML

aspeake:
## Pull Request Description

Adds a new argument `apply_defaults` in `BuildResidentialHPXML` to apply default values to the HPXML file. This allows a user to avoid running `HPXMLtoOpenstudio` if only the default HPXML file is needed.

## Checklist

Not all may apply:

### DIFF
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -3263,11 +3263,7 @@ class HPXMLFile
 
     if apply_defaults
       eri_version = Constants.ERIVersions[-1]
-      epw_path = args[:weather_station_epw_filepath]
-      if not File.exist? epw_path
-        epw_path = File.join(File.expand_path(File.join(File.dirname(__FILE__), '..', 'weather')), epw_path) # a filename was entered for weather_station_epw_filepath
-      end
-      weather, epw_file = Location.apply_weather_file(model, runner, epw_path, '')
+      weather = WeatherProcess.new(model, runner)      
       HPXMLDefaults.apply(hpxml, eri_version, weather, epw_file: epw_file)
     end
 
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -3263,7 +3263,7 @@ class HPXMLFile
 
     if apply_defaults
       eri_version = Constants.ERIVersions[-1]
-      weather = WeatherProcess.new(model, runner)      
+      weather = WeatherProcess.new(model, runner)
       HPXMLDefaults.apply(hpxml, eri_version, weather, epw_file: epw_file)
     end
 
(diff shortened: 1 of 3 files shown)


shorowit:
> @shorowit should there be a check somewhere in `HPXMLtoOpenstudio` so that `in.xml` is not written twice?

@aspeake I suggest avoiding this question by changing the argument to "Apply Defaults" and just apply the defaults to the HPXML output file that is already being written. No point in writing out two files.

aspeake:
> > @shorowit should there be a check somewhere in `HPXMLtoOpenstudio` so that `in.xml` is not written twice?
> 
> @aspeake I suggest avoiding this question by changing the argument to "Apply Defaults" and just apply the defaults to the HPXML output file that is already being written. No point in writing out two files.

@shorowit okay - so instead of writing in.xml, just apply the defaults to the existing hpxml object that gets written?

BuildResidentialHPXML/measure.rb
```diff
@@ -2868,6 +2876,12 @@ def arguments(model)
     arg.setDefaultValue(1.0)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeBoolArgument('apply_defaults', false)
+    arg.setDisplayName('Apply default values')
+    arg.setDescription('Sets default values to the HPXML object')
```
shorowit, line 2881:
```suggestion
    arg.setDescription('Sets OS-HPXML default values in the HPXML output file')
```

BuildResidentialHPXML/measure.rb
```diff
@@ -3241,6 +3255,22 @@ def self.create(runner, model, args, epw_file, skip_error_checking)
...
+      apply_defaults = false
+    end
+
+    if apply_defaults
+      eri_version = Constants.ERIVersions[-1]
+      epw_path = args[:weather_station_epw_filepath]
+      if not File.exist? epw_path
+        epw_path = File.join(File.expand_path(File.join(File.dirname(__FILE__), '..', 'weather')), epw_path) # a filename was entered for weather_station_epw_filepath
+      end
+      weather, epw_file = Location.apply_weather_file(model, runner, epw_path, '')
```
shorowit, line 3270: This code seems largely duplicative of [code found elsewhere in the measure](https://github.com/NREL/OpenStudio-HPXML/blob/9cbea592a2f3a07abf75366e084e992604cced7c/BuildResidentialHPXML/measure.rb#L2906-L2914). Can you consolidate these code blocks and then just pass around the, e.g. `epw_file`, as needed?

shorowit (commented):
The implementation looks correct but I have a suggestion and a comment.

shorowit (approved):
Looks good!

PR #948, opened 2022-01-07
Support multiple schedule file paths

joseph-robertson:
## Pull Request Description

Changes `SoftwareInfo/extension/SchedulesFilePath` to multiple `SoftwareInfo/extension/SchedulesFiles/SchedulesFile` elements with `Name` and `Path` child elements. See below. Likewise, the `SchedulesFile` class gets updated to import schedules from each file in `SchedulesFiles`.

```
  <SoftwareInfo>
    <extension>
      <SimulationControl>
        <Timestep>60</Timestep>
      </SimulationControl>
      <SchedulesFiles>
        <SchedulesFile>
          <Path>../../HPXMLtoOpenStudio/resources/schedule_files/smooth.csv</Path>
        </SchedulesFile>
        <SchedulesFile>
          <Path>../../HPXMLtoOpenStudio/resources/schedule_files/tank/hourly_setpoint_schedule.csv</Path>
        </SchedulesFile>
      </SchedulesFiles>
```

## Checklist

Not all may apply:

### DIFF
--- a/BuildResidentialScheduleFile/measure.rb
+++ b/BuildResidentialScheduleFile/measure.rb
@@ -64,7 +64,7 @@ class BuildResidentialScheduleFile < OpenStudio::Measure::ModelMeasure
 
     arg = OpenStudio::Measure::OSArgument.makeStringArgument('output_csv_path', true)
     arg.setDisplayName('Schedules: Output CSV Path')
-    arg.setDescription('Absolute (or relative) path of the csv file containing user-specified occupancy schedules.')
+    arg.setDescription('Absolute/relative path of the csv file containing user-specified occupancy schedules.')
     args << arg
 
     arg = OpenStudio::Measure::OSArgument.makeStringArgument('hpxml_output_path', true)
@@ -116,12 +116,19 @@ class BuildResidentialScheduleFile < OpenStudio::Measure::ModelMeasure
     # modify the hpxml with the schedules path
     doc = XMLHelper.parse_file(hpxml_path)
     extension = XMLHelper.create_elements_as_needed(XMLHelper.get_element(doc, '/HPXML'), ['SoftwareInfo', 'extension'])
-    schedules_filepath = XMLHelper.get_value(extension, 'SchedulesFilePath', :string)
-    if !schedules_filepath.nil?
-      runner.registerWarning("Overwriting existing SchedulesFilePath element: #{schedules_filepath}")
-      XMLHelper.delete_element(extension, 'SchedulesFilePath')
+    schedules_files = XMLHelper.get_elements(extension, 'SchedulesFiles')
+    if !schedules_files.empty?
+      runner.registerWarning('Overwriting existing SchedulesFiles.')
+      XMLHelper.delete_element(extension, 'SchedulesFiles')
     end
-    XMLHelper.add_element(extension, 'SchedulesFilePath', args[:output_csv_path], :string)
+    paths = { 'Occupancy' => args[:output_csv_path] }
+
+    schedules_files = XMLHelper.add_element(extension, 'SchedulesFiles')
+    paths.each do |name, path|
+      schedules_file = XMLHelper.add_element(schedules_files, 'SchedulesFile')
+      XMLHelper.add_element(schedules_file, 'Name', name, :string)
+      XMLHelper.add_element(schedules_file, 'Path', path, :string)
+    end
 
     # write out the modified hpxml
     hpxml_output_path = args[:hpxml_output_path]
@@ -129,10 +136,8 @@ class BuildResidentialScheduleFile < OpenStudio::Measure::ModelMeasure
       hpxml_output_path = File.expand_path(File.join(File.dirname(__FILE__), hpxml_output_path))
     end
 
-    if (hpxml_path != hpxml_output_path) || (schedules_filepath != args[:output_csv_path])
-      XMLHelper.write_file(doc, hpxml_output_path)
-      runner.registerInfo("Wrote file: #{hpxml_output_path}")
-    end
+    XMLHelper.write_file(doc, hpxml_output_path)
+    runner.registerInfo("Wrote file: #{hpxml_output_path}")
 
     return true
   end
--- a/HPXMLtoOpenStudio/resources/hpxml.rb
+++ b/HPXMLtoOpenStudio/resources/hpxml.rb
@@ -854,14 +854,19 @@ class HPXML < Object
   end
 
   class Header < BaseElement
+    def initialize(hpxml_object, *args)
+      @schedules_files = SchedulesFiles.new(hpxml_object)
+      super(hpxml_object, *args)
+    end
     ATTRS = [:xml_type, :xml_generated_by, :created_date_and_time, :transaction,
              :software_program_used, :software_program_version, :eri_calculation_version,
              :eri_design, :timestep, :building_id, :event_type, :state_code, :zip_code,
              :sim_begin_month, :sim_begin_day, :sim_end_month, :sim_end_day, :sim_calendar_year,
              :dst_enabled, :dst_begin_month, :dst_begin_day, :dst_end_month, :dst_end_day,
              :use_max_load_for_heat_pumps, :allow_increased_fixed_capacities,
-             :apply_ashrae140_assumptions, :energystar_calculation_version, :schedules_filepath]
+             :apply_ashrae140_assumptions, :energystar_calculation_version]
     attr_accessor(*ATTRS)
+    attr_reader(:schedules_files)
 
     def check_for_errors
       errors = []
@@ -888,6 +893,8 @@ class HPXML < Object
       end
 
       errors += HPXML::check_dates('Daylight Saving', @dst_begin_month, @dst_begin_day, @dst_end_month, @dst_end_day)
+
+      errors += @schedules_files.check_for_errors
 
       return errors
     end
@@ -948,10 +955,7 @@ class HPXML < Object
         XMLHelper.add_element(hvac_sizing_control, 'UseMaxLoadForHeatPumps', @use_max_load_for_heat_pumps, :boolean, @use_max_load_for_heat_pumps_isdefaulted) unless @use_max_load_for_heat_pumps.nil?
         XMLHelper.add_element(hvac_sizing_control, 'AllowIncreasedFixedCapacities', @allow_increased_fixed_capacities, :boolean, @allow_increased_fixed_capacities_isdefaulted) unless @allow_increased_fixed_capacities.nil?
       end
-      if not @schedules_filepath.nil?
-        extension = XMLHelper.create_elements_as_needed(software_info, ['extension'])
-        XMLHelper.add_element(extension, 'SchedulesFilePath', @schedules_filepath, :string) unless @schedules_filepath.nil?
-      end
+      @schedules_files.to_oga(software_info)
 
       building = XMLHelper.add_element(hpxml, 'Building')
       building_building_id = XMLHelper.add_element(building, 'BuildingID')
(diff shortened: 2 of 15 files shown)


shorowit:
@joseph-robertson Can we just make it one or more `SchedulesFilePath` elements at the same level (i.e., not inside a new parent element)? As you've proposed, it becomes a breaking change for software tools using OS-HPXML.

joseph-robertson:
> @joseph-robertson Can we just make it one or more `SchedulesFilePath` elements at the same level (i.e., not inside a new parent element)? As you've proposed, it becomes a breaking change for software tools using OS-HPXML.

```
  <SoftwareInfo>
    <extension>
      <SchedulesFilePath>../../HPXMLtoOpenStudio/resources/schedule_files/smooth.csv</SchedulesFilePath>
      <SchedulesFilePath>../../HPXMLtoOpenStudio/resources/schedule_files/stochastic.csv</SchedulesFilePath>
      ...
    </extension>
  </SoftwareInfo>
```
Like that? Schema permits that?

shorowit:
Yes, like that. This is inside an `extension` element so the HPXML schema doesn't restrict anything here, you can do anything that is valid XML.

shorowit:
Here's [a similar example](https://github.com/NREL/hescore-hpxml/blob/master/examples/house1_v3.xml#L60-L61) of doing this in an HPXML file.

shorowit:
Do we need error-checking for a column header that is duplicated?

tasks.rb
```diff
@@ -2505,11 +2505,11 @@ def apply_hpxml_modification(hpxml_file, hpxml)
...
     hpxml.header.allow_increased_fixed_capacities = true
   elsif ['base-schedules-detailed-stochastic.xml'].include? hpxml_file
-    hpxml.header.schedules_filepath = '../../HPXMLtoOpenStudio/resources/schedule_files/stochastic.csv'
+    hpxml.header.schedules_filepaths << '../../HPXMLtoOpenStudio/resources/schedule_files/stochastic.csv'
   elsif ['base-schedules-detailed-stochastic-vacancy.xml'].include? hpxml_file
-    hpxml.header.schedules_filepath = '../../HPXMLtoOpenStudio/resources/schedule_files/stochastic-vacancy.csv'
+    hpxml.header.schedules_filepaths << '../../HPXMLtoOpenStudio/resources/schedule_files/stochastic-vacancy.csv'
   elsif ['base-schedules-detailed-smooth.xml'].include? hpxml_file
-    hpxml.header.schedules_filepath = '../../HPXMLtoOpenStudio/resources/schedule_files/smooth.csv'
+    hpxml.header.schedules_filepaths << '../../HPXMLtoOpenStudio/resources/schedule_files/smooth.csv'
```
shorowit, line 2512: Hmm. Can we start calling the BuildResSchedule measure for these sample files so we don't have to manually modify the HPXML object?

workflow/sample_files/base-misc-defaults.xml
```diff
@@ -6,7 +6,9 @@
     <CreatedDateAndTime>2000-01-01T00:00:00-07:00</CreatedDateAndTime>
     <Transaction>create</Transaction>
   </XMLTransactionHeaderInformation>
-  <SoftwareInfo/>
+  <SoftwareInfo>
+    <extension/>
+  </SoftwareInfo>
```
shorowit, line 11: Can you update the HPXML class to prevent this change?

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1333,52 +1378,89 @@ def get_external_file
...
-      ColumnPoolPump => false,
-      ColumnPoolHeater => false,
-      ColumnHotTubPump => false,
-      ColumnHotTubHeater => false,
-      ColumnHotWaterDishwasher => true,
-      ColumnHotWaterClothesWasher => true,
-      ColumnHotWaterFixtures => true,
-      ColumnVacancy => nil,
-    }
+    return SchedulesFile.OccupancyColumnNames
```
shorowit, line 1396: I don't understand... `ColumnNames` is the same as `OccupancyColumnNames`? What's the point?

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1333,52 +1378,89 @@ def get_external_file
   end
 
   def set_vacancy
-    return unless @tmp_schedules.keys.include? 'vacancy'
-    return if @tmp_schedules['vacancy'].all? { |i| i == 0 }
+    return unless @tmp_schedules.keys.include? ColumnVacancy
+    return if @tmp_schedules[ColumnVacancy].all? { |i| i == 0 }
```
shorowit, line 1382: Happy to see these changes.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -948,9 +948,11 @@ def to_oga(doc)
         XMLHelper.add_element(hvac_sizing_control, 'UseMaxLoadForHeatPumps', @use_max_load_for_heat_pumps, :boolean, @use_max_load_for_heat_pumps_isdefaulted) unless @use_max_load_for_heat_pumps.nil?
         XMLHelper.add_element(hvac_sizing_control, 'AllowIncreasedFixedCapacities', @allow_increased_fixed_capacities, :boolean, @allow_increased_fixed_capacities_isdefaulted) unless @allow_increased_fixed_capacities.nil?
       end
-      if not @schedules_filepath.nil?
+      if not @schedules_filepaths.nil?
```
shorowit, line 951: Related to the base-misc-defaults.xml comment, maybe needs to be `if (not @schedules_filepaths.nil?) && (not @schedules_filepaths.empty?)`

shorowit (commented):
This looks pretty good, nice work @joseph-robertson. Just a few comments.

HPXMLtoOpenStudio/resources/schedules.rb
joseph-robertson: It's the setup for the `wh-geb` branch; eventually they won't be the same.

tasks.rb
joseph-robertson: This is tough. The `SchedulesFilePath` is relative the `workflow/sample_files` but it is also the path for exporting the schedules, relative `BuildResidentialScheduleFile/measure.rb`.

joseph-robertson: So, we could use an absolute path, but this would require that the sample files get updated on the ci...

shorowit: I don't follow what you mean by "but it is also the path for exporting the schedules, relative BuildResidentialScheduleFile/measure.rb". We definitely don't want to use an absolute path -- what works on your machine won't work on my machine.

BuildResidentialScheduleFile/measure.rb
```diff
@@ -96,6 +96,12 @@ def run(model, runner, user_arguments)
       fail "'#{hpxml_path}' does not exist or is not an .xml file."
     end
 
+    hpxml_output_path = args[:hpxml_output_path]
+    unless (Pathname.new hpxml_output_path).absolute?
+      hpxml_output_path = File.expand_path(File.join(File.dirname(__FILE__), hpxml_output_path))
+    end
+    args[:hpxml_output_path] = hpxml_output_path
```
shorowit, line 103: I know all you did was move some existing code around but... I really dislike seeing relative paths as interpreted relative to the measure itself. We need to stop doing that, it is totally confusing when the measure's location is unpredictable. I'm tempted to suggest that we should stop allowing relative paths for input/output paths in any measure. Does anyone ever really need that?

(On the other hand, for paths that end up _inside the HPXML file_, it makes complete sense to allow relative paths -- e.g., to point to a file that's in the same directory as the HPXML file itself.)

I suppose we don't have to address this now, but it might be as good a time as any.

tasks.rb
```diff
@@ -382,12 +382,16 @@ def create_hpxmls
...
       args = {}
+      sch_args = {}
       all_hpxml_files.each do |f|
-        set_measure_argument_values(f, args)
+        set_measure_argument_values(f, args, sch_args)
       end
 
+      measures = {}
+      measures['BuildResidentialHPXML'] = [args] if !args.empty?
+      measures['BuildResidentialScheduleFile'] = [sch_args] if !sch_args.empty?
```
shorowit, line 392: Awesome, thanks for making this change! 💯

workflow/tests/base_results/results.csv
```diff
@@ -310,8 +310,8 @@ base-pv-battery-outside-degrades.xml,34.98,8.43,14.53,0.0,0.0,0.0,0.0,0.0,0.0,0.
...
+base-schedules-detailed-stochastic.xml,35.08,35.08,15.16,0.0,0.0,0.0,0.0,0.0,0.0,0.39,3.55,0.69,9.16,0.0,0.0,4.51,0.0,0.33,0.0,0.0,0.0,0.0,2.22,0.0,0.0,0.32,0.37,1.51,1.53,0.0,2.12,8.38,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,15.16,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,14.33,11.6,9.23,0.61,0.0,0.0,0.0,1.0,4702.0,5523.0,31.07,21.49,0.0,3.88,3.98,0.54,3.99,0.71,-0.13,0.0,0.0,3.06,-0.1,5.49,0.0,0.83,0.0,3.46,-11.47,0.0,-0.02,-0.45,-0.01,3.83,-0.03,8.9,0.0,0.0,-6.01,-0.05,-1.14,-6.55,-0.16,0.0,3.07,10.27,1355.0,998.0,11397.0,2615.0
```
shorowit, line 314: Why did you have to make changes to these schedule files?

joseph-robertson: Well, we're calling BuildResidentialScheduleFile now to generate those files. Apparently they got updated. It appears the vacancy period is something other than Dec 15 - Jan 15 -- I'll look into `stochastic-vacancy.csv`. Not sure about `stochastic.csv`.

shorowit: Ah, right. Awesome! I hadn't even thought of the benefit that we'll now be ensuring that those files stay up to date (and seeing if they ever accidentally/incorrectly change).

BuildResidentialScheduleFile/measure.rb
joseph-robertson: Right, maybe we don't need relative paths. For ResStock-HPXML, it looks like we're using absolute paths within the `BuildExistingModel` meta measure: https://github.com/NREL/resstock/blob/restructure-v3/measures/BuildExistingModel/measure.rb#L229.

joseph-robertson: @shorowit So what should be done here? Update all comments by removing all "relative path" references? Also, have checks that all provided path arguments are absolute?

shorowit: I will create a separate issue for this. Let's not hold up your PR any further.

workflow/tests/base_results/results.csv
shorowit: I'd still like to understand why the schedules have changes. In particular, the `hot_water_foo` columns in stochastic.csv are quite a bit different than before. Can we figure out why?

joseph-robertson: One thing I think of is the pseudo random number generator `prng` that floats around in `schedules.rb`. The order that it's used determines random offsets.

This is hard to track down.

shorowit: What do you mean by "the order that it's used"?

joseph-robertson: For example, if in this example `sinks` came before `showers`, you'd get something different: https://github.com/NREL/OpenStudio-HPXML/blob/master/BuildResidentialScheduleFile/resources/schedules.rb#L711-L723

shorowit: I see. Okay, I won't belabor the point any further.

shorowit (approved):

workflow/tests/base_results/results.csv
joseph-robertson: But I do think those schedule csv files will be much more tractable. Starting now...

PR #952, opened 2022-01-11
Increase output precision as needed

### LINKED ISSUE #951: Increase output resolution
When running simulations with shorter timesteps (e.g., 1 minute instead of 1 hour) or abbreviated run periods (e.g., 1 day instead of a full year), the output resolution for annual and timeseries data may be too low.

cc @yzhou601 @GamalielL

shorowit:
## Pull Request Description

Fixes #951. Adds more decimal places in output files as needed for simulations with shorter timesteps and/or abbreviated run periods.

## Checklist

Not all may apply:


yzhou601 (approved):
I have run a test file for 07/16 (Wall R-5), the energy curve resolution is much better! Well done! It is more obvious for fan/pump energy reporting because they're smaller numbers.
Before:
image
image


After:
image
image


I observed there're spikes in the sensitivity analysis results (graph 1&2), while they no longer exist in HPXML-OS results (graph 3&4) , and I compared the idfs between these two runs, the only diffs are caused by sizing, so should be something unrelated to the modeling approach.


And results_annual.csv also shows more digits, these changes look great to me!

PR #971, opened 2022-01-28
Allow heat pump w/ central backup system

### LINKED ISSUE #947: Allow HP separate backup system to be a central system?
When a heat pump has a backup system that is "separate" (i.e., not "integrated"), [we don't allow it to be a central system](https://openstudio-hpxml.readthedocs.io/en/v1.3.0/workflow_inputs.html#footnote-reference-151). Investigate why this limitation exists and whether it can be relaxed; I don't remember offhand if it was a limitation in EnergyPlus that prevents modeling this, or would just take a fair amount of effort to support.

As a workaround, a user might be able to specify a DSE with a non-central system.

cc @chrisbalbach

shorowit:
## Pull Request Description

Closes #947. Allows a heat pump separate backup system to be a central system (e.g, central furnace w/ ducts). Previously only non-central system types were allowed (boiler, electric resistance, wall/floor furnace, etc.); now every type of heating system is allowed.

(Note: This system cannot be configured through the BuildResidentialHPXML measure as it requires two air distribution systems to be specified -- one for the heat pump and one for the backup central system -- while the measure currently only supports a single air distribution system.)

## Checklist

Not all may apply:


shorowit:
FYI @chrisbalbach

PR #979, opened 2022-02-11
Water heater approach for combi systems

yzhou601:
## Pull Request Description

Addresses https://github.com/NREL/OpenStudio-HPXML/issues/811

## Checklist

Not all may apply:


yzhou601:
@shorowit Could you take a look at the simulation results, and see how should we proceed?

shorowit:
@yzhou601 will try to determine how much of the energy difference is due to 1) pump power, 2) boiler performance curve, and 3) other things. If (1) is important, maybe there's a way to capture the pump power in the estimated EF, but @yzhou601 thinks it is not important.

yzhou601:
I was thinking the pump energy in the wrong place, it should be in the graph: 
image
The above is the diffs between current master pump energy and current water heater approach pump energy
If I removed pump energy, this graph is no longer displayed, they're all just 0.0 data points.

However, I have confirmed that the performance curve is the reason of `End Use: Natural Gas: Hot Water(MBtu)` diffs.
With boiler performance curves:
image
Without boiler performance curves (constant 1.0 curve):
image

There's [an input in WaterHeater:Mixed object that allows user to define a part load fraction curve](https://bigladdersoftware.com/epx/docs/9-6/input-output-reference/group-water-heaters.html#field-part-load-factor-curve-name). I know that the boiler performance curves are more complicated curves, and the "part load" is different concepts in these two situations. For water heater approach, it can only be water heating loads being accounted for in calculating PLF, however, in the boiler modeling approach, it is the total of space heating and water heating loads that is accounted for in calculating PLR. So I am not sure how much useful this field is and whether we can make good use of it.

shorowit:
I came across a couple resources from other residential software tools that tell their uses to model combi systems as water heaters: 
- One says to use EF = boiler AFUE x 0.8
- The other says to use EF = boiler AFUE x 0.75

If we take into account both end uses (fan power and hot water), what fraction do we calculate for the above data points?

yzhou601:
> I came across a couple resources from other residential software tools that tell their uses to model combi systems as water heaters:
> 
> * One says to use EF = boiler AFUE x 0.8
> * The other says to use EF = boiler AFUE x 0.75
> 
> If we take into account both end uses (fan power and hot water), what fraction do we calculate for the above data points?

Pump and hot water are of different end uses, changing EF can address the hot water natural gas diffs in some way, but pump electricity end use loss will not be captured (though they're small numbers). What about trying 0.8 first to see how natural gas regression results change?

shorowit:
> What about trying 0.8 first to see how natural gas regression results change?

Sure, go for it.

Maybe I'm not thinking about it correctly, but I thought we could simply calculate our equivalent factor for these runs -- seems to be ~12% reduction in gas energy use (since the +/- lines are 10%), so ~0.88?

yzhou601:
> > What about trying 0.8 first to see how natural gas regression results change?
> 
> Sure, go for it.
> 
> Maybe I'm not thinking about it correctly, but I thought we could simply calculate our equivalent factor for these runs -- seems to be ~12% reduction in gas energy use (since the +/- lines are 10%), so ~0.88?

Yep I can try both 0.88 and 0.8. Got your point, it is actually a question about whether we want the water heater approach to be more close to previous approach results, or more aligned with other software assumptions. Our previous plant loop approach is modeling more details about the systems, though its robustness is a question. I don't know how the other software comes up with numbers of 0.75 and 0.8.

yzhou601:
The closest adjustment to water heater eta_c will be around 0.86 to 0.87.
0.86:
image
0.87:
image

shorowit:
I think the alternative two boiler approach, which has already been merged, is a good compromise. While this PR would provide some significant code simplification, it also adds some complications related to boiler vs water heater inputs/models. So I'm going to close this.

Issue #958, opened 2022-01-18
Address relative paths for measure arguments


shorowit:
From a discussion [here](https://github.com/NREL/OpenStudio-HPXML/pull/948#discussion_r783542243).

Measure arguments that ask for a path often allow either absolute or relative paths. In general, relative paths are being interpreted as relative to the measure itself. This is likely very confusing as the measure's location can be unpredictable (like when an OS workflow is run).

We should either:
1. Continue to allow relative paths but document that they are relative to the measure's location,
2. Disallow relative paths (i.e., only allow absolute paths), or
3. Continue to allow relative paths but change what they are relative to.

cc @joseph-robertson

PR #981, opened 2022-02-16
HPXML File Path as relative to run directory (for PAT)

joseph-robertson:
## Pull Request Description

Currently, for both the `BuildResidentialHPXML` and `HPXMLtoOpenStudio` measures, if you specify `hpxml_path` as relative then this code gets executed:

`hpxml_path = File.expand_path(File.join(File.dirname(__FILE__), hpxml_path))`

For the purposes of collecting exported `in.xml` files, the previous line has different implications when using `OSApp` vs `PAT`:
* `OSApp`: `in.xml` is found relative to measures in `MyMeasures Directory`. Probably where you'd expect it, and where you'd want it to be.
* `PAT`: `in.xml` is found relative to measures found in the project's `analysis_xxx` measure folder. And therefore not contained in the `data_point.zip`. This is a problem because then you cannot easily download the `in.xml` files from OS-Server.

**This PR adds an argument** (`hpxml_path_relative_to_run_directory`) to both `BuildResidentialHPXML` and `HPXMLtoOpenStudio` for optionally changing the above line to:

`hpxml_path = File.expand_path(hpxml_path)`

Again, different implications for `OSApp` vs `PAT`:
* `OSApp`: Now found in a temporary directory (e.g., `~/AppData/.../ApplyMeasuresNow/run/in.xml`). Probably not a useful argument when using `OSApp`.
* `PAT`: Now found relative to the measures folder found in the data point's run folder. Therefore contained in the `data_point.zip`. Can easily download the `in.xml` files from OS-Server.

I suppose this most closely addresses item 3 from #958.

## Checklist

Not all may apply:


shorowit:
I'm not sure I fully understand all the nuances here, but here are my thoughts:
1. Supporting PAT seems generally more important than supporting the OSApp. We have seen multiple automated uses of OS-HPXML via PAT, while it's hard to imagine a lot of users wanting to manually run the measures via OSApp's Apply Measures Now.
2. The behavior defined by `hpxml_path_relative_to_run_directory` (which is probably better described as relative to the _working_ directory; the _run_ directory is used in OS to mean where the IDF file ends up and the simulation is run from) seems more useful than the current behavior. In the current behavior, things are relative to the measure dir, which may end up god knows where in any given OS workflow.
3. The behavior defined by `hpxml_path_relative_to_run_directory` seems to be aligned with other OpenStudio reporting measures -- like [the openstudio results reporting measure](https://github.com/NREL/openstudio-common-measures-gem/blob/develop/lib/measures/openstudio_results/measure.rb#L355-L356), which is probably the most widely used reporting measure the exists. If it's good enough for users of this measure, it ought to be good enough for us.

So my inclination is to not add a new argument, but change to the behavior you proposed for PAT.

shorowit:
Or maybe we can do what we do for our reporting measures and use [this](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportSimulationOutput/measure.rb#L435). I think that perfectly ensures that our output products end up in the same directory as the IDF/EPW. Though maybe it's not available in a ModelMeasure...

shorowit:
Whatever we do, we need to also change the BuildResidentialScheduleFile measure. And please audit the code for any other uses.

joseph-robertson:
> Or maybe we can do what we do for our reporting measures and use [this](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportSimulationOutput/measure.rb#L435). I think that perfectly ensures that our output products end up in the same directory as the IDF/EPW, which is really what we want. Though maybe it's not available in a ModelMeasure...

It does not appear available in a ModelMeasure.

PR #989, opened 2022-03-01
Operational Calculation (occupancy-based loads)

### LINKED ISSUE #644: Operational Calculation (occupancy-based loads)
Allow a user to do an operational, instead of asset, calculation through a new HPXML input field. For example, we could adopt the [ResStock occupancy adjustments](https://github.com/NREL/resstock/blob/a5a60718ec58bfbbc3c04c186ca30155184c106d/measures/ResStockArguments/measure.rb#L413-L424) for appliances and hot water when an operational calculation is to be performed.

joseph-robertson:
## Pull Request Description

Closes #644.

**Breaking change**: The occupancy calculation type is entered in ``/HPXML/SoftwareInfo/extension/OccupancyCalculationType``: either "asset" or "operational".

If OccupancyCalculationType is "asset", various end uses (e.g., clothes washer) are calculated using number of bedrooms and/or conditioned floor area.

If OccupancyCalculationType is "operational", end uses based on number of bedrooms are adjusted for the number of occupants using the relationship from [RECS 2015](https://www.eia.gov/consumption/residential/reports/2015/overview/):

- single-family detached or manufactured home: NumberofBedrooms = -1.47 + 1.69 * NumberofResidents
- single-family attached or apartment unit: NumberofBedrooms = -0.68 + 1.09 * NumberofResidents

Adjusts:

- [ ] Water fixtures
- [ ] Clothes washer
- [ ] Clothes dryer
- [ ] Dishwasher
- [ ] Cooking range
- [ ] ~Other plug load~ (not dependent on nbeds)
- [ ] Television plug load
- [ ] ~Vehicle plug load~ (not dependent on nbeds)
- [ ] Well pump plug load
- [ ] Grill, lighting, fireplace fuel loads
- [ ] Pool pump, heater
- [ ] Hot tub pump, heater
- [ ] ~Lighting~ (not dependent on nbeds)
- [ ] ~Refrigerators/freezers~ (not dependent on nbeds)

## Checklist

Not all may apply:


HPXMLtoOpenStudio/measure.rb
```diff
@@ -1929,21 +1942,46 @@ def self.add_lighting(runner, model, epw_file, spaces)
   end
 
   def self.add_pools_and_hot_tubs(runner, model, spaces)
+    occ_calc_type = @hpxml.header.occupancy_calculation_type
+    unit_type = @hpxml.building_construction.residential_facility_type
+    cfa = @hpxml.building_construction.conditioned_floor_area
+    nbeds = @hpxml.building_construction.number_of_bedrooms
+    noccs = @hpxml.building_occupancy.number_of_residents
+    if occ_calc_type == 'operational'
+      adj_factor = HPXMLDefaults.get_misc_luls_adjustment_factor(unit_type, nbeds, noccs, cfa)
```
shorowit, line 1951: Rather than calculate the same factor multiple places across the source code, just calculate it once and reuse it. For example, you could calculate it in hpxml_defaults.rb and stored it on `hpxml.header.additional_properties`.

joseph-robertson:
@shorowit Can you give me some ideas/hints on how to structure the documentation around this? I put in a few placeholders, but could use some suggestions.

BuildResidentialHPXML/measure.rb
```diff
@@ -60,6 +60,15 @@ def arguments(model)
...
     args << arg
 
+    occupancy_calculation_type_choices = OpenStudio::StringVector.new
+    occupancy_calculation_type_choices << 'asset'
+    occupancy_calculation_type_choices << 'operational'
+
+    arg = OpenStudio::Measure::OSArgument.makeChoiceArgument('occupancy_calculation_type', occupancy_calculation_type_choices, true)
+    arg.setDisplayName('Occupancy Calculation Type')
+    arg.setDescription("The type of occupancy calculation type. If 'asset' is chosen, usage multipliers not provided continue to default to 1.0 (meaning energy use is calculated based on floor area, # bedrooms, etc.). If 'operational' is chosen, usage multipliers not provided are calculated based on # occupants, floor area, # bedrooms, etc.")
+    args << arg
```
shorowit, line 70: Did you test this? Looking at the code suggests this will have no impact because all usage multiplier arguments throughout this measure are required, are of type `double`, and default to 1.0. So for an operational calculation, how would someone let the default usage multipliers apply?

joseph-robertson: I wouldn't pay attention to this argument description. (That's what you're referring too, right?) This was our original approach, but has since changed. The approach now is to adjust usage multipliers no matter their value (default value or not).

shorowit: Oh, right! Thanks for reminding me, I had forgotten which approach we ended up with. I'll review with that in mind.

BuildResidentialHPXML/measure.rb
```diff
@@ -60,6 +60,15 @@ def arguments(model)
     arg.setDescription('Absolute/relative path of the HPXML file.')
     args << arg
 
+    occupancy_calculation_type_choices = OpenStudio::StringVector.new
+    occupancy_calculation_type_choices << 'asset'
+    occupancy_calculation_type_choices << 'operational'
```
shorowit, line 65: Use HPXML constants here and elsewhere.

BuildResidentialHPXML/measure.rb
```diff
@@ -60,6 +60,15 @@ def arguments(model)
...
+
+    arg = OpenStudio::Measure::OSArgument.makeChoiceArgument('occupancy_calculation_type', occupancy_calculation_type_choices, true)
+    arg.setDisplayName('Occupancy Calculation Type')
+    arg.setDescription("The type of occupancy calculation type. If 'asset' is chosen, usage multipliers not provided continue to default to 1.0 (meaning energy use is calculated based on floor area, # bedrooms, etc.). If 'operational' is chosen, usage multipliers not provided are calculated based on # occupants, floor area, # bedrooms, etc.")
```
shorowit, line 69:
```suggestion
    arg.setDescription("The type of occupancy calculation type. If 'asset' is chosen, usages of plug loads, appliances, hot water, etc. are based on number of bedrooms or conditioned floor area. If 'operational' is chosen, these usages are adjusted based on the number of occupants relative to the number of bedrooms.")
```

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1932,18 +1934,22 @@ def self.add_pools_and_hot_tubs(runner, model, spaces)
     @hpxml.pools.each do |pool|
       next if pool.type == HPXML::TypeNone
 
+      pool.heater_usage_multiplier *= @hpxml.header.additional_properties.misc_loads_adj_factor if pool.heater_type != HPXML::TypeNone
```
shorowit, line 1937: Doesn't seem like `if pool.heater_type != HPXML::TypeNone` is needed since this code is right below `next if pool.type == HPXML::TypeNone`.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -2604,4 +2621,25 @@ def self.get_orientation_from_azimuth(azimuth)
...
+  def self.get_misc_loads_adj_factor(unit_type, nbeds, noccs, cfa)
+    if [HPXML::ResidentialTypeApartment, HPXML::ResidentialTypeSFA].include? unit_type
+      c = [-0.68, 1.09]
+    elsif [HPXML::ResidentialTypeSFD].include? unit_type
+      c = [-1.47, 1.69]
+    end
+    adj_factor = (0.5 + 0.25 * (c[0] + c[1] * noccs) / 3.0 + 0.25 * cfa / 1920.0) /
+                 (0.5 + 0.25 * nbeds / 3.0 + 0.25 * cfa / 1920.0)
+    return adj_factor
+  end
```
shorowit, line 2644: Hmm. I see that you derived these equations from [here](https://github.com/NREL/resstock/blob/223d5de88caf07f1d7cfd82a4600cbdeca355d56/measures/ResStockArguments/measure.rb#L491-L497) for LULs, but is it correct the same the adjustments to [plug loads](https://github.com/NREL/resstock/blob/223d5de88caf07f1d7cfd82a4600cbdeca355d56/measures/ResStockArguments/measure.rb#L432-L440)?

shorowit: I will look into this a bit.

shorowit: For LULs, it looks like the equations can be reasonably simplified to be more consistent with the hot water and appliance equations:
- SFD: `occ_factor = occ_to_nbr_ratio**0.385`
- SFA/MF: `occ_factor = 0.68 * occ_to_nbr_ratio**0.224`

image

shorowit: Sorry, disregard my previous comment. Let me investigate more.

HPXMLtoOpenStudio/measure.rb
joseph-robertson: Pool can, and cannot, have a heater, right? (`type` vs `heater_type`)

shorowit: Ah, you are right, I didn't notice that subtlety. Thanks.

shorowit (commented):
I'm getting stuck on two things here:
1. We're always applying the occupant adjustment factor to all the LULs when an operational calculation is requested. But the LULs should only be adjusted if they are defaulted; if their kWh/yr or therm/yr is provided, we should assume that input reflects the actual usage to be modeled, not some base asset usage that still needs to be adjusted for occupants.
2. It doesn't seem like we are covering all the end uses we should be. For example, this adjusts certain LULs but not MELs/MGLs. Seems like we need to adjust MELs/MGELs as well even if ResStock is going to provide its own numbers from RECS data.

joseph-robertson:
For (1), should we move the adjustments to `hpxml_defaults.rb`?

For (2), I may have missed https://github.com/NREL/resstock/blob/develop/measures/ResStockArguments/measure.rb#L444-L452 for "other" plug loads.

shorowit:
For (1), yes I think that's the right solution.
For (2), note that it's a bit tricky because we want to preserve the current OS-HPXML plug load equation (based on CFA/Nbeds from the RESNET 301 Standard) but adjust it in a way similar to ResStock. Which is why ResStock will need to continue providing its own values from RECS.

joseph-robertson:
Actually for (1), couldn't I just keep the adjustments where they are but use the `xxx_isdefaulted` property to determine if it was defaulted or not?

shorowit:
Yes that should work too.

shorowit (approved):
Looks good to me.

PR #1003, opened 2022-03-15
Requesting user-defined output variables

### LINKED ISSUE #997: Requesting user-defined output variables
The current `ReportSimulationOutput` measure provides standard sets of commonly requested outputs. What if we want to request outputs outside of these standard sets?

Revive `TimeseriesCSVExport` or at least the portion of it that enables requesting output variables? (We could use existing BCL measures like `AddOutputVariable` and its companion reporting measure (?) to make these requests, but how would we integrate the results with `results_timeseries.csv`?)

joseph-robertson:
## Pull Request Description

Closes #997.

This PR:
* Adds a new optional `user_output_variables` argument to `ReportSimulationOutput`
* Argument is comma-separated string if multiple output variables requested
* Requests and reports all key values (i.e., `*`) for each output variable requested
* Adds new output reporting tests for when `user_output_variables` is populated:
  * Standard variables (e.g., `Zone People Occupant Count`)
  * Advanced variables (e.g., `Surface Construction Index`) - `add_component_loads` must be set to `true`

## Checklist

Not all may apply:

- [ ] ~EPvalidator.xml has been updated~


Changelog.md
```diff
@@ -17,6 +17,7 @@ __New Features__
   - Adds an optional argument for whether the HPXML file is written with default values applied; defaults to false.
   - Adds an optional argument for whether the HPXML file is validated; defaults to false.
 - ReportSimulationOutput measure:
+  - Add optional argument for generating user-defined timeseries output variables.
```
shorowit, line 20:
```suggestion
  - Add optional argument for requesting timeseries EnergyPlus output variables.
```

ReportSimulationOutput/measure.rb
```diff
@@ -123,6 +123,11 @@ def arguments(model)
     arg.setDefaultValue(false)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeStringArgument('user_output_variables', false)
+    arg.setDisplayName('Generate Timeseries Output: User-Defined Output Variables')
+    arg.setDescription('Optionally generates timeseries output variables. If multiple output variables, use a comma-separated list. Do not include key values; by default all key values will be requested.')
```
shorowit, line 128:
```suggestion
    arg.setDisplayName('Generate Timeseries Output: EnergyPlus Output Variables')
    arg.setDescription('Optionally generates timeseries EnergyPlus output variables. If multiple output variables are desired, use a comma-separated list. Do not include key values; by default all key values will be requested. Example: "Zone People Occupant Count, Zone People Total Heating Energy"')
```

I took these output variable names from your test below so I assume it's correct, but I haven't tested it.

ReportSimulationOutput/measure.rb
```diff
@@ -1558,18 +1582,23 @@ def write_timeseries_output_results(runner, outputs, output_format,
     else
       weather_data = []
     end
+    if !@output_variables.empty?
+      output_variables_data = @output_variables.values.select { |x| x.timeseries_output.sum(0.0) != 0 }.map { |x| [x.name, x.timeseries_units] + x.timeseries_output.map { |v| v.round(dig) } }
```
shorowit, line 1586: Since we don't know what kind of units/output is being requested, it's probably a bad idea to round the value. I suggest we just report what EnergyPlus gives us.

```suggestion
      output_variables_data = @output_variables.values.select { |x| x.timeseries_output.sum(0.0) != 0 }.map { |x| [x.name, x.timeseries_units] + x.timeseries_output }
```

ReportSimulationOutput/measure.rb
```diff
@@ -2188,6 +2243,15 @@ def get_timeseries_units_from_fuel_type(fuel_type)
     @weather.each do |weather_type, weather_data|
       weather_data.name = "Weather: #{weather_type}"
     end
+
+    # Output Variables
+    @output_variables_requests = {}
+    if @user_output_variables
+      output_variables = @user_output_variables.split(',').map(&:strip)
```
shorowit, line 2250: 👍 Glad to see the `.map(&:strip)` here!

ReportSimulationOutput/tests/output_report_test.rb
```diff
@@ -976,6 +1026,59 @@ def test_timeseries_local_time_dst_and_utc
...
+                  'include_timeseries_fuel_consumptions' => false,
+                  'include_timeseries_end_use_consumptions' => false,
+                  'include_timeseries_emissions' => false,
+                  'include_timeseries_hot_water_uses' => false,
+                  'include_timeseries_total_loads' => false,
+                  'include_timeseries_component_loads' => false,
+                  'include_timeseries_zone_temperatures' => false,
+                  'include_timeseries_airflows' => false,
+                  'include_timeseries_weather' => false,
+                  'user_output_variables' => 'Zone People Occupant Count, Zone People Total Heating Energy' }
```
shorowit, line 1041: Can you add a test (or update this one) to include an invalid output variable request (like "Foo")? I think we'd like the measure to still run successfully, but without this column in the timeseries output. Even better would be if we can register a warning for each output variable that is invalid (i.e., doesn't produce EnergyPlus output) and check for the warning in the test.

docs/source/workflow_outputs.rst
```diff
@@ -356,6 +356,7 @@ Depending on the outputs requested, the file may include:
    Zone Temperatures                    Average temperatures (in deg-F) for each space modeled (e.g., living space, attic, garage, basement, crawlspace, etc.).
    Airflows                             Airflow rates (in cfm) for infiltration, mechanical ventilation (including clothes dryer exhaust), natural ventilation, whole house fans.
    Weather                              Weather file data including outdoor temperatures, relative humidity, wind speed, and solar.
+   Output Variables                     These are optional and user-defined. For advanced variables, the ``--add-component-loads`` argument must be used.
```
shorowit, line 359: I think it's pretty unusual to request advanced EnergyPlus output variables and think this language will just confuse most users. If a user is requesting an advanced variable, they should know what they're doing.

```suggestion
   EnergyPlus Output Variables          These are optional and can be requested with the ReportSimulationOutput ``user_output_variables`` argument.
```

workflow/template.osw
```diff
@@ -29,7 +29,8 @@
         "include_timeseries_airflows": false,
         "include_timeseries_weather": false,
         "add_timeseries_dst_column": false,
-        "add_timeseries_utc_column": false
+        "add_timeseries_utc_column": false,
+        "user_output_variables": ""
```
shorowit, line 33: Let's make this change to the other OSWs too.

ReportSimulationOutput/measure.rb
```diff
@@ -1558,18 +1582,23 @@ def write_timeseries_output_results(runner, outputs, output_format,
     else
       weather_data = []
     end
+    if !@output_variables.empty?
```
shorowit, line 1585:
```suggestion

    # EnergyPlus output variables
    if !@output_variables.empty?
```

shorowit (commented):
This looks really good. Just some minor suggestions primarily around descriptions.

Is it worth adding this capability to run_simulation.rb? E.g., `openstudio run_simulation.rb -x foo.xml --add-output-variables "Zone People Occupant Count, Zone People Total Heating Energy"` or similar?

workflow/run_simulation.rb
```diff
@@ -125,6 +126,11 @@ def run_workflow(basedir, rundir, hpxml, debug, timeseries_output_freq, timeseri
     options[:timeseries_time_column_types] << t
   end
 
+  options[:timeseries_output_variables] = []
+  opts.on('--add-timeseries-output-variable NAME', 'Add timeseries output variable; can be called multiple times') do |t|
+    options[:timeseries_output_variables] << t
+  end
```
shorowit, line 132: I tried this new option but realized it only works if you request a standard timeseries output too because the timeseries frequency needs to be set. For example, I tried `openstudio workflow/run_simulation.rb -x workflow/sample_files/base.xml -o test --add-timeseries-output-variable 'Site Total Sky Cover'` and no timeseries output file was produced.

Any ideas on how to address this?

If I had to do over, I would replace all the `--hourly`, `--daily`, etc with a `--timeseries-frequency` and `--add-timeseries-output-type` or something. That would have been smarter. But I'd prefer not to make a breaking change at this point.

workflow/tests/hpxml_translator_test.rb
```diff
@@ -207,6 +207,29 @@ def test_run_simulation_timeseries_dst_and_utc
     assert_equal(1, timeseries_rows[0].select { |r| r == 'TimeUTC' }.size)
   end
 
+  def test_run_simulation_timeseries_output_variables
+    # Check that the simulation produces timeseries with request output variables
+    rb_path = File.join(File.dirname(__FILE__), '..', 'run_simulation.rb')
+    xml = File.join(File.dirname(__FILE__), '..', 'sample_files', 'base.xml')
+    command = "#{OpenStudio.getOpenStudioCLI} #{rb_path} -x #{xml} --hourly ALL --add-timeseries-output-variable 'Zone People Occupant Count' --add-timeseries-output-variable 'Zone People Total Heating Energy'"
```
shorowit, line 214: Once we address the above issue, let's remove the `--hourly ALL` part here to verify it still works.

workflow/tests/hpxml_translator_test.rb
```diff
@@ -207,6 +207,29 @@ def test_run_simulation_timeseries_dst_and_utc
...
+    sql_path = File.join(File.dirname(xml), 'run', 'eplusout.sql')
+    assert(File.exist? sql_path)
+    csv_output_path = File.join(File.dirname(xml), 'run', 'results_annual.csv')
+    assert(File.exist? csv_output_path)
+    csv_output_path = File.join(File.dirname(xml), 'run', 'results_hpxml.csv')
+    assert(File.exist? csv_output_path)
+    csv_output_path = File.join(File.dirname(xml), 'run', 'results_timeseries.csv')
+    assert(File.exist? csv_output_path)
+
+    # Check TimeDST and TimeUTC exist
```
shorowit, line 227:
```suggestion
    # Check timeseries output variables exist
```

shorowit (commented):
I was just about to approve this when I found a problem.

workflow/run_simulation.rb
joseph-robertson: Just assume hourly timeseries (with no output types) when output variables are requesting but a frequency is not specified.

shorowit: Good idea. I'm going to switch it to `timestep` instead of `hourly`, which will obviously be the same thing for hourly simulations but seems like a better default to me.

shorowit (approved):

PR #1012, opened 2022-03-22
Additional properties updates

joseph-robertson:
## Pull Request Description

Checklist:
- [x] Convenience method for retrieving additional properties
- [x] Cleanup `tear_down_model` method
- [x] Add optional "|"-delimited argument to `BuildResidentialHPXML` for passing `<key>=<value>` pairs. Pairs are stored similar to how it's described below.

Questions:
- [ ] ~Maybe _any_ objects with additional properties should not be removed?~

## Checklist

Not all may apply:


shorowit:
My two cents here. I'm willing to be convinced otherwise, but I don't like the idea of _selectively_ removing model objects. That may lead to unexpected behavior in some circumstances.

To address the need at hand, I would suggest:
1. Add an `additional_properties` string argument to the BuildResHPXML measure where you can provide key/value pairs (e.g., "foo=1,bar,description=test").
2. Store these additional properties in the HPXML like so:
```
    <SoftwareInfo>
    <extension>
      <foo>1</foo>
      <bar/>
      <description>test</description>
    </extension>
  </SoftwareInfo>
```

This provides a nice benefit for any OS-HPXML user that wants to store additional arbitrary data in the HPXMLs they're producing.

shorowit:
Maybe it would be better to use a delimiter other than a comma. Perhaps "|" instead?

You also might need to be careful about certain characters like "<", ">", "/", etc. Not sure what would happen if you try to create XML elements with these characters.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -9,6 +9,7 @@
       <sch:assert role='ERROR' test='count(h:SoftwareInfo/h:extension/h:SimulationControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SoftwareInfo/extension/SimulationControl</sch:assert> <!-- See [SimulationControl] -->
       <sch:assert role='ERROR' test='count(h:SoftwareInfo/h:extension/h:HVACSizingControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SoftwareInfo/extension/HVACSizingControl</sch:assert> <!-- See [HVACSizingControl] -->
       <sch:assert role='ERROR' test='count(h:SoftwareInfo/h:extension/h:SchedulesFilePath) &gt;= 0'>Expected 0 or more element(s) for xpath: SoftwareInfo/extension/SchedulesFilePath</sch:assert>
+      <sch:assert role='ERROR' test='count(h:SoftwareInfo/h:extension/h:AdditionalProperties) &lt;= 1'>Expected 0 or 1 element(s) for xpath: SoftwareInfo/extension/AdditionalProperties</sch:assert>
```
shorowit, line 12: No need to add this. The only thing we're doing here is, for users using the BuildResHPXML measure, we're providing a way for them to store custom information. For users generating their own HPXMLs, they can already store custom information in any `extension` element.

docs/source/workflow_inputs.rst
```diff
@@ -281,6 +281,12 @@ If no default value is available, a warning will be issued.
   wood pellets  --              --             --
   ============  ==============  =============  =============
 
+HPXML Additional Properties
```
shorowit, line 284: Not needed for same reason as above.

tasks.rb
```diff
@@ -2174,7 +2175,9 @@ def set_measure_argument_values(hpxml_file, args, sch_args)
   end
 
   # Misc
-  if ['base-misc-defaults.xml'].include? hpxml_file
+  if ['base-misc-additional-properties.xml'].include? hpxml_file
+    args['additional_properties'] = 'foo=1|bar|description=test|comma=,|special=<|special2=>|special3=/|special4="\"'
```
shorowit, line 2179: I assume you wanted `<special4>\<special4>`, but this code produces `<special4>'\'<special4>`.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -966,6 +967,13 @@ def to_oga(doc)
           XMLHelper.add_element(extension, 'SchedulesFilePath', schedules_filepath, :string)
         end
       end
+      if (not @extension_properties.nil?)
```
shorowit, line 970: Need to check for empty hash too, so as not to produce a AdditionalProperties element with nothing inside.

shorowit (commented):
A few comments below -- I am going to address this in a commit shortly.

shorowit (approved):
Looks good. Will merge if the CI is still happy.

tasks.rb
```diff
@@ -2176,7 +2176,7 @@ def set_measure_argument_values(hpxml_file, args, sch_args)
 
   # Misc
   if ['base-misc-additional-properties.xml'].include? hpxml_file
-    args['additional_properties'] = 'foo=1|bar|description=test|comma=,|special=<|special2=>|special3=/|special4="\"'
+    args['additional_properties'] = 'LowIncome=false|Remodeled|Description=2-story home in Denver|comma=,|special=<|special2=>|special3=/|special4=\\'
```
joseph-robertson, line 2179: 👍

PR #1013, opened 2022-03-22
New "End Use: <Fuel>: Heating Heat Pump Backup" output

### LINKED ISSUE #1002: Create "End Use: <Fuel Type>: Supplemental Heating" workflow output
Split out from `End Use: <Fuel Type>: Heating`?
Different from "Secondary Heating"?

Notes from @shorowit:
I think it would be a moderate effort because there are some complications in OS-HPXML (DSE, need to handle HP backup systems that are either integrated or separate systems, etc.), so it would have to be done carefully. And this would presumably create a breaking change for software tools that use OS-HPXML/OS-ERI. But I do see the value in reporting this.

joseph-robertson:
## Pull Request Description

Closes #1002.

**Breaking change**: Splits out `End Use: <Fuel Type>: Heating Heat Pump Backup` from `End Use: <Fuel Type>: Heating` for both "integrated" and "separate" heat pump backup systems.

Questions:
- [ ] ~Add a `Load: Heating Heat Pump Backup: Delivered` output?~ Update: no.
- [ ] ~Add a `Peak Load: Heating Heat Pump Backup: Delivered` output?~ Update: no.

## Checklist

Not all may apply:

- [ ] ~EPvalidator.xml has been updated~


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -97,6 +97,7 @@ def self.apply_air_source_hvac_systems(model, runner, cooling_system, heating_sy
         htg_coil.setNominalCapacity(UnitConversions.convert(heating_system.heating_capacity, 'Btu/hr', 'W'))
         htg_coil.setName(obj_name + ' htg coil')
         htg_coil.additionalProperties.setFeature('HPXML_ID', heating_system.id) # Used by reporting measure
+        htg_coil.additionalProperties.setFeature('IsSupplementalCoil', heating_system.is_heat_pump_backup_system) # Used by reporting measure
```
shorowit, line 100: Trying to make the same change to all appropriate objects, and ensure new objects in the future are covered, feels a bit brittle. It seems like a more robust solution would be, in the `get_object_output_variables_by_key` method of the reporting measure, to retrieve the HPXML_ID that is already being set on the OSM object, and then check if the HPXML object with that ID is/has a backup system. (For a HPXML HeatingSystem object that serves as backup, you can use [this method](https://github.com/NREL/OpenStudio-HPXML/blob/d3a3bd73ba2c6a398fd722826b608ce5bbe20ba9/HPXMLtoOpenStudio/resources/hpxml.rb#L3239) to check.)

shorowit (commented):
Code changes look good and results seem reasonable at first glance. Will take a closer look at results tomorrow, but this should be very close to dropping in.

shorowit (approved):
Nice work, @joseph-robertson. Thanks!

PR #1036, opened 2022-04-13
Speed up utility bill measure

shorowit:
## Pull Request Description

Changes utility bill calculations to be monthly for improved runtime speed. Hourly calculations will be used only for detailed electric rates when they are implemented in the future.

## Checklist

### DIFF
--- a/ReportUtilityBills/measure.rb
+++ b/ReportUtilityBills/measure.rb
@@ -187,7 +187,9 @@ class ReportUtilityBills < OpenStudio::Measure::ReportingMeasure
   end
 
   def timeseries_frequency
-    return 'hourly'
+    # Use monthly for fossil fuels and simple electric rates
+    # Use hourly for detailed electric rates (URDB tariff or real time pricing)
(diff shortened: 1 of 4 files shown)


joseph-robertson:
This question is probably not specific to this PR, but what happens for partial month run periods? Fixed monthly costs are applied even if, say, only a single day from the month is in the run period?

shorowit:
> This question is probably not specific to this PR, but what happens for partial month run periods? Fixed monthly costs are applied even if, say, only a single day from the month is in the run period?

Good question, I thought about that as well when making these changes. Previously fixed charges were only applied for a full month, now they are applied for a partial month too. So the previous approach would have skipped fixed charges for a month that is missing a single day. I don't know that either approach is more correct, it's pretty arbitrary.

joseph-robertson:
Prorate the fixed charges?

shorowit:
That's a great idea. I'll do that.

joseph-robertson (approved):

PR #1056, opened 2022-05-02
Detailed HVAC setpoints (GEB), try 2

### LINKED ISSUE #844: Detailed HVAC setpoints (GEB)
Allow using an 8760 (or more frequent if running at shorter timesteps) schedule for heating/cooling setpoints. Similar to water heater setpoint capability in https://github.com/NREL/OpenStudio-HPXML/issues/635. Might need to be careful about other parts of the code (e.g., natural ventilation, ceiling fan temperature offset, etc.) that are based on setpoints.

joseph-robertson:
## Pull Request Description

Closes #844.

## Checklist


BuildResidentialHPXML/measure.rb
```diff
@@ -3387,6 +3392,14 @@ def self.set_header(hpxml, runner, args)
     hpxml.header.xml_generated_by = 'BuildResidentialHPXML'
     hpxml.header.transaction = 'create'
 
+    if args[:schedules_filepaths].is_initialized
+      hpxml.header.schedules_filepaths = []
+      schedules_filepaths = args[:schedules_filepaths].get.split(',').map(&:strip)
+      schedules_filepaths.each do |schedules_filepath|
+        hpxml.header.schedules_filepaths << schedules_filepath
+      end
```
shorowit, line 3400: @joseph-robertson Can't this be simply `hpxml.header.schedules_filepaths = args[:schedules_filepaths].get.split(',').map(&:strip)`?

docs/source/workflow_inputs.rst
```diff
@@ -158,6 +158,7 @@ They can be smooth schedules, or they can reflect real-world or stochastic occup
 Detailed schedule inputs are provided via CSV file(s) that should be referenced in the HPXML file as a ``/HPXML/SoftwareInfo/extension/SchedulesFilePath``.
 
 Occupancy related schedule columns must be normalized to MAX=1; that is, these schedules only define *when* energy is used, not *how much* energy is used.
+Thermostat setpoint schedule columns do not have these constraints.
```
shorowit, line 161: Thinking ahead to the WH GEB changes, I suggest creating a new column below called 'Units', which is either `frac` or `F`, and then making this change:

```suggestion
Columns with units of `frac` must be normalized to MAX=1; that is, these schedules only define *when* energy is used, not *how much* energy is used.
```

tasks.rb
```diff
@@ -370,6 +372,10 @@ def create_hpxmls
     'base-pv-battery-lifetime-model.xml' => 'base-pv-battery.xml',
     'base-pv-battery-garage.xml' => 'base-enclosure-garage.xml',
     'base-schedules-simple.xml' => 'base.xml',
+    'base-schedules-detailed-setpoints.xml' => 'base.xml',
```
shorowit, line 375: Should we create one more sample file (base-schedules-detailed-all.xml) that includes all possible schedule columns (occupancy and setpoints)?

shorowit (commented):
Looking good. I want to do a bit of manual testing still but here are some initial comments.

tasks.rb
joseph-robertson: In addition to `base-schedules-detailed-setpoints-10-mins.xml` which already does that?

shorowit: Are we looking at the same thing? That file only references the setpoints.csv. I want a sample file that references two schedule CSVs (setpoints and occupancy). Also, we should probably rename the smooth and stochastic CSV files to have an 'occupancy-' prefix (so, `occupancy-smooth.csv`, for example). I like that you named all the setpoint CSVs using a 'setpoints-' prefix.

joseph-robertson: See https://github.com/NREL/OpenStudio-HPXML/blob/detailed-hvac-setpoints-2/workflow/sample_files/base-schedules-detailed-setpoints-10-mins.xml#L14-L15.

shorowit (approved):

shorowit:
Thanks @joseph-robertson!

PR #1069, opened 2022-05-09
Additional boiler capabilities

### LINKED ISSUE #1068: Additional boiler capabilities
- [ ] Allow outdoor reset. Currently [always disabled](https://github.com/NREL/OpenStudio-HPXML/blob/e0c04fb52d6d96401a2fbac5909987c8531adf3a/HPXMLtoOpenStudio/resources/hvac.rb#L426-L430). (For some reason [our code only allows it for condensing boilers](https://github.com/NREL/OpenStudio-HPXML/blob/e0c04fb52d6d96401a2fbac5909987c8531adf3a/HPXMLtoOpenStudio/resources/hvac.rb#L505), but [this article](https://www.hpacmag.com/features/outdoor-reset/) suggests it can be used with non-condensing boilers as well.)
- [ ] Allow condensing vs non-condensing boiler efficiency curves. Currently hard-coded as [non-condensing](https://github.com/NREL/OpenStudio-HPXML/blob/e0c04fb52d6d96401a2fbac5909987c8531adf3a/HPXMLtoOpenStudio/resources/hvac.rb#L425). Should be optional and defaulted based on AFUE.

shorowit:
## Pull Request Description

Closes #1068.

## Checklist


shorowit:
The only major remaining TODO is to expose the boiler supply temperature as an optional input. There is an existing [SupplyTemperature](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Systems/HVAC/HVACDistribution/DistributionSystemType/HydronicDistribution/SupplyTemperature) element we can use. We could use defaults based on the hydronic distribution type (e.g., 180F in general, 160F for radiant ceiling, 120F for radiant floor). However there are a few complications:
- What defaults do we use if a condensing boiler? Perhaps 150F in general, but what about radiant ceiling/floor?
- The SupplyTemperature element is not available for fan coils, because we call them AirDistribution (so they can be ducted) and not HydronicDistribution. 😬 
- How does the supply temperature relate to the higher setpoint for a boiler with outdoor reset control? BEopt allowed both to be independently set, but the only option BEopt shipped with outdoor reset uses the same value for both. Does it make sense to allow independent values? Or should the supply temperature not be an input if there is outdoor reset control?

PR #1083, opened 2022-05-14
HVAC seasons improvements

### LINKED ISSUE #1076: Revisit heating/cooling season implementation
Why do we use an ideal air system outside of heating/cooling seasons? Maybe there was a good reason for this implementation approach, but it can result in the model, e.g., maintaining a heating setpoint outside the heating season, whereas the user would expect the heating setpoint to not have an effect outside the heating season.

shorowit:
## Pull Request Description

Closes #1076. We were previously using an ideal air system when modeling HVAC seasons, as it would prevent heating/cooling energy use outside the seasons while also avoiding unmet hours from being reported. However, this has the unintended effect of causing conditioned space to be maintained at setpoint outside the seasons as well.

This PR removes the ideal air system and lets the temperatures float outside the HVAC seasons. To continue to avoid unmet hours from being reported, it uses an EMS program with knowledge of the heating/cooling seasons. This PR also allows timeseries unmet hours to be reported.

(**Bonus**: This PR also relaxes the constraint that the heating/cooling seasons must span the entire year. Now there can be periods of the year without any space conditioning.)

Here's the before/after for an example home in Denver with heating season Nov 1 - May 1 and a cooling season Jun 1 - Oct 1:

image

Note to future developers: It's tempting to adjust the setpoints outside of the heating/cooling seasons to prevent unmet hours being reported. This is a dangerous idea. These setpoints are used by natural ventilation, Kiva initialization, and probably other things. I started to go down this path -- it's a rabbit hole and _not_ worth it.

## Checklist


shorowit:
@yzhou601 Could you test this in your pre-cooling simulations, to see if it performs similarly to your workaround of adjusting the heating setpoint?

yzhou601 (approved):
Ran test file on master branch:
image

Ran test file on this branch: 
image

The issue was fixed. 

A question related to the Unmet Hour reporting, so if running a sub-hourly simulation, does the timeseries Unmet Hours still describe in hourly basis, or is it the portion of the time step that the loads are not met? (eg. running 15-min simulation where there's a 0.2 unmet hour, does it mean 60 * 0.2 = 12 mins or does it mean 15 * 0.2 = 3 mins ?)

shorowit:
> A question related to the Unmet Hour reporting, so if running a sub-hourly simulation, does the timeseries Unmet Hours still describe in hourly basis, or is it the portion of the time step that the loads are not met? (eg. running 15-min simulation where there's a 0.2 unmet hour, does it mean 60 * 0.2 = 12 mins or does it mean 15 * 0.2 = 3 mins ?)

It's in units of hours, so it ought to be the former.

shorowit:
Thanks for testing, @yzhou601.

PR #1090, opened 2022-05-17
Purge `Constants.Auto`

### LINKED ISSUE #1072: Replace use of Constants.Auto in BuildResHPXML measure?
Should we replace all uses of Constants.Auto (typically used with required arguments) in the BuildResHPXML measure with optional arguments? It would be a breaking change, but it would simplify the inputs to the measure.

The Constants.Auto approach has some significant disadvantages:
- It requires numeric values to be entered as stings in the OS workflow (e.g., "4.0" instead of 4.0)
- It requires lots of (required) arguments to be specified in the OS workflow, even when they don't apply
- It's confusing as to why some arguments are required and use Constants.Auto while others are optional

So why do we do it?
- It's possible that it was simply borne out of how BEopt handled inputs, which was then preserved in the OS BEopt measures and now shows up here, or
- It's possible that there is a benefit to having required arguments for, e.g., ResStock (integrity checks?), or
- It's possible that there is some other benefit that we can't immediately remember.

cc @joseph-robertson

shorowit:
## Pull Request Description

Closes #1072.

BuildResidentialHPXML measure:
  - **Breaking change**: Replaces arguments using 'auto' for defaults with optional arguments of the appropriate data type. New `heat_pump_sizing_methodology` argument and new boolean `foo_present` arguments for lighting, appliances, etc.

ReportUtilityBills measure:
  - **Breaking change**: Replaces arguments using 'auto' for defaults with optional arguments of the appropriate data type.

## Checklist


shorowit:
Verified that you can still apply the measure by just specifying the HPXML path:

image

BuildResidentialHPXML/measure.rb
```diff
@@ -48,7 +48,7 @@ def description
 
   # human readable description of modeling approach
   def modeler_description
-    return ''
+    return "Note: OS-HPXML default values can be found in the OS-HPXML documentation or can be seen by using the 'apply_defaults' argument."
```
joseph-robertson, line 51: 👍

docs/source/workflow_inputs.rst
```diff
@@ -370,7 +370,7 @@ Building occupancy is entered in ``/HPXML/Building/BuildingDetails/BuildingSumma
   ========================================  ========  =====  ===========  ========  ====================  ========================
   Element                                   Type      Units  Constraints  Required  Default               Notes
   ========================================  ========  =====  ===========  ========  ====================  ========================
-  ``NumberofResidents``                     integer          >= 0         See [#]_  <number of bedrooms>  Number of occupants [#]_
+  ``NumberofResidents``                     double           >= 0         See [#]_  <number of bedrooms>  Number of occupants [#]_
```
joseph-robertson, line 373: Why double?

joseph-robertson (approved):

docs/source/workflow_inputs.rst
shorowit: Because the documentation was wrong.

- https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/BuildingSummary/BuildingOccupancy/NumberofResidents
- https://github.com/NREL/OpenStudio-HPXML/blob/ac525ddca5d5178e089bb1104e86edc59ddbde4e/HPXMLtoOpenStudio/resources/hpxml.rb#L1284

Does the stochastic generator support a non-integer? If not, it should probably round to the nearest integer.

shorowit: @rajeee Do you know if the stochastic generator supports a non-integer number of occupants?

Issue #920, opened 2021-11-15
Utility bill calculation measure


shorowit:
SAM SDK?
- bundled in energyplus

Alternatives?
- dGen (Pieter Gagnon)
- python with openstudio (pycall)
- python OpenStudio measure that calls SAM?
- EMS?
- https://github.com/NREL/EnergyPlus/issues/9154

joseph-robertson:
FYI The `UtilityBillCalculations` reporting measure last appeared in ResStock v2.0.0: https://github.com/NREL/resstock/tree/v2.0.0/measures/UtilityBillCalculations

joseph-robertson:
Checklist:
- [ ] dGen for selection of tariffs
- [ ] Need to look into ffi / fiddle
- [ ] Issues with PV (sellback rates, etc.) and use of `utilityrate3`, `utilityrate5`, etc.
- [ ] Issues with leap years

joseph-robertson:
On python measure release, Kyle says: "im not sure. i guess I need to talk to Tim about that. I'm wrestling with some technical issues with Ruby at the moment that I don't have a solution for. I think I'll be able to resolve but not sure"

joseph-robertson:
On ffi, Julien has created a test PR for adding it to the OS CLI: https://github.com/NREL/openstudio-gems/pull/52

joseph-robertson:
From https://openei.org/services/doc/rest/util_rates/?version=3:

OpenEI retired the version 1 API on January 2nd, 2014. Version 2 was released on May 4th, 2013, and version 3 was released on June 16th, 2014, at which point version 2 database updates ceased, with all updates now appearing only in version 3 or greater. Please contact the OpenEI Webmaster if you have any questions or concerns.

joseph-robertson:
Plan:

Step 1
- [ ] Create a new `ReportUtilityBills` measure
- [ ] Start with existing `UtilityBillCalculations` measure, at least in terms of arguments
- [ ] Start with simple electric bills, fossil fuel bills
- [ ] Support only latest version of URDB tariffs?
- [ ] Support only full annual utility bill calculations
- [ ] Need monthly output
  - [ ] Start with E+ consumption output (DSE would have different outputs)
  - [ ] TODO: see below
- [ ] Add unit tests
- [ ] Enable "auto" arguments for state averages (Electricity.csv, Gas.csv, etc.)

Steps 2 and 3
- [ ] Detailed bills (URDB considerations, etc.)
[...]

PR #1092, opened 2022-05-18
Utility bill calculation measure, phase 2

joseph-robertson:
## Pull Request Description

Addresses #920. Adds detailed calculations for tiered, time-of-use, or real-time pricing electric rates.

## Checklist


ReportUtilityBills/resources/util.rb
```diff
@@ -125,3 +125,35 @@ def self.calculate_monthly_prorate(header, month)
...
+  puts 'Selecting residential rates...'
+  residential_rates = []
+  rates.each do |rate|
+    next if !rate.keys.include?('_id')
+    next if !rate['_id'].keys.include?('$oid')
+    next if !rate.keys.include?('sector')
+    next if rate['sector'] != 'Residential'
+
+    residential_rates << rate
+  end
```
shorowit, line 144: Nice. Can you also incorporate skipping rates with [these keywords](https://github.com/NREL/BEopt2/blob/c84d36bfe42a087e5ff6a6cf4756b2fc31ca0460/Modeling/Utilities/get_utility_rates.py#L109-L121)?

shorowit (commented):
A few initial comments:

- Lets further downselect tariffs of interest (potentially to include as a compressed archive). Lets remove any tariff with an `enddate`. Let's also remove any tariff with non-zero demand charges. **There's only a handful; I just left them in.** How many tariffs does that put us at? **5,949.**
- Move sample files to tests directory. Update Sample Real-Time Pricing file so that it reflects the updated JSON schema.
- Rename JSON files to "[utility] - [name].json" or "[utility] - [name] (Effective: [startdate]).json". **Can't have special characters in file names.**

ReportUtilityBills/resources/util.rb
```diff
@@ -125,3 +514,113 @@ def self.calculate_monthly_prorate(header, month)
...
+              'yard light',
+              'security light',
+              'lumens',
+              'watt hps',
+              'incandescent',
+              'halide',
+              'lamps',
+              '[partial]',
+              'rider',
+              'irrigation']
```
shorowit, line 541: add "grain" to the list, I found a few more agricultural rates.

ReportUtilityBills/resources/util.rb
```diff
@@ -125,3 +514,113 @@ def self.calculate_monthly_prorate(header, month)
...
+
+  puts 'Parsing CSV...'
+  rates = CSV.read(filepath, headers: true)
+  rates = rates.map { |d| d.to_hash }
+
+  puts 'Selecting residential rates...'
```
shorowit, line 547: I looked at the JSONs sorted by file size and found a few that were really small and basically have no utility rate info in them. Can you filter out tariffs that don't have the minimum fields needed to perform the calculation?

ReportUtilityBills/measure.rb
```diff
@@ -298,54 +301,42 @@ def report_runperiod_output_results(runner, utility_bills, bill_scenario_name)
...
+          if tariff.keys.include?(:realtimepricing)
+            rate.fixedmonthlycharge = tariff[:fixedmonthlycharge] if tariff.keys.include?(:fixedmonthlycharge)
+            rate.realtimeprice = tariff[:realtimepricing]
 
-              rate.energyratestructure = tariff[:energyratestructure] if tariff.keys.include?(:energyratestructure)
-              rate.energyweekdayschedule = tariff[:energyweekdayschedule] if tariff.keys.include?(:energyweekdayschedule)
-              rate.energyweekendschedule = tariff[:energyweekendschedule] if tariff.keys.include?(:energyweekendschedule)
+          else
+            rate.fixedmonthlycharge = tariff[:fixedmonthlycharge] if tariff.keys.include?(:fixedmonthlycharge)
+            rate.fixedmonthlycharge = tariff[:fixedchargefirstmeter] if tariff.keys.include?(:fixedchargefirstmeter)
```
shorowit, line 329: I'm confused, why do you look for two different fields? The [documentation](https://openei.org/services/doc/rest/util_rates/?version=3#response-fields) only lists `fixedmonthlycharge`, what is this other one?

ReportUtilityBills/measure.rb
```diff
@@ -298,54 +301,42 @@ def report_runperiod_output_results(runner, utility_bills, bill_scenario_name)
...
-              rate.demandratestructure = tariff[:demandratestructure] if tariff.keys.include?(:demandratestructure)
-              rate.demandweekdayschedule = tariff[:demandweekdayschedule] if tariff.keys.include?(:demandweekdayschedule)
-              rate.demandweekendschedule = tariff[:demandweekendschedule] if tariff.keys.include?(:demandweekendschedule)
+            rate.energyratestructure = tariff[:energyratestructure] if tariff.keys.include?(:energyratestructure)
+            rate.energyweekdayschedule = tariff[:energyweekdayschedule] if tariff.keys.include?(:energyweekdayschedule)
+            rate.energyweekendschedule = tariff[:energyweekendschedule] if tariff.keys.include?(:energyweekendschedule)
 
-              rate.flatdemandstructure = tariff[:flatdemandstructure] if tariff.keys.include?(:flatdemandstructure)
+            if tariff.keys.include?(:demandratestructure) || tariff.keys.include?(:flatdemandstructure)
+              warnings << 'Demand charges are not currently supported when calculating detailed utility bills.'
```
shorowit, line 338: Does this warning prevent the utility bills from being calculated? I would think it should since it's an incomplete calculation when demand charges are excluded.

Can you exclude JSONs from the tar.gz that have these fields?

ReportUtilityBills/tests/Sample Tiered Rate.json
```diff
@@ -0,0 +1,659 @@
+{
+  "items": [
+    {
+      "name": "Sample Tiered Rate",
+      "sector": "Residential",
+      "description": "Sample file defined by BEopt.",
+      "fixedmonthlycharge": 9,
+      "fixedchargeunits": "$/month",
+      "flatratebuy": 0,
+      "flatratefueladj": 0,
```
shorowit, line 10: I'm confused by some of the fields in these sample files. For example, these three fields aren't in the [documentation](https://openei.org/services/doc/rest/util_rates/?version=3#response-fields), so why are they here? We need to make these sample files consistent with version 3 rates from URDB.

Does your code actually use these fields?

ReportUtilityBills/measure.rb
joseph-robertson: I saw `fixedchargefirstmeter` in at least one tariff file.

ReportUtilityBills/tests/Sample Tiered Rate.json
joseph-robertson: I believe it uses `flatratebuy` and `flatratefueladj`.

ReportUtilityBills/measure.rb
joseph-robertson: Hmm perhaps usurdb.csv download is version 1? For example, column `fixedchargefirstmeter` exists and is documented. However, version 3's `fixedmonthlycharge` does not exist as a column.

shorowit: If we need to support `fixedchargefirstmeter`, the code above is definitely not correct. It's always being treated as a *monthly* fixed charge, but there is a separate `fixedchargeunits` which says whether it's daily/monthly/annual.

ReportUtilityBills/tests/Sample Tiered Rate.json
joseph-robertson: I removed `flatratebuy`, `flatratefueladj`, and `fixedchargeunits` from the "Sample *.json" test files. I updated the `self.detailed_electric` method to no longer use `flatratebuy` or `flatratefueladj`.

ReportUtilityBills/measure.rb
joseph-robertson: I updated the `process_usurdb` method to map to `fixedmonthlycharge`, `minmonthlycharge`, and `annualmincharge` fields based on values in `fixedchargeunits` and `minchargeunits`.

ReportUtilityBills/measure.rb
joseph-robertson: Yes, it does prevent the utility bill (for that scenario) from being calculated.

I _could_. Although keep in mind that we have a unit test that operates on a tariff with demand charges.

shorowit: 👍 that it prevents the utility bill from being calculated and 👍 that you have a unit test.

I'd prefer that the tar.gz exclude these JSONs. It will be confusing to a BEopt user if we don't.

ReportUtilityBills/measure.rb
shorowit: Sounds good. Do our calculations operate on all three (fixedmonthlycharge, minmonthlycharge, and annualmincharge) of those fields? BEopt has inputs for the first two, but not the latter... maybe I should add it.

ReportUtilityBills/resources/util.rb
shorowit: Maybe we can add a unit test around one of these files too.

joseph-robertson: I filtered out those without `energyweekdayschedule`, `energyweekendschedule`, or `energyratestructure`.

ReportUtilityBills/measure.rb
joseph-robertson: I think annualmincharge is used for "California-style calculations". See the "TODO" items in `util.rb`. I posted this as a question at the top of this PR.

ReportUtilityBills/resources/util.rb
joseph-robertson: I added `test_warning_missing_required_fields`.

ReportUtilityBills/measure.rb
joseph-robertson: I'll exclude packaging rates with any of the following fields: `demandweekdayschedule`, `demandweekendschedule`, `demandratestructure`, or `flatdemandstructure`.

shorowit:
- [ ] Identify differences between OpenEI v3 and v7; if not very different, update to v7. If significantly different, stick with v3.
- [ ] Specify what version of OpenEI files we expect in the measure arg (and documentation?).

ReportUtilityBills/tests/utility_bills_test.rb
```diff
@@ -363,66 +412,1081 @@ def test_monthly_prorate
...
+      tmpdir = Dir.tmpdir
+      tmpfile = Tempfile.new(['rate', '.json'], tmpdir)
+      tmp_path = tmpfile.path.to_s
+
+      File.open(tmp_path, 'wb') do |f|
+        f.print entry.read
+
+        @hpxml.header.utility_bill_scenarios.each do |utility_bill_scenario|
+          utility_bill_scenario.elec_tariff_filepath = tmp_path
+          fuels = _load_timeseries('../tests/PV_None.csv', utility_bill_scenario)
```
shorowit, line 1403: Why can't you do this once up front?

joseph-robertson: I am. I put a "skip" in there to prevent it from running on CI.

joseph-robertson: Oh, sorry. I re-read. You mean literally the `_load_timeseries` line?

shorowit: Yes. Should only need to call it a single time in the method, no?

ReportUtilityBills/resources/util.rb
```diff
@@ -125,3 +590,164 @@ def self.calculate_monthly_prorate(header, month)
...
+    # ignore rates without a "rate" key
+    next if rate['energyratestructure'].collect { |r| r.collect { |s| s.keys.include?('rate') } }.flatten.any? { |t| !t }
+
+    # ignore rates with a "sell" key
+    next if rate['energyratestructure'].collect { |r| r.collect { |s| s.keys } }.flatten.uniq.include?('sell')
+
+    # ignore rates with units other than "kWh"
+    # next if rate['energyratestructure'].collect { |r| r.collect { |s| s.keys.include?('unit') } }.flatten.any? { |t| !t }
+    # next if rate['energyratestructure'].collect { |r| r.collect { |s| s['unit'] == 'kWh' } }.flatten.any? { |t| !t }
+    rate['energyratestructure'].collect { |r| r.collect { |s| s['unit'] = 'kWh' } } # this field is suddenly gone, so assume kWh?
```
shorowit, line 716: We should find out the answer to this before merging. I noticed that there are ~1000 additional JSONs now, so it's possible that some of them were being filtered out by this previously (and should continue to be filtered out).

ReportUtilityBills/resources/util.rb
```diff
@@ -125,3 +590,164 @@ def self.calculate_monthly_prorate(header, month)
...
+    end
+
+    # map schedules and structures
+    structures = {}
+    rate.each do |k, v|
+      if ['eiaid'].include?(k)
+        rate[k] = Integer(Float(v))
+      elsif k.include?('schedule')
+        # all of a sudden some fields have an "L" character (?)
+        rate[k] = eval(v.gsub('L', '')) # arrays
```
shorowit, line 659: Let's see if we can find out what's happening here too.

BuildResidentialHPXML/measure.rb
```diff
@@ -3557,6 +3562,21 @@ def self.set_header(hpxml, args)
...
+      if args[:utility_bill_electricity_filepaths].is_initialized
+        bills_electricity_filepaths = args[:utility_bill_electricity_filepaths].get.split(',').map(&:strip)
+        bills_electricity_filepaths.each_with_index do |bills_electricity_filepath, i|
+          begin
+            FilePath.check_path(bills_electricity_filepath,
+                                File.dirname(hpxml_path),
+                                'Tariff File')
+          rescue
+            bills_electricity_filepaths[i] = nil
+          end
```
shorowit, line 3574: I don't like the idea of simply ignoring any file paths we don't find. Also, we don't do this for any other similar input in this measure (schedules_filepaths, emissions_electricity_values_or_filepaths, etc.), so we should be consistent.

I would propose just passing the bills_electricity_filepath to the HPXML file without checking anything, and let the downstream HPXMLtoOpenStudio measure take care of checking the path. It looks like there are a few code changes related to making the `hpxml_path` available in this method, so that all could be reverted.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -111,6 +111,9 @@
     <sch:rule context='/h:HPXML/h:SoftwareInfo/h:extension/h:UtilityBillScenarios/h:UtilityBillScenario/h:UtilityRate[h:FuelType="electricity"]'>
       <sch:assert role='ERROR' test='count(h:FixedCharge) &lt;= 1'>Expected 0 or 1 element(s) for xpath: FixedCharge</sch:assert>
       <sch:assert role='ERROR' test='count(h:MarginalRate) &lt;= 1'>Expected 0 or 1 element(s) for xpath: MarginalRate</sch:assert>
+      <sch:assert role='ERROR' test='count(h:TariffFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: TariffFilePath</sch:assert>
+      <sch:assert role='ERROR' test='count(h:FixedCharge) + count(h:TariffFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: FixedCharge | TariffFilePath</sch:assert>
+      <sch:assert role='ERROR' test='count(h:MarginalRate) + count(h:TariffFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: MarginalRate | TariffFilePath</sch:assert>
```
shorowit, line 116: Isn't there a lot of redundancy here? Seems like all that's needed is:

```suggestion
      <sch:assert role='ERROR' test='count(h:FixedCharge) + count(h:TariffFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: FixedCharge | TariffFilePath</sch:assert>
      <sch:assert role='ERROR' test='count(h:MarginalRate) + count(h:TariffFilePath) &lt;= 1'>Expected 0 or 1 element(s) for xpath: MarginalRate | TariffFilePath</sch:assert>
```

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -284,7 +284,7 @@ def self.apply_utility_bill_scenarios(runner, hpxml)
     hpxml.header.utility_bill_scenarios.each do |scenario|
       hpxml_doc = hpxml.to_oga if hpxml_doc.nil?
 
-      if HPXML::has_fuel(hpxml_doc, HPXML::FuelTypeElectricity)
+      if HPXML::has_fuel(hpxml_doc, HPXML::FuelTypeElectricity) && scenario.elec_tariff_filepath.nil?
```
shorowit, line 287: Can you update the tests in test_defaults.rb to demonstrate that when there's a tariff file provided, the defaults below are not applied?

docs/source/workflow_inputs.rst
```diff
@@ -318,6 +318,10 @@ Electricity Rates
 
 For each scenario, electricity rates can be optionally entered as an ``/HPXML/SoftwareInfo/extension/UtilityBillScenarios/UtilityBillScenario/UtilityRate``.
 
+**Simple**
+
+If the utility rate type is simple, utility bills are calculated based on fixed charges and a marginal rate.
```
shorowit, line 323: We should probably move away from the simple vs detailed terminology. Yes, the tariff JSON files can be detailed, but they can be simple flat rates too. I will plan to address this.

ReportUtilityBills/tests/Sample Flat Rate Min Annual Charge.json
```diff
@@ -0,0 +1,649 @@
+{
+  "items": [
+    {
+      "name": "Sample Tiered Rate",
+      "sector": "Residential",
+      "description": "Sample file defined by BEopt.",
+      "fixedchargeunits": "$/month",
+      "fixedchargefirstmeter": 8,
+      "minchargeunits": "$/year",
+      "mincharge": 10000,
```
shorowit, line 10: I'm debating whether to make the min charge values (annual and monthly) more realistic. This is a note to self.

ReportUtilityBills/tests/utility_bills_test.rb
```diff
@@ -337,55 +387,1210 @@ def test_monthly_prorate
...
+      _check_bills(expected_bills, actual_bills)
+    end
+  end
+
+  def test_detailed_flat_pv_1kW_net_metering_user_excess_rate_min_monthly_charge
+    @hpxml.header.utility_bill_scenarios[-1].elec_tariff_filepath = '../../ReportUtilityBills/tests/Sample Flat Rate Min Monthly Charge.json'
+    @hpxml.pv_systems.each { |pv_system| pv_system.max_power_output = 1000.0 / @hpxml.pv_systems.size }
+    @hpxml.header.utility_bill_scenarios.each do |utility_bill_scenario|
+      actual_bills = _bill_calcs(@fuels_pv_1kw_detailed, @hpxml.header, @hpxml.pv_systems, utility_bill_scenario)
+      @expected_bills['Test: Electricity: Marginal (USD)'] = 1281
```
shorowit, line 516: I don't love that the minimum charges increase the "Marginal" values. Ideally we would rename "Fixed" to "Charges & Fees" or something similar, and have the minimum charges apply to this value instead. (Fixed Charges and Grid Connection Fees already apply to this value.) On the fence as to whether to do this. This is a note to self.

shorowit (commented):
This is looking good. I made a couple comments for @joseph-robertson to address and a couple comments for me to think more about.

docs/source/workflow_inputs.rst
shorowit: On second thought, it's pervasively used everywhere so I will just attempt to clarify what simple vs detailed means, as opposed to making lots of changes.

ReportUtilityBills/tests/utility_bills_test.rb
shorowit: We should really also use "Energy" instead of "Marginal", as is the case in BEopt.

shorowit (approved):
Thanks, @joseph-robertson! This will merge in if the CI comes back green.

PR #1150, opened 2022-08-05
Commercial water heaters

### LINKED ISSUE #654: Commercial water heater
Accommodate commercial water heaters and allow for Thermal Efficiency (TE) and Standby Loss (SL) inputs. Refer to combi boiler model for SL -> tank UA methodology?

shorowit:
## Pull Request Description

Closes #654. Adds ability to model commercial water heaters.

**Breaking Change**: Replaces `StandbyLoss` with `StandbyLoss[Units="F/hr"]/Value` for indirect water heaters.

## Checklist

### DIFF
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -1762,7 +1762,7 @@ class BuildResidentialHPXML < OpenStudio::Measure::ModelMeasure
     arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('water_heater_standby_loss', false)
     arg.setDisplayName('Water Heater: Standby Loss')
     arg.setDescription('The standby loss of water heater. Only applies to space-heating boilers. If not provided, the OS-HPXML default is used.')
-    arg.setUnits('deg-F/hr')
+    arg.setUnits(HPXML::UnitsDegFPerHour)
     args << arg
 
     arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('water_heater_jacket_rvalue', false)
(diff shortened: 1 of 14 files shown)


jmaguire1:
Some answers to the TODO questions:

**What are the thermal efficiency cutoffs for power vent and condensing?**
We only really care about this for capturing how much of the losses go to the space. Even minimum efficiency commercial gas WHs have a flue damper (likely required by law, although I haven't confirmed that), which in terms of space conditioning impacts is more like a power vent. There's a really clear cutoff in the AHRI data between TEs of 0.82 and 0.91 between condensing and noncondensing equipment, so I'd propose treat anything with a TE > 0.9 as "condensing" and everything with a TE < 0.9 as "power vent".
**For tankless, what do we do about performance adjustment?**
The performance adjustment we apply now (an 8% derate) is a function of the draw profile more than anything else. Since we're going to be subjecting these to the same draw profile regardless of whether it's a residential or commercial project, I'd suggest consistency and keeping the 8% derate. Of course someday we could implement an explicit tankless model and not have to worry about this, but I think in terms of annual gas consumption we're talking +/- a few %.
**Double-check DOE test conditions relative to RESNET spreadsheet.**
Residential and commercial products are tested under different conditions, and what's in the spreadsheet right now looks like the residential conditions, not commercial. In the commercial test procedure the mains temperature is 70 F (not 58) and ambient air temperature is 75 F (not 67.5). WH setpoint temperature is also 140 F (not 125).
**How do we calculate tank UA from %/hr?**
The [CFR ](https://www.ecfr.gov/current/title-10/part-431/subpart-G) actually has equations for this specified. Interestingly, even when testing a gas product the equations in the CFR first calculate the losses as %/hr and then convert to btu/hr using this equation: SL (Btu per hour) = S (% per hour) × 8.25 (Btu/gal- °F) × Measured Volume (gal) × 70 ( °F). This equation came from section 7.2.6.2 of Appendix A if anyone is curious.
**Why does AHRI have non-zero standby losses for instantaneous tankless water heaters?**
Definitely wasn't expecting this! It turns out the standard for commercial WHs does actually require this to be measured (along with the heat exchanger volume). It seems like they do this test by running some steady state draws at the specified outlet temperature (140 F), and then let the system sit idle until the measured water temperature is 35 F lower than it started at. In terms of what this means for our models: not much. Any losses when the tankless is operating are subsumed into our 8% derate, so I don't think we need to do anything special here. If we had a more explicit tankless model then this would likely be an input to our model, but as it stands now nothing special needs to be done about this.
   
@shorowit : I think that's the current list, but if any more questions come up let me know!

PR #1154, opened 2022-08-15
Aspect ratio fix

joseph-robertson:
## Pull Request Description

Appears that FB vs LR wall lengths were being flipped (for single-family attached and multifamily, not single-family-detached)  relative our definition of aspect ratio (FB/LR). This is caught by comparing the adiabatic wall length (right side of unit) to the wall length containing the (front) door.

## Checklist


workflow/sample_files/base-bldgtype-multifamily-adjacent-to-multifamily-buffer-space.xml
```diff
@@ -107,7 +107,7 @@
             <WallType>
               <WoodStud/>
             </WallType>
-            <Area>685.9</Area>
+            <Area>783.8</Area>
```
shorowit, line 110: If we proceed with this fix, I suggest you change the aspect ratio provided to generate these sample files so that they don't change. (The sample files were created prior to the Build measure existing and being used to generate them.)

joseph-robertson: You mean change the aspect ratio to 1 / 1.8 = 0.6, or whatever value produces unchanged sample files?

shorowit: Yep.

workflow/sample_files/base-bldgtype-single-family-attached-2stories.xml
```diff
@@ -183,7 +183,7 @@
             <WallType>
               <WoodStud/>
             </WallType>
-            <Area>587.9</Area>
+            <Area>587.7</Area>
```
shorowit, line 186: Still some changes here, which might cause downstream tests (e.g., in OS-ERI) to fail. Can you increase the number of sig fig in the aspect ratio?

shorowit: Ha! Nevermind, you saw it too.

HPXMLtoOpenStudio/tests/test_airflow.rb
```diff
@@ -546,7 +546,7 @@ def test_infiltration_compartmentalization_area
     # Test multifamily
     hpxml = HPXML.new(hpxml_path: File.absolute_path(File.join(sample_files_dir, 'base-bldgtype-multifamily.xml')))
     total_area, exterior_area = hpxml.compartmentalization_boundary_areas
-    assert_in_delta(686, exterior_area, 1.0)
+    assert_in_delta(784, exterior_area, 1.0)
```
shorowit, line 549: I assume some of these changes to expected values are no longer needed.

shorowit:
@aspeake Can you confirm that you agree with this change?

aspeake:
> @aspeake Can you confirm that you agree with this change?

Yes, I agree with aligning the MF/SFA wall area calcs with SFD and with the definition of the `aspect_ratio` argument. To maintain our same approach in ResStock, I think just a simple update to the MF/SFA aspect_ratio will need to be accompanied with this PR.

shorowit (approved):
Looks good to me.

PR #1164, opened 2022-08-25
DST w/ DView

### LINKED ISSUE #1160: DST w/ DView
The timeseries outputs in DView appear to be off by 1 hour when the model has DST. Look at, e.g., daily natural ventilation outputs for base-enclosure-windows-natural-ventilation-availability.xml, where the last hour of the days w/ natural ventilation show up in the following day's outputs (e.g., days w/o natural ventilation are not zero). The problem goes away if DST is disabled in the HPXML file.

We should confirm the issue by looking at outputs other than natural ventilation, since natural ventilation uses EMS and that adds complexity to the outputs.

joseph-robertson:
## Pull Request Description

Closes #1160.

## Checklist

- [x] ~Changelog has been updated~

### DIFF
--- a/ReportSimulationOutput/tests/output_report_test.rb
+++ b/ReportSimulationOutput/tests/output_report_test.rb
@@ -1054,22 +1054,24 @@ class ReportSimulationOutputTest < MiniTest::Test
     args_hash = { 'hpxml_path' => File.join(File.dirname(__FILE__), '../../workflow/sample_files/base.xml'),
                   'output_format' => 'csv_dview',
                   'timeseries_frequency' => 'timestep',
-                  'include_timeseries_fuel_consumptions' => true }
+                  'include_timeseries_fuel_consumptions' => true,
+                  'include_timeseries_weather' => true }
     annual_csv, timeseries_csv = _test_measure(args_hash)
     assert(File.exist?(annual_csv))
     assert(File.exist?(timeseries_csv))
-    assert_equal('wxDVFileHeaderVer.1', CSV.readlines(timeseries_csv)[0][0].strip)
-  end
+    timeseries_csv = CSV.readlines(timeseries_csv)
+    assert_equal('wxDVFileHeaderVer.1', timeseries_csv[0][0].strip)
 
-  def test_timeseries_for_dview_dst_disabled
-    args_hash = { 'hpxml_path' => File.join(File.dirname(__FILE__), '../../workflow/sample_files/base-simcontrol-daylight-saving-disabled.xml'),
-                  'output_format' => 'csv_dview',
-                  'timeseries_frequency' => 'timestep',
-                  'include_timeseries_fuel_consumptions' => true }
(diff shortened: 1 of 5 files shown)


ReportSimulationOutput/tests/output_report_test.rb
```diff
@@ -1054,8 +1054,18 @@ def test_timeseries_for_dview
...
+    assert(File.exist?(annual_csv))
+    assert(File.exist?(timeseries_csv))
+    assert_equal('wxDVFileHeaderVer.1', CSV.readlines(timeseries_csv)[0][0].strip)
+  end
+
+  def test_timeseries_for_dview_dst_disabled
+    args_hash = { 'hpxml_path' => File.join(File.dirname(__FILE__), '../../workflow/sample_files/base-simcontrol-daylight-saving-disabled.xml'),
+                  'output_format' => 'csv_dview',
+                  'timeseries_frequency' => 'timestep',
+                  'include_timeseries_fuel_consumptions' => true }
```
shorowit, line 1068: These tests are pretty unsatisfying. They would pass even if you haven't made any code changes. How about comparing the timeseries_csv rows between base.xml and base-simcontrol-daylight-saving-disabled.xml and asserting that A) the data is identical for the first hour of the year and B) the data is offset by 1 hour in the middle of the year? Or something else that demonstrates the output data is correct.

shorowit (commented):
The implementation looks reasonable.

A couple corner cases I'm curious about:
- [x] Have you tested that there is a 1-hr shift even when the timestep is shorter than an hour? Answer: I tried running base-simcontrol-timestep-10-mins.xml with and without the code change; there is a 1-hr shift.
- [x] Does everything work correctly if the simulation run period is shorter than a year? E.g., The simulation starts before DST start and ends before DST end, or the simulation starts after DST start and ends before DST end. Answer: I tried modifying start/end dates in base-simcontrol-runperiod-1-month.xml; now working.

ReportSimulationOutput/tests/output_report_test.rb
joseph-robertson: I don't think it's that easy. There are energy use differences between base.xml and base-simcontrol-daylight-saving-disabled.xml. Perhaps we compare something that doesn't change, like `Weather| Drybulb Temperature`?

shorowit: Great points. Yes, using a weather variable should work well.

joseph-robertson:
@shorowit We only want to shift csv_dview output when the timeseries frequency is timestep or hourly, right?

shorowit:
> @shorowit We only want to shift csv_dview output when the timeseries frequency is timestep or hourly, right?

Yes, that's right, good catch.

shorowit (approved):
Confirmed that it fixes the original issue. Looks good to me. Thanks @joseph-robertson!

PR #1175, opened 2022-09-12
Use E+ 22.2 Incident solar multiplier

### LINKED ISSUE #1051: Use E+ simple shading object
This will simplify our code and eliminate shading surfaces/calculations, thus providing some performance improvement (especially for HPXML files with lots of windows). Might be available in E+ 22.2?

shorowit:
## Pull Request Description

Closes #1051.

## Checklist


shorowit:
Component load residuals are much higher when using this object. Need to investigate by looking at window heat transfer output variables. Maybe one of them is not properly updating based on this object?

shorowit:
I spent a little time today investing further. Documenting what I found...

From looking at hourly output, it seems like this IncidentSolarMultiplier (ISM) improves the accuracy of the window heat transfer relative to our current approach (of using a shading surface and modifying the view factor to adjust the ground reflectance). If you compare a home w/ no shading to a home w/ 100% shading, the ISM results (red line) perfectly matches the no shading results (blue line) for all hours of the year where there is no sun. Meanwhile, the current approach (orange line) is consistently offset from the ISM results for every hour of the year. Not really sure why that is -- maybe the window convection is somehow being changed in the current approach?

image

So that's a good sign, but other results are strange. The difference in window heat transfer between the ISM approach and current approach is ~3.5 MBtu. Our reported window component loads also show a net change of approximately 3.5 MBtu (see below). But the heating/cooling loads (and thus energy use) change by _far more than that_. Thus, the sum of component loads is nowhere close to the delivered loads.

image

So:
1. Need to track down why the delivered loads change so much more than the component loads. Should probably run with an ideal air system to further simplify the models.
2. Ideally would track down why the original approach is offset from the ISM approach for every hour of the year.

shorowit:
HPXML files used:
[shading0.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/11410217/shading0.xml.txt)
[shading1.xml.txt](https://github.com/NREL/OpenStudio-HPXML/files/11410218/shading1.xml.txt)

shorowit:
Seems like the ISM has a bug: https://github.com/NREL/EnergyPlus/issues/10001

shorowit:
It seems like the original approach may be excluding radiation exchange between the window and the sky, which is why it's offset for every hour of the year. According to [the transmittance schedule documentation](https://bigladdersoftware.com/epx/docs/22-2/input-output-reference/group-thermal-zone-description-geometry.html#field-transmittance-schedule-name): "Shading devices are assumed to be opaque to long-wave radiation no matter what the solar transmittance value is."

shorowit:
The E+ bug should be fixed for the next release, which should finally allow us to start using this new object.

shorowit:
- [ ] Remove https://github.com/NREL/OpenStudio-HPXML/blob/43abcae2902cc4b266545198ffee0e1ed89da549/workflow/tests/hpxml_translator_test.rb#L566

PR #1212, opened 2022-11-07
Average resilience hours

joseph-robertson:
## Pull Request Description

Add "Resilience Hours: Battery (hr)" to `results_annual.csv`.

## Checklist

### DIFF
--- a/ReportSimulationOutput/measure.rb
+++ b/ReportSimulationOutput/measure.rb
@@ -193,6 +193,7 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     all_outputs << @peak_loads
     all_outputs << @component_loads
     all_outputs << @hot_water_uses
+    all_outputs << @average_resilience
 
     output_names = []
     all_outputs.each do |outputs|
@@ -2149,6 +2150,13 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     attr_accessor(:ems_variable)
   end
 
+  class AverageResilience < BaseOutput
+    def initialize
+      super()
+    end
+    attr_accessor()
+  end
+
   class IdealLoad < BaseOutput
     def initialize(variables: [])
       super()
@@ -2456,6 +2464,14 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
       comp_load.name = "Component Load: #{load_type.gsub(': Delivered', '')}: #{comp_load_type}"
       comp_load.annual_units = 'MBtu'
       comp_load.timeseries_units = 'kBtu'
+    end
+
+    # Average Resilience
+    @average_resilience = {}
+    [AR::Battery].each do |type|
+      @average_resilience[type] = AverageResilience.new
+      @average_resilience[type].name = "Average Resilience: #{type}"
+      @average_resilience[type].annual_units = 'hr'
     end
 
     # Unmet Hours
--- a/ReportSimulationOutput/measure.rb
+++ b/ReportSimulationOutput/measure.rb
@@ -193,7 +193,7 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     all_outputs << @peak_loads
     all_outputs << @component_loads
     all_outputs << @hot_water_uses
-    all_outputs << @average_resilience
+    all_outputs << @resilience_hours
 
     output_names = []
     all_outputs.each do |outputs|
@@ -871,6 +871,25 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
       @totals[TE::Net].timeseries_output = @totals[TE::Total].timeseries_output.zip(outputs[:elec_prod_timeseries]).map { |x, y| x + y * unit_conv }
     end
 
+    # Resilience Hours
+    @resilience_hours.each do |key, resilience_hour|
+      next unless key == RHT::Battery
+
+      resilience_hour.variables.map { |v| v[0] }.uniq.each do |sys_id|
+        keys = resilience_hour.variables.select { |v| v[0] == sys_id }.map { |v| v[1] }
+        vars = resilience_hour.variables.select { |v| v[0] == sys_id }.map { |v| v[2] }
+
+        batt_kwh = @hpxml.batteries[0].usable_capacity_kwh
+        batt_kw = @hpxml.batteries[0].rated_power_output
+        batt_roundtrip_efficiency = 0.95
+        batt_soc_kwh = get_report_variable_data_timeseries(keys, vars, 1, 0, timeseries_frequency)
+        crit_load = outputs[:elec_net_timeseries]
+
+        resilience_hours = get_resilience_hours(batt_kwh, batt_kw, batt_roundtrip_efficiency, batt_soc_kwh, crit_load)
+        resilience_hour.annual_output = resilience_hours.sum / resilience_hours.size
+      end
+    end
+
     # Zone temperatures
     if include_timeseries_zone_temperatures
       zone_names = []
@@ -1237,6 +1256,10 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     results_out << [line_break]
     @hot_water_uses.each do |_hot_water_type, hot_water|
       results_out << ["#{hot_water.name} (#{hot_water.annual_units})", hot_water.annual_output.to_f.round(n_digits - 2)]
+    end
+    results_out << [line_break]
+    @resilience_hours.each do |_type, resilience_hour|
+      results_out << ["#{resilience_hour.name} (#{resilience_hour.annual_units})", resilience_hour.annual_output.to_f.round(n_digits)]
     end
 
     results_out = append_sizing_results(results_out, line_break)
@@ -1873,6 +1896,39 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     return val
   end
 
+  def get_resilience_hours(batt_kwh, batt_kw, batt_roundtrip_efficiency, batt_soc_kwh, crit_load)
+    resilience_hours = Array.new(crit_load.size, 0)
+    n_timesteps = crit_load.size
+    (0...1).each do |init_time_step|
+      (0...n_timesteps).each do |i|
+        t = (init_time_step + i) % n_timesteps # for wrapping around end of year
+        load_kw = crit_load[t]
+
+        if load_kw < 0 # load is met
+          if batt_soc_kwh[i] < batt_kwh # charge battery if there's room in the battery
+            batt_soc_kwh[i] += [
+              batt_kwh - batt_soc_kwh[i], # room available
+              batt_kw * batt_roundtrip_efficiency, # inverter capacity
+              -load_kw * batt_roundtrip_efficiency, # excess energy
+            ].min
+          end
+        else # check if we can meet load with generator then storage
+          if [batt_kw, batt_soc_kwh[i]].min >= load_kw # battery can carry balance
+            # prevent battery charge from going negative
+            batt_soc_kwh[i] = [0, batt_soc_kwh[i] - load_kw].max
+            load_kw = 0
+          end
+        end
+
+        if load_kw > 0 # failed to meet load in this time step
+          resilience_hours[init_time_step] = i
+        end
+      end
+    end
+
+    return resilience_hours
+  end
+
   def get_report_meter_data_timeseries(meter_names, unit_conv, unit_adder, timeseries_frequency)
     return [0.0] * @timestamps.size if meter_names.empty?
 
@@ -2002,7 +2058,7 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     @model.getModelObjects.each do |object|
       next if object.to_AdditionalProperties.is_initialized
 
-      [EUT, HWT, LT, ILT].each do |class_name|
+      [EUT, HWT, LT, ILT, RHT].each do |class_name|
         vars_by_key = get_object_output_variables_by_key(@model, object, class_name)
         next if vars_by_key.size == 0
 
@@ -2098,6 +2154,14 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     attr_accessor(:variables, :annual_output_by_system, :timeseries_output_by_system)
   end
 
+  class ResilienceHours < BaseOutput
+    def initialize(variables: [])
+      super()
+      @variables = variables
+    end
+    attr_accessor(:variables)
+  end
+
   class PeakFuel < BaseOutput
     def initialize(meters:, report:)
       super()
@@ -2133,13 +2197,6 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
       @ems_variable = ems_variable
     end
     attr_accessor(:ems_variable)
-  end
-
-  class AverageResilience < BaseOutput
-    def initialize
-      super()
-    end
-    attr_accessor()
   end
 
   class IdealLoad < BaseOutput
@@ -2376,6 +2433,15 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
       hot_water.timeseries_units = 'gal'
     end
 
+    # Resilience Hours
+    @resilience_hours = {}
+    @resilience_hours[RHT::Battery] = ResilienceHours.new(variables: get_object_variables(RHT, RHT::Battery))
+
+    @resilience_hours.each do |resilience_hours_type, resilience_hour|
+      resilience_hour.name = "Resilience Hours: #{resilience_hours_type}"
+      resilience_hour.annual_units = 'hr'
+    end
+
     # Peak Fuels
     # Using meters for energy transferred in conditioned space only (i.e., excluding ducts) to determine winter vs summer.
     @peak_fuels = {}
@@ -2449,14 +2515,6 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
       comp_load.name = "Component Load: #{load_type.gsub(': Delivered', '')}: #{comp_load_type}"
       comp_load.annual_units = 'MBtu'
       comp_load.timeseries_units = 'kBtu'
-    end
-
-    # Average Resilience
-    @average_resilience = {}
-    [AR::Battery].each do |type|
-      @average_resilience[type] = AverageResilience.new
-      @average_resilience[type].name = "Average Resilience: #{type}"
-      @average_resilience[type].annual_units = 'hr'
     end
 
     # Unmet Hours
@@ -2807,6 +2865,13 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
 
       end
 
+    elsif class_name == RHT
+
+      # Resilience Hours
+
+      if object.to_ElectricLoadCenterStorageLiIonNMCBattery.is_initialized
+        return { RHT::Battery => ['Electric Storage Charge Fraction'] }
+      end
     end
 
     return {}
--- a/HPXMLtoOpenStudio/resources/output.rb
+++ b/HPXMLtoOpenStudio/resources/output.rb
@@ -103,6 +103,11 @@ class UHT
   Cooling = 'Cooling'
 end
 
+class RHT
+  # Resilience Hours Types
+  Battery = 'Battery'
+end
+
 class ILT
   # Ideal Load Types
   Heating = 'Heating'
@@ -137,11 +142,6 @@ class WT
   WindSpeed = 'Wind Speed'
   DiffuseSolar = 'Diffuse Solar Radiation'
   DirectSolar = 'Direct Solar Radiation'
-end
-
-class AR
-  # Average Resilience
-  Battery = 'Battery'
 end
 
 class OutputMethods
--- a/HPXMLtoOpenStudio/resources/output.rb
+++ b/HPXMLtoOpenStudio/resources/output.rb
@@ -139,6 +139,11 @@ class WT
   DirectSolar = 'Direct Solar Radiation'
 end
 
+class AR
+  # Average Resilience
+  Battery = 'Battery'
+end
+
 class OutputMethods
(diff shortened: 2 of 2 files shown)


joseph-robertson (commented):
@jmaguire1 Some questions for you.

ReportSimulationOutput/measure.rb
```diff
@@ -870,6 +871,25 @@ def get_outputs(runner, timeseries_frequency,
...
+
+      resilience_hour.variables.map { |v| v[0] }.uniq.each do |sys_id|
+        keys = resilience_hour.variables.select { |v| v[0] == sys_id }.map { |v| v[1] }
+        vars = resilience_hour.variables.select { |v| v[0] == sys_id }.map { |v| v[2] }
+
+        batt_kwh = @hpxml.batteries[0].usable_capacity_kwh
+        batt_kw = @hpxml.batteries[0].rated_power_output
+        batt_roundtrip_efficiency = 0.95
+        batt_soc_kwh = get_report_variable_data_timeseries(keys, vars, 1, 0, timeseries_frequency)
+        crit_load = outputs[:elec_net_timeseries]
```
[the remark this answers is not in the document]
jmaguire1: @joseph-robertson: Close, but needs a couple of changes:
- The SoC in kWh would be the usable capacity multiplied by the SoC (SoC is 0-1). We also need to account for our min SoC not being 0, so I think we would calculate this as (get_report_variable_data_timeseries(keys, vars, 1, 0, timeseries_frequency)- minimum_storage_state_of_charge_fraction) * batt_kWh
-  Might be a good idea to make a new variable for the battery roundtrip efficiency and use that in the battery model and here, rather than relying on a hard coded value matching the default in case we ever change this. 
- Critical load seems fine. In the future we could choose only certain loads to be "critical", but I think whole home net power (treating every load as critical) is good for the initial implementation.

ReportSimulationOutput/measure.rb
```diff
@@ -1872,6 +1896,39 @@ def get_report_variable_data_annual(key_values, variables, unit_conv = UnitConve
     return val
   end
 
+  def get_resilience_hours(batt_kwh, batt_kw, batt_roundtrip_efficiency, batt_soc_kwh, crit_load)
+    resilience_hours = Array.new(crit_load.size, 0)
+    n_timesteps = crit_load.size
+    (0...n_timesteps).each do |init_time_step|
+      (0...n_timesteps).each do |i|
```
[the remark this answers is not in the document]
jmaguire1: Yes, that's correct. Every hour you look how much energy is currently in the battery and the energy consumption over the next several hours to see how long it'd take for the battery to be drained if it were the only source of energy for the home . Then we average those results (which is in essence assuming a power outage is equally likely to happen on any hour of the year, maybe not true but tough to find data for weighting these) to get an annual value we can report.

joseph-robertson:
Sounds like we want to pursue this approach (i.e., report resilience hours in `results_timeseries.csv` and not a separate file). We'll do the calculation hourly at a minimum, but sometimes shorter (i.e., when subhourly timestep is requested). Hourly can be rolled to daily, monthly.

shorowit:
> Sounds like we want to pursue this approach (i.e., report resilience hours in `results_timeseries.csv` and not a separate file). We'll do the calculation hourly at a minimum, but sometimes shorter (i.e., when subhourly timestep is requested). Hourly can be rolled to daily, monthly.

@joseph-robertson  We have some code that does this kind of rollup (hourly->daily or hourly->monthly) for emission calculations:
https://github.com/NREL/OpenStudio-HPXML/blob/7c43a1d60f33381efc9ce78a3ea858f9357e9d53/ReportSimulationOutput/measure.rb#L1020-L1040

It'd be great if you could convert it into a method and use the same code both places.

ReportSimulationOutput/measure.rb
```diff
@@ -872,6 +913,76 @@ def get_outputs(runner, args)
...
+          resilience_timeseries << get_resilience_timeseries(init_time_step, batt_kw, batt_soc_kwh[init_time_step], crit_load, n_timesteps, ts_per_hr)
+        end
+
+        resilience.annual_output = resilience_timeseries.sum(0.0) / resilience_timeseries.size
+        resilience.timeseries_output = resilience_timeseries
+
+        # Aggregate up from hourly to the desired timeseries frequency
+        next unless ['daily', 'monthly'].include? args[:timeseries_frequency]
+
+        year = 1999 # Try non-leap year for calculations
```
joseph-robertson, line 972: @shorowit I could use your help here. In your emissions example, looks like you "try" a couple different years? I'm not sure how to translate that here.

ReportSimulationOutput/measure.rb
```diff
@@ -1715,6 +1831,27 @@ def get_report_variable_data_annual(key_values, variables, unit_conv = UnitConve
...
+      if load_kw > 0
+        if [batt_kw, batt_soc_kwh].min >= load_kw # battery can carry balance
+          # prevent battery charge from going negative
+          batt_soc_kwh = [0, batt_soc_kwh - load_kw].max
+          load_kw = 0
+        end
+      end
+
+      if load_kw > 0 # failed to meet load in this time step
+        return i / Float(ts_per_hr)
```
joseph-robertson, line 1848: @jmaguire1 I changed `return i` to be `return i / Float(ts_per_hr)`, where `ts_per_hr` is either (A) 1 when performed on hourly data, or (B) the number of timesteps per hour when performed on subhourly data.

ReportSimulationOutput/measure.rb
shorowit: I don't remember why I use this kind of brute-force approach to figuring out if it's a leap year or not. Maybe you can come up with something better?

joseph-robertson: Use `@hpxml.header.sim_calendar_year`?

shorowit: Yeah, maybe it's that simple...

shorowit: No wait, the issue is that you could have a leap-year specified as the calendar year but could run a weather file with 365 days, in which case you'll get 365 days of output from E+. So it's not that simple.

joseph-robertson: Oh, looks like your use is specific to dealing with emissions hourly elec factors. For these calculations, I think we can just pass in `@hpxml.header.sim_calendar_year`.

joseph-robertson: Hmm, `base-simcontrol-calendar-year-custom.xml` throws an error if you change the `CalendarYear` to, e.g., 2012: `Error: Specified a leap year (2012) but weather data has 8760 hours.`. Also, `hpxml.header.sim_calendar_year` gets used in other places throughout the measure...

jmaguire1:
Just reviewed the results here at both hourly and subhourly timesteps. There's a little bit of a difference in the resilience hours, but it's down to how EnergyPlus reports outputs. If you run with hourly timesteps, the reported SoC is equal to the SoC at the end of the hour. If you use shorter timesteps, the reported SoC is the average value over the hour (which is less than the final value for hours where the battery is charging and vice versa). It's a 1-2 hour difference in the timeseries results, but I think it's unavoidable with how E+ reports outputs.

jmaguire1:
@joseph-robertson: Good news! After the change you made to account for battery losses in the critical load, I'm not getting exactly the same result in my hand calculation as was in the code:

image
Let me explain a little what I'm doing here: in the resilience hours code, from the initial SoC we calculate what the battery SoC would be next time step based on the critical load and the round trip efficiency. Something we were missing previously was the ability for the battery to charge (up to the maximum SoC) if the home was net producing, ie it had solar. When I hand calculate the SoC over time until it hits 0, I get the exact same # of hours that can be supported for this example time, 44 hours.

Worth pointing out that the last time I showed results here there was a major issue (and in OCHRE too, so I talked to Michael Blonsky about that) that the battery would not attempt to recharge if the home was net producing. This means we can now get some very high numbers of resilience hours:

image

Some pretty interesting results from the test file. You get the largest # of resilience hours in relatively mild (so electricity consumption is low) but sunny weather. In summer the AC is running enough that there isn't enough net production to keep the battery charged. 

@shorowit : just tagging you so you can see the latest results. Realizing we weren't letting the battery charge off solar was a big change, but I'm confident in these results. There were also a couple of other bug fixes/changes, we weren't respecting inverter capacity (which rarely comes up) and we decided to set resilience hours to 0 if the SoC is 0 even if you're net producing (this caused a lot of differences between timestep and hourly results).

ReportSimulationOutput/measure.rb
```diff
@@ -127,6 +127,12 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     arg.setDefaultValue(true)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeBoolArgument('include_annual_resilience', false)
+    arg.setDisplayName('Generate Annual Output: Resilience')
+    arg.setDescription('Generates annual resilience.')
```
shorowit, line 132:
```suggestion
    arg.setDescription('Generates annual resilience outputs.')
```

ReportSimulationOutput/measure.rb
```diff
@@ -223,6 +229,12 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     arg.setDefaultValue(false)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeBoolArgument('include_timeseries_resilience', false)
+    arg.setDisplayName('Generate Timeseries Output: Resilience')
+    arg.setDescription('Generates timeseries resilience.')
```
shorowit, line 234:
```suggestion
    arg.setDescription('Generates timeseries resilience outputs.')
```

ReportSimulationOutput/measure.rb
```diff
@@ -872,6 +919,82 @@ def get_outputs(runner, args)
...
+        sim_start_day_of_year, sim_end_day_of_year, _sim_start_hour, _sim_end_hour = get_sim_times_of_year(year)
+        n_hours_per_period = get_n_hours_per_period(args[:timeseries_frequency], sim_start_day_of_year, sim_end_day_of_year, year)
+
+        resilience_timeseries = []
+        start_hour = 0
+        n_hours_per_period.each do |n_hours|
+          resilience_timeseries << resilience.timeseries_output[start_hour..start_hour + n_hours - 1].sum(0.0) / resilience.timeseries_output[start_hour..start_hour + n_hours - 1].size
+          start_hour += n_hours
+        end
+        resilience.timeseries_output = resilience_timeseries
```
shorowit, line 994: You created a `get_n_hours_per_period` method, which is good and helps reduce the duplicated code, but there is still a lot of duplicated code between here and the emissions code below. Can we replace that method with a `rollup_timeseries_output_to_daily_or_monthly` or similar method that includes ALL of this code?

E.g.:
```
resilience.timeseries_output = resilience_timeseries
if ['daily', 'monthly'].include? args[:timeseries_frequency]
  resilience.timeseries_output = rollup_timeseries_output_to_daily_or_monthly(resilience.timeseries_output, ...)
end
```

ReportSimulationOutput/measure.rb
```diff
@@ -353,6 +365,28 @@ def energyPlusOutputRequests(runner, user_arguments)
...
+      if args[:include_annual_resilience] || args[:include_timeseries_resilience]
+        resilience_frequency = 'timestep'
+        if args[:timeseries_frequency] != 'timestep'
+          resilience_frequency = 'hourly'
+
+          # We need the following request in case no other output types are requested
+          # I.e., this will force a eplusout_xxx.msgpack from which we get timestamps
+          if ['daily', 'monthly'].include?(args[:timeseries_frequency])
+            result << OpenStudio::IdfObject.load("Output:Variable,*,Site Outdoor Air Drybulb Temperature,#{args[:timeseries_frequency]};").get
+          end
```
shorowit, line 379: Hmm. I don't immediately understand why we need this for resilience outputs but not for emissions outputs. How does it work when you request daily/monthly emissions outputs?

ReportSimulationOutput/measure.rb
```diff
@@ -1715,6 +1843,39 @@ def get_report_variable_data_annual(key_values, variables, unit_conv = UnitConve
     return val
   end
 
+  def get_resilience_timeseries(init_time_step, batt_kwh, batt_kw, batt_soc_kwh, crit_load, batt_roundtrip_eff, n_timesteps, ts_per_hr)
+    (0...n_timesteps).each do |i|
```
shorowit, line 1847:
```suggestion
    for i in 0...n_timesteps
```

ReportSimulationOutput/measure.rb
```diff
@@ -353,6 +365,28 @@ def energyPlusOutputRequests(runner, user_arguments)
       if args[:include_timeseries_fuel_consumptions]
         result << OpenStudio::IdfObject.load("Output:Meter,ElectricStorage:ElectricityProduced,#{args[:timeseries_frequency]};").get
       end
+
+      # Resilience
+      if args[:include_annual_resilience] || args[:include_timeseries_resilience]
+        resilience_frequency = 'timestep'
```
shorowit, line 371: So for an HPXML w/ a battery, the default will now be to request timeseries data for multiple variables and then do some calculations on a timeseries basis. I want to make sure there is not a significant impact on runtime from this; if there is, we should probably default `include_annual_resilience` to false.

docs/source/workflow_outputs.rst
```diff
@@ -403,6 +403,17 @@ Annual hot water uses are listed below.
    Hot Water: Distribution Waste (gal) 
    ===================================  ====================
 
+Resilience Hours
+~~~~~~~~~~~~~~~~
```
shorowit, line 407: Use same name as the output.

```suggestion
Resilience
~~~~~~~~~~
```

docs/source/workflow_outputs.rst
```diff
@@ -403,6 +403,17 @@ Annual hot water uses are listed below.
...
+
+Resilience hours by system are listed below.
```
shorowit, line 409:
```suggestion
Resilience outputs are listed below.
```

docs/source/workflow_outputs.rst
```diff
@@ -403,6 +403,17 @@ Annual hot water uses are listed below.
...
+
+   ===================================  ====================
+   Type                                 Notes
+   ===================================  ====================
+   Resilience: Battery (hr)             Use the battery state of charge and the load profile to figure out how long the battery would have been able to support all of the electrical loads. This treats all of the loads as "critical loads" that would be supported during an outage. Calculation is done for every timestep and then averaged, which assumes an outage is equally likely to occur every hour of the year. Resilience hours are set to 0 for any timestep where the battery is not charged, even if there is sufficient PV to power the building.
```
shorowit, line 414: Simplify the note to a shorter description and move the rest of it into a table footnote. I haven't looked but I assume the table will look ridiculous with this much text.

docs/source/workflow_outputs.rst
```diff
@@ -497,6 +508,7 @@ Depending on the outputs requested, the file may include:
    Zone Temperatures                    Zone temperatures (in deg-F) for each space (e.g., living space, attic, garage, basement, crawlspace, etc.) plus heating/cooling setpoints.
    Airflows                             Airflow rates (in cfm) for infiltration, mechanical ventilation (including clothes dryer exhaust), natural ventilation, whole house fans.
    Weather                              Weather file data including outdoor temperatures, relative humidity, wind speed, and solar.
+   Resilience                           Average resilience hours for each applicable system.
```
shorowit, line 511:
```suggestion
   Resilience                           Resilience outputs (currently only average resilience hours for battery storage).
```

ReportSimulationOutput/tests/output_report_test.rb
```diff
@@ -635,11 +642,13 @@ def test_annual_disabled_outputs
...
                   'include_annual_component_loads' => false,
                   'include_annual_hot_water_uses' => false,
-                  'include_annual_hvac_summary' => false }
+                  'include_annual_hvac_summary' => false,
+                  'include_annual_resilience' => false }
     annual_csv, timeseries_csv = _test_measure(args_hash)
     assert(File.exist?(annual_csv))
     assert(!File.exist?(timeseries_csv))
     actual_annual_rows = _get_actual_annual_rows(annual_csv)
+    puts actual_annual_rows.keys
```
shorowit, line 651: Remove

ReportSimulationOutput/measure.rb
```diff
@@ -872,6 +919,82 @@ def get_outputs(runner, args)
       @totals[TE::Net].timeseries_output = @totals[TE::Total].timeseries_output.zip(outputs[:elec_prod_timeseries]).map { |x, y| x + y * unit_conv }
     end
 
+    # Resilience
+    @resilience.each do |key, resilience|
+      next unless key == RT::Battery
+
+      resilience.variables.map { |v| v[0] }.uniq.each do |sys_id|
```
shorowit, line 926: I don't think this code would actually work if there were multiple batteries, since the `sys_id` is not showing up in the output at all. It looks like the last battery calculation would just override the previous ones. I guess nothing necessarily has to be done in the code here since it gives the correct output, but it is/will be confusing if we ever do support multiple batteries.

shorowit (commented):
Initial comments below.

I'm not going to try to understand what the algorithm is doing, I assume you both have tested it and are happy with its results. That said, there are a lot of combinations of timesteps (hourly vs sub-hourly) and reporting frequencies (timestep, hourly, daily/monthly) that need to be tested. Do you feel like you've checked them all?

shorowit:
@jmaguire1 How does the calculation work for timesteps near the end of the year -- does it cycle around to look at electricity consumption at the beginning of the year?

jmaguire1:
> @jmaguire1 How does the calculation work for timesteps near the end of the year -- does it cycle around to look at electricity consumption at the beginning of the year?

@shorowit: Yes exactly that, it wraps to the beginning of the year. Seemed like the most reasonable thing we can do.

ReportSimulationOutput/measure.rb
joseph-robertson: @shorowit I think it may be broken for when you request monthly emissions (and no other timeseries outputs). Try:

```
$ /c/openstudio-3.6.1/bin/openstudio.exe workflow/run_simulation.rb -x workflow/sample_files/base-pv-battery-scheduled.xml --monthly emissions
```

and you don't see either eplusout_monthly.msgpack or results_timeseries.csv files.

shorowit: @joseph-robertson That's because there are no emissions scenarios defined in that HPXML. If you copy emissions scenarios from base-misc-emissions.xml into it and then try again, you'll get monthly emissions outputs.

joseph-robertson: Here's my solution (similar to what is done for emissions):
```
  def setup_timeseries_includes(emissions, args)
    # To calculate timeseries emissions or timeseries fuel consumption, we also need to select timeseries
    # end use consumption because EnergyPlus results may be post-processed due to HVAC DSE.
    # TODO: This could be removed if we could account for DSE inside EnergyPlus.
    args = args.dup # We don't want to modify the original arguments
    args[:include_hourly_electric_end_use_consumptions] = false
    if not emissions.empty?
      args[:include_hourly_electric_end_use_consumptions] = true # Need hourly electricity values for Cambium
      if args[:include_timeseries_emissions] || args[:include_timeseries_emission_end_uses] || args[:include_timeseries_emission_fuels]
        args[:include_timeseries_fuel_consumptions] = true
      end
    end
    if args[:include_timeseries_total_consumptions] || args[:include_timeseries_resilience] <== ADDED
      args[:include_timeseries_fuel_consumptions] = true
    end
    if args[:include_timeseries_fuel_consumptions]
      args[:include_timeseries_end_use_consumptions] = true
    end
    if args[:include_timeseries_system_use_consumptions]
      args[:include_timeseries_end_use_consumptions] = true
    end
    return args
  end
```

shorowit: Thanks Joe, that looks good. If you remove the ADDED code, do you know if any of our CI tests will fail? I'd like to ensure we have a test that covers this situation.

ReportSimulationOutput/measure.rb
joseph-robertson: Sample File | Timestep | Timeseries Freq | Completed simulation | Completed In
-- | -- | -- | -- | --
base-pv-battery.xml | 60 | Timestep | 3.5s | 7.6s
base-pv-battery.xml | 60 | Hourly | 3.4s | 7.1s
base-pv-battery.xml | 60 | Monthly | 3.1s | 6.5s
base-pv-battery.xml | 60 | Daily | 3.2s | 6.6s
base-pv-battery.xml | 10 | Timestep | 16.9s | 24.1s
base-pv-battery.xml | 10 | Hourly | 16.3s | 20.1s
base-pv-battery.xml | 10 | Monthly | 15.6s | 19.1s
base-pv-battery.xml | 10 | Daily | 16.1s | 19.6s

ReportSimulationOutput/measure.rb
joseph-robertson: Should we therefore leave this as it is now, until 2+ batteries are allowed? Or what changes, specifically, are you suggesting?

shorowit: I guess it's fine to leave for now. We'd need to refactor the code a bit if we ever allow multiple batteries or additional resilience outputs, but that can happen in the future. I'll mark this as resolved.

joseph-robertson:
> Initial comments below.
> 
> I'm not going to try to understand what the algorithm is doing, I assume you both have tested it and are happy with its results. That said, there are a lot of combinations of timesteps (hourly vs sub-hourly) and reporting frequencies (timestep, hourly, daily/monthly) that need to be tested. Do you feel like you've checked them all?

I added a "Avg hr" column to the table in https://github.com/NREL/OpenStudio-HPXML/pull/1212#discussion_r1258912568. This might help show level of agreement between different combinations of timesteps and reporting frequencies. I'll leave it up to @jmaguire1 to comment here further.

shorowit (approved):
Looks good to me. Will merge when the CI comes back green. Thanks @joseph-robertson and @jmaguire1!

PR #1215, opened 2022-11-08
Component loads alternative

shorowit:
## Pull Request Description

Alternative approach to component loads for hours where there is no heating/cooling load. This approach compares the average of the heating and cooling setpoint to the zone temperature during a timestep to assign a load to heating or cooling. If the zone temperature is higher than the average setpoint, loads are assigned cooling, if lower, than loads are assigned heating. This differs from the original approach which defaulted to heating and cooling seasons to assign loads during "floating periods". This approach better captures periods when heating occurs in cooling seasons and vice versa. This is particularly impactful during shoulder seasons and locations with high diurnal temperature swings.

image

## Checklist

- [ ] ~~Schematron validator (`EPvalidator.xml`) has been updated~~
- [ ] ~~Sample files have been added/updated (via `tasks.rb`)~~
- [ ] ~~Unit tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests`)~~
- [ ] ~~Documentation has been updated~~


aspeake:
Some scatter plots of regression tests for select outputs:

image
image
image

aspeake:
@shorowit can't request your review because it is your PR, but this is ready for you review...

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2362,17 +2373,47 @@ def self.add_component_loads_output(model, conditioned_zone, liv_load_sensors, i
...
+    if (not htg_sp_sensor.nil?) && (not clg_sp_sensor.nil?)
+      program.addLine("  Set Tmid_setpoint = (#{htg_sp_sensor.name} + #{clg_sp_sensor.name}) / 2") # Average of heating/cooling setpoints
+    else
+      program.addLine("  Set Tmid_setpoint = #{UnitConversions.convert(73.0, 'F', 'C')}") # Assumption when no HVAC system
+    end
+    program.addLine("  If #{tin_sensor.name} > Tmid_setpoint")
+    program.addLine("    Set clg_mode = #{total_cool_load_served}")
+    program.addLine('  Else')
+    program.addLine("    Set htg_mode = #{total_heat_load_served}")
+    program.addLine('  EndIf')
```
shorowit, line 2416: Seems like this approach wouldn't handle well the situation where heating and cooling setpoints are the same. In that case, it would always choose heating mode.

shorowit: Though I just realized this code only gets hit when the temperature is floating (fallback to when there's no heating or cooling load), and you should never be floating if the heating and cooling setpoints are the same.

shorowit:
I'm pretty happy with this, but wanted to think a bit more about how homes without e.g. space cooling equipment are handled.

shorowit:
I've been doing some testing and am really happy with the results. It does seem like a significant improvement over the current methodology. Sorry for the long delay on this and thanks for your work on this, @aspeake!

PR #1237, opened 2022-12-02
Fix tsp input

aspeake:
## Pull Request Description
Allows heating and cooling setpoints to be set via user arguments when there is no heating/cooling systems.

## Checklist

### DIFF
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -4965,17 +4965,16 @@ class HPXMLFile
     return if (args[:heating_system_type] == 'none') && (args[:cooling_system_type] == 'none') && (args[:heat_pump_type] == 'none')
 
     # Heating
-    if hpxml.total_fraction_heat_load_served > 0
-
-      if args[:hvac_control_heating_weekday_setpoint].is_initialized && args[:hvac_control_heating_weekend_setpoint].is_initialized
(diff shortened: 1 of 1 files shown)


shorowit (commented):
Rather than providing heating/cooling setpoints when there is no heating/cooling equipment (which is weird and confusing), solely for the purpose of natural ventilation, it's probably a better idea to add an explicit natural ventilation setpoint input in the HPXML file. It can be optional and default to the average of the heating/cooling setpoints, but could be changed by a user. That would also allow more flexibility even when heating/cooling setpoints are provided.

BuildResidentialHPXML/measure.rb
```diff
@@ -5036,6 +5041,11 @@ def self.set_hvac_control(hpxml, args, epw_file, weather)
 
     end
 
+    # Natural Ventilation
```
aspeake, line 5044: I'm a little unsure where this input should live, so I have tentatively added it to HVACControl

@shorowit thoughts?

shorowit: My initial thought is to add it next to the `NaturalVentilationAvailabilityDaysperWeek` element; see https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#hpxml-windows for where that description.

aspeake: Yep I think that would make sense, I'll make the change

PR #1242, opened 2022-12-06
Battery outputs fix

### LINKED ISSUE #1240: Electricity:Total timeseries output doesn't reflect battery
When a home has a battery, most outputs look correct -- but the **timeseries** Electricity:Total end use does not. (It's particularly strange because the **annual** value looks correct, and we have error-checking in the reporting measure that is supposed to throw an error if the annual value doesn't match the sum of the timeseries values.)

joseph-robertson:
## Pull Request Description

Closes #1240.

Add ElectricStorage:ElectricityProduced onto Electricity:Facility. We were doing this, but then (incorrectly) removed it when switching to the EMS approach for round trip efficiency.

## Checklist


shorowit (approved):
Looks good. I tested base-pv-battery.xml and base-battery-scheduled.xml and saw the change I expected. Will merge if the CI is still happy after my latest commit.

PR #1260, opened 2022-12-29
Allow user-specified precision for timeseries outputs

### LINKED ISSUE #1259: Allow user-specified precision for timeseries outputs
For most use cases, I think we provide reasonable default precision for timeseries results:

https://github.com/NREL/OpenStudio-HPXML/blob/9112f5941b4e7965e0ddd0e13f30da2500ea8c77/ReportSimulationOutput/measure.rb#L1726-L1735

But for use cases like ResStock, where results may be aggregated across millions of homes, increased precision may be of interest.

I propose that we add an optional argument to the ReportSimulationOutput reporting measure where someone can override the default number of decimal places if they want to.

cc @rajeee

shorowit:
## Pull Request Description

Closes #1259. ReportSimulationOutput measure: Allows specifying the number of decimal places for timeseries output. Also increases the number of default decimal places by 1.

## Checklist

- [x] ~Unit tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests`)~

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 __New Features__
 - Allows modeling one or more occupant vacancy periods (`VacancyPeriods` in the HPXML file).
+- ReportSimulationOutput measure: Allows specifying the number of decimal places for timeseries output.
 
 __Bugfixes__
 - Fixes error if calculating utility bills for an all-electric home with a detailed JSON utility rate.
--- a/ReportSimulationOutput/measure.rb
+++ b/ReportSimulationOutput/measure.rb
@@ -142,6 +142,11 @@ class ReportSimulationOutput < OpenStudio::Measure::ReportingMeasure
     arg.setDefaultValue('start')
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeIntegerArgument('timeseries_num_decimal_places', false)
+    arg.setDisplayName('Generate Timeseries Output: Number of Decimal Places')
+    arg.setDescription('Allows overriding the default number of decimal places for timeseries output.')
+    args << arg
+
     arg = OpenStudio::Measure::OSArgument::makeBoolArgument('add_timeseries_dst_column', false)
     arg.setDisplayName('Generate Timeseries Output: Add TimeDST Column')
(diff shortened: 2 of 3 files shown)


shorowit:
@joseph-robertson You'll want to use this when it gets pulled into resstock; consult with @rajeee.

joseph-robertson:
@shorowit @rajeee

Looks like `n_digits` now defaults to one more than before (3 vs 2)? Is this the number of decimal places we want to use in resstock's example project yml files?

shorowit:
No, you'll definitely want to user a higher number. I don't know what is sufficient though.

rajeee:
We have run into issues (in EUSS) due to rounding up of 0.00075 to 0.001. So, at least 5 digits seems necessary.

nmerket:
@shorowit and @joseph-robertson: @rajeee and I were discussing this problem yesterday. Ultimately it comes down to the fact that for ResStock we're converting a binary double precision floating point number in the sqlite outputs from EnergyPlus to a text representation (csv) with a limited precision and then back to a binary double precision floating point number (parquet). Then adding them up across 35k timesteps across hundreds of thousands of simulations. The rounding errors accumulate and end up being quite large. Increasing the number of digits will help, but it will balloon the file size of these outputs. The best solution to this problem would remove the csv step where the rounding happens. Would it be possible to save these outputs in a binary format like parquet or msgpack that we could then read and aggregate in ResStock?

nmerket:
It looks like you're [already have an option for msgpack outputs](https://github.com/NREL/OpenStudio-HPXML/blob/cf920af910e62102598ecc8a3cd02e14f867f635/ReportSimulationOutput/measure.rb#L1347-L1348), but we'd still have the same problems because the [rounding happens upstream of that](https://github.com/NREL/OpenStudio-HPXML/blob/cf920af910e62102598ecc8a3cd02e14f867f635/ReportSimulationOutput/measure.rb#L1324). If we could get unrounded msgpack outputs, that would be really helpful.

shorowit:
https://github.com/NREL/OpenStudio-HPXML/issues/1262

In the meantime, they can use https://github.com/NREL/OpenStudio-HPXML/pull/1260

nmerket:
Ah two steps ahead of me.

PR #1274, opened 2023-01-21
Refine Component Load Outputs

aspeake:
## Pull Request Description

Changes/adds the following outputs for component loads:
- Lighting is broken out as its own component, separate from internal gains
- Windows is broken out into `windows_solar` and `windows_conduction`

## Checklist

- [ ] ~~Schematron validator (`EPvalidator.xml`) has been updated~~
- [ ] ~~Sample files have been added/updated (via `tasks.rb`)~~


Changelog.md
```diff
@@ -7,6 +7,7 @@ __New Features__
 - LightingGroups can now be specified using kWh/year annual consumption values as an alternative to fractions of different lighting types.
 - Allows modeling one or more occupant vacancy periods (`VacancyPeriods` in the HPXML file).
 - ReportSimulationOutput measure: Allows specifying the number of decimal places for timeseries output.
+- Adds new component loads outputs `Windows Conduction`, `Windows Solar`, `Skylights Conduction`, `Skylights Solar`, `Lighting`, and `Residual`, removes `Windows`, `Skylights`, and `Internal Mass` components.
```
shorowit, line 10: Software applications may be using the current component load types, so we need to mark this as a breaking change. Let's also try to make it a bit clearer what the changes are.

```suggestion
- **Breaking change**: Updates component loads outputs:
  - Replaces `Windows` and `Skylights` with `Windows Conduction`, `Windows Solar`, `Skylights Conduction`, and `Skylights Solar`.
  - Disaggregates `Lighting` from `Internal Gains`.
  - Adds `Residual` (calculated as difference between energy delivered and sum of component loads).
  - Removes `Internal Mass`.
```

shorowit: Also, the documentation needs to be updated. See https://openstudio-hpxml.readthedocs.io/en/latest/workflow_outputs.html#annual-component-building-loads

ReportSimulationOutput/measure.rb
```diff
@@ -1318,6 +1318,13 @@ def report_runperiod_output_results(runner, outputs, output_format, annual_outpu
       @component_loads.each do |_load_type, load|
         results_out << ["#{load.name} (#{load.annual_units})", load.annual_output.to_f.round(n_digits)]
       end
+
+      sum_comp_htg = results_out.filter_map { |row| row[1] if (not row[0].nil?) && (row[0].start_with? 'Component Load: Heating:') }.sum()
+      sum_comp_clg = results_out.filter_map { |row| row[1] if (not row[0].nil?) && (row[0].start_with? 'Component Load: Cooling:') }.sum()
+      residual_htg = @loads[LT::Heating].annual_output.to_f - sum_comp_htg
+      residual_clg = @loads[LT::Cooling].annual_output.to_f - sum_comp_clg
+      results_out << ['Component Load: Heating: Residual (MBtu)', residual_htg.round(n_digits)]
+      results_out << ['Component Load: Cooling: Residual (MBtu)', residual_clg.round(n_digits)]
```
shorowit, line 1327: This produces the Residual load for annual output. I assume similar code is needed for timeseries output?

shorowit (commented):
This looks pretty reasonable and definitely improves things. Thanks, @aspeake. Just a few comments below. Once those are addressed, I'll double-check the CI output and do some manual spot-checks.

ReportSimulationOutput/measure.rb
aspeake: I debated whether to include a timeseries output for the residual. I am not sure it really makes sense at a timestep-level because the heating/cooling delivered does not change concurrently with the component loads. So we might see several hours of high error when the HVAC energy = 0, then a quick change in the opposite direction once the HVAC system turns on, roughly offsetting to 0 over the long term (hopefully).

The downside is that since it is reported as just another component load here, it may be confusing to not see in the timeseries.

shorowit: I see. Given that, I do wonder if it's really that useful to provide the Residual output. If anyone wants it, it can be easily calculated from the annual outputs that are provided.

shorowit: If there is any confusion as to the relationship between the two, I'm tempted to say we should improve the documentation rather than add another output, which would require even _more_ explanation.

shorowit (approved):
Looks good to me. I will merge this in assuming the CI comes back with a ✔️.

PR #1296, opened 2023-02-28
BuildResidentialHPXML: Check attic floor insulation assembly R-value

### LINKED ISSUE #1295: BuildResidentialHPXML: Check attic floor insulation assembly R-value
We are not gracefully handling an error (divide by zero) when there is an attic and `floor.insulation_assembly_r_value=0` here: https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L873

This can happen since we are applying defaults from `BuildResidentialHPXML` but not necessarily first applying validation on the HPXML file.

Should resstock `apply_validation=true` from BuildExistingModel?

joseph-robertson:
## Pull Request Description

Closes #1295.

## Checklist

### DIFF
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -3270,6 +3270,9 @@ class BuildResidentialHPXML < OpenStudio::Measure::ModelMeasure
     error = (args[:window_aspect_ratio] <= 0)
     errors << 'Window aspect ratio must be greater than zero.' if error
 
+    error = [HPXML::AtticTypeVented, HPXML::AtticTypeUnvented].include?(args[:geometry_attic_type]) && (args[:ceiling_assembly_r] == 0.0)
+    errors << 'Home with unconditioned attic type has zero ceiling insulation assembly R-value specified.' if error
+
     return errors
   end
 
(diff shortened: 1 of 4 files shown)


shorowit:
The change is of course fine, but there are probably a hundred ways that similar errors can occur, and this only catches one of them. And it duplicates error-checking that occurs in the Schema/Schematron documents.

My inclination is to go a different path:
1. Force `apply_validation` to be true when `apply_defaults` is true. (Update the `apply_defaults` arg description to make this clear.)
2. Apply validation _before_ the defaults are applied, and exit the measure prematurely if the validation is unsuccessful. (We could technically apply validation twice, once before the defaulting occurs and again after, but we'd take a hit on performance twice for doing that. I think we should assume that the defaulting code will not somehow produce an invalid HPXML.)

joseph-robertson:
I had thought about essentially doing (1) (i.e., set `apply_validation=true` from BuildExistingModel) - but was worried about a performance hit in the context of resstock. I suppose we could time it using ci...

shorowit:
If validation is enabled in the BuildResHPXML measure, it could then be turned off in the HPXMLtoOS measure ([here](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/measure.rb#L58)).

shorowit:
(So there would be no net performance hit in ResStock.)

shorowit:
@joseph-robertson This can be closed now, right?

joseph-robertson:
Went with this approach instead: https://github.com/NREL/OpenStudio-HPXML/pull/1298.

PR #1298, opened 2023-02-28
BuildResidentialHPXML: Apply validation when applying defaults

joseph-robertson:
## Pull Request Description

This will solve the issue described in https://github.com/NREL/OpenStudio-HPXML/issues/1295.

## Checklist

### DIFF
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -3012,7 +3012,7 @@ class BuildResidentialHPXML < OpenStudio::Measure::ModelMeasure
 
     arg = OpenStudio::Measure::OSArgument::makeBoolArgument('apply_defaults', false)
     arg.setDisplayName('Apply Default Values?')
-    arg.setDescription('If true, applies OS-HPXML default values to the HPXML output file.')
+    arg.setDescription('If true, applies OS-HPXML default values to the HPXML output file. Setting to true will also force validation of the HPXML output file before applying OS-HPXML default values.')
     arg.setDefaultValue(false)
     args << arg
 
@@ -3388,7 +3388,7 @@ class HPXMLFile
     end
 
     # Check for errors in the HPXML object
-    if args[:apply_validation].is_initialized && args[:apply_validation].get
+    if (args[:apply_defaults].is_initialized && args[:apply_defaults].get) || (args[:apply_validation].is_initialized && args[:apply_validation].get)
       errors = hpxml.check_for_errors()
       if errors.size > 0
         fail "ERROR: Invalid HPXML object produced.\n#{errors}"
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -3272,31 +3272,6 @@ class BuildResidentialHPXML < OpenStudio::Measure::ModelMeasure
 
     return errors
   end
-
-  def validate_hpxml(runner, hpxml_path, hpxml_doc)
-    is_valid = true
-
-    # Validate input HPXML against schema
-    schema_dir = File.join(File.dirname(__FILE__), '../HPXMLtoOpenStudio/resources/hpxml_schema')
-    schema_path = File.join(schema_dir, 'HPXML.xsd')
-    xsd_errors, xsd_warnings = XMLValidator.validate_against_schema(hpxml_path, schema_path)
-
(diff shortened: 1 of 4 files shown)


BuildResidentialHPXML/measure.rb
```diff
@@ -3388,7 +3388,7 @@ def self.create(runner, model, args, epw_path)
     end
 
     # Check for errors in the HPXML object
-    if args[:apply_validation].is_initialized && args[:apply_validation].get
+    if (args[:apply_defaults].is_initialized && args[:apply_defaults].get) || (args[:apply_validation].is_initialized && args[:apply_validation].get)
```
shorowit, line 3391: That's good, but it doesn't cover the Schema/Schematron validation, which occurs [here](https://github.com/NREL/OpenStudio-HPXML/blob/apply-validation-defaults/BuildResidentialHPXML/measure.rb#L3085). And it occurs _after_ the HPXML defaulting, so somehow the order needs to be reversed.

Also, rather than trying to update lines of code where `apply_validation` is referenced, a better approach would be to just set `apply_validation` to true when `apply_defaults` is true. Then it will be guaranteed to apply all the appropriate validation _without the possibility of missing references_.

joseph-robertson: Ah I see. So currently BuildResHPXML:
* builds up the `hpxml` object
* runs `hpxml.check_for_errors()` if `apply_validation=true`
* applies defaults if `apply_defaults=true`
* exports `hpxml` to file
* applies Schema/Schematron validation if `apply_validation=true`

We want to apply defaults only after BOTH (1) `hpxml.check_for_errors()` and (2) Schema/Schematron validation have occurred.

joseph-robertson: If my previous comment is correct, wouldn't we need to then export `hpxml` to file twice? Once before defaulting, and then again after defaulting?

shorowit: I think that's right. But exporting to file is extremely fast -- I timed it and found it to be 0.02 seconds w/ WSL on my machine.

shorowit: To provide a bit more info... the `hpxml.check_for_errors()` call is additional error-checking that cannot be performed through Schematron, so we resort to checks in ruby. Realistically, all three validations (Schema, Schematron, ruby) should probably happen in the same place in the measure.

BuildResidentialHPXML/measure.rb
```diff
@@ -3388,23 +3363,58 @@ def self.create(runner, model, args, epw_path)
     end
 
     # Check for errors in the HPXML object
-    if args[:apply_validation].is_initialized && args[:apply_validation].get
+    if args[:apply_validation]
       errors = hpxml.check_for_errors()
       if errors.size > 0
         fail "ERROR: Invalid HPXML object produced.\n#{errors}"
       end
     end
```
shorowit, line 3371: Can we move this into the `validate_hpxml` method? I don't think it needs to be handled separately here.

BuildResidentialHPXML/measure.rb
```diff
@@ -3388,23 +3363,58 @@ def self.create(runner, model, args, epw_path)
...
+      if not validate_hpxml(runner, hpxml_path, hpxml_doc)
+        return false
+      end
+
+      if args[:apply_defaults]
+        eri_version = Constants.ERIVersions[-1]
+        HPXMLDefaults.apply(runner, hpxml, eri_version, weather, epw_file: epw_file)
+        hpxml_doc = hpxml.to_oga()
+      end
+    end
```
shorowit, line 3388: Seems like this could be simplified to:

```suggestion
    if args[:apply_validation]
      # Check for invalid HPXML file
      if not validate_hpxml(runner, hpxml_path, hpxml_doc)
        return false
      end
    end

    if args[:apply_defaults]
      eri_version = Constants.ERIVersions[-1]
      HPXMLDefaults.apply(runner, hpxml, eri_version, weather, epw_file: epw_file)
      hpxml_doc = hpxml.to_oga()
    end
```

shorowit (commented):
Looks pretty good, two small suggestions below.

Also, can you add a test that a ceiling assembly R-value of zero generates the appropriate error message?

shorowit (approved):
Looks good to me.

PR #1317, opened 2023-03-13
Heat pump detailed performance inputs

### LINKED ISSUE #1028: Allow NEEP ccASHP inputs for variable-speed heat pumps
https://ashp.neep.org

### LINKED ISSUE #1341: Replace performance curves w/ TableLookup?
For https://github.com/NREL/OpenStudio-HPXML/issues/1028, we will be using [TableLookup](https://bigladdersoftware.com/epx/docs/22-2/input-output-reference/group-performance-tables.html#tablelookup) objects in EnergyPlus. These objects also allow the HVAC installation quality adjustments to be directly incorporated.

However, we have a number of existing performance curves. Should we convert them all to use TableLookup? This would mean we can use the same HVAC installation quality approach for all HVAC models; even better, that single approach can avoid EMS and its associated runtime penalty.

@yzhou601 @jonwinkler

shorowit:
## Pull Request Description

Updated assumptions for variable-speed air conditioners, heat pumps, and mini-splits.

Also allows detailed heating and cooling performance data (min/max COPs and capacities at different outdoor temperatures) as an optional set of inputs. Data can be sourced from e.g. [NEEP's Cold Climate Air Source Heat Pump List](https://ashp.neep.org).

Closes #1028 and closes #1341. Takes advantage of https://github.com/hpxmlwg/hpxml/pull/359.

TODO:
- [x] Replace existing curve objects with Energyplus lookup table (maybe we can find cutler curve and directly use those data; Or we need to determine the sample size for granularity)
- [x] Implement net to gross capacity conversion
- [x] Use Energyplus lookup table to implement detailed performance inputs
- [x] Consolidate with capacity retention inputs
- [x] Default performance curves from NEEP database
  - [x] Rated COP from hspf and seer for all speeds, to determine for each speed we either can use different regressions (preferred) or use a ratio
  - [x] Rated capacities for all speeds
- [x] Deprecate old performance-related methods
- [x] HVAC autosizing under different scenarios

## Checklist


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2729,86 +2699,56 @@ def self.convert_curve_biquadratic(coeff)
...
+      ind_var.setValues(var[:values])
+      table.addIndependentVariable(ind_var)
+    end
+    table.setMinimumOutput(output_min) unless output_min.nil?
+    table.setMaximumOutput(output_max) unless output_max.nil?
+    table.setOutputValues(output_values)
+    return table
   end
 
   def self.create_curve_quad_linear(model, coeff, name)
```
yzhou601, line 2754: For GSHP curves I didn't move forward to replace them because they need to create 4 or 5 independent variables for each curve, I am not sure if it's an efficient path to go.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2775,10 +2775,34 @@ def self.create_curve_quint_linear(model, coeff, name)
     curve.setCoefficient6z(coeff[5])
     return curve
   end
+  
+  def self.convert_rated_to_gross_capacity_cop_clg(clg_ap)
+    clg_ap.detailed_performance_net_cap_clg.keys().sort.each do |out_db|
+      clg_ap.detailed_performance_net_cap_clg[out_db].each_with_index do |cap, i|
+        net_power = cap / clg_ap.detailed_performance_net_cop_clg[out_db][i]
+        #gross_cap = cap + clg_ap.fan_powers_rated
```
yzhou601, line 2783: @shorowit @jonwinkler resdx is using pretty complicated approach to calculate the cooling_rated_fan_power for each speed,   it uses free efficacy, efficacy slope and the operating pressure to calculate power. See [here](https://github.com/bigladder/resdx/blob/main/resdx/neep_vchp.py#L126), [here](https://github.com/bigladder/resdx/blob/main/resdx/fan.py#L409-L412), and [here](https://github.com/bigladder/resdx/blob/main/resdx/fan.py#L91-L98). 

There're a bunch of function calls, and I don't think we have anything that complicated in our workflow yet. In writing FanSystemModel E+ object, we are using power_fraction = fan_ratio**3 as fan power curve. Does it sound reasonable to use `power_fraction = fan_ratio**3` to calculate the fan power at each stages?
For cfm/ton, we still keep the assumption from previous work, right? With just using the first and last items in the array as min and max?

yzhou601: Proceed with the power_fraction = fan_ratio**3 curve

yzhou601:
Update: Unit tests are failing because I hardcoded the cops for now. Real home 49 is not meeting cooling loads (exceeded the tolerance of 350 hr in the test) due to the change of capacity ratio. The user input capacity for MSHP used to be multiplied by 1.2 capacity ratio for maximum speed, but now changed to be 1.0 (Further change will be applying Jon's assumption approach).

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1520,7 +1520,9 @@ def self.apply_detailed_performance_data_for_var_speed_systems(hpxml)
     hpxml.heat_pumps.each do |heat_pump|
       next unless [HPXML::HVACTypeHeatPumpAirToAir,
                    HPXML::HVACTypeHeatPumpMiniSplit].include? heat_pump.heat_pump_type
-
+      # Assign minimum capacity here to avoid E+ simulation error
+      min_capacity = 1.0
+      heat_pump.heating_capacity = [heat_pump.heating_capacity, min_capacity].max
```
yzhou601, line 1525: This will execute before writing in.xml, so our minimum value will be reflected on the in.xml. Couldn't come up with a better approach off the top of my head. I tried applying minimum value for each data point later at the `ensure_nonzero_sizing_values`, it will still catch some E+ errors saying that the maximum speed airflow should be greater than minimum speed. Will leave it here to try to pass ci first

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2022,6 +2023,7 @@ def self.process_heat_pump_adjustment(hvac_sizing_values, weather, hvac_heating,
       heating_db = @hpxml.header.manualj_heating_design_temp
     end
 
+    # FIXME: Replace performance coefficients with new defaulting correlations
```
yzhou601, line 2026: There's a sequence issue to replace coefficients with the new performance data,. The performance data defaulting method has to be called after the hvac sizing because we need the autosized capacity as 47F rated capacity, and then build other data points based on that, but here the sizing needs the detailed performance data to adjust the capacity. So the correlation cannot be directly called. Is it acceptable to retain the coefficients here? Or we can write the normalized correlation (get rid of capacity) here based on the defaulting method.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
shorowit: @yzhou601 Can you give me an example of how to reproduce the E+ error (assuming I comment out this code)? I can take a closer look at whether there's a better way to address this if I understand the situation better.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: I think I understand. We are using a linear heating capacity as a function of temperature relationship going forward, is that right? Can you just use the heating capacity retention fraction -- either A) entered directly by the user, B) calculated from detailed performance data, or C) defaulted?

shorowit: Oh, I take that back -- the detailed performance data may not be linear.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
yzhou601: [in.idf.txt](https://github.com/NREL/OpenStudio-HPXML/files/12296569/in.idf.txt)
[eplusout.err.txt](https://github.com/NREL/OpenStudio-HPXML/files/12296570/eplusout.err.txt)
See attached the E+ error by running .\workflow\sample_files\base-hvac-mini-split-heat-pump-ducted-cooling-only.xml
The code snippet is like adding the following lines to the ensure_non_zero_sizing_values:

>       if not hp_sys.heating_detailed_performance_data.empty?
>         hp_sys.heating_detailed_performance_data.each do |dp|
>           dp.capacity = [dp.capacity, min_capacity].max
>         end
>       end
>       if not hp_sys.cooling_detailed_performance_data.empty?
>         hp_sys.cooling_detailed_performance_data.each do |dp|
>           dp.capacity = [dp.capacity, min_capacity].max
>         end
>       end

The warnings thrown by E+ is complaining the small airflow.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: So if the detailed performance data is provided, you can use that. If it's not provided, you can use the heating capacity retention fraction (user entered or defaulted). Does that work? I realize that's two pathways for the calculations, and that you were perhaps hoping for a single pathway.

yzhou601: If the detailed performance data is provided, the autosizing shouldn't mean too much since we plan to require the rated capacity. So that won't be an issue for the simulation. But if we report the sizing for other uses then yes we should use the user inputs to adjust. 
And for path when detailed performance data is not provided, then the heating capacity retention path sounds like the most ideal one, and cooling we can apply Jon's defaulting approach. In that case, the only independent variable is the outdoor temperature, the code to introduce indoor condition sensitivity is [here](https://github.com/NREL/OpenStudio-HPXML/pull/1317/files#diff-1747bd7cd2671c2c55032c376b1d67f9d6a37d6df45e254574430f5796c5b830R2871). It's introduced by applying 1 speed cutler coefficients. The capacity coefficients of the indoor condition terms (second and third terms of cap_ft_spec), though, are not nothing, but look different across different systems: 
single speed: 

> cap_ft_spec = [[3.68637657, -0.098352478, 0.000956357, 0.005838141, -0.0000127, -0.000131702]]

var speed ashp max speed: 

> cap_ft_spec = [[1.8769221728395058, -0.04768641975308643, 0.0006885802469135801, 0.006643395061728395, 1.4209876543209876e-05, -0.00024043209876543206]]

mini-split max speed:

> cap_ft_spec = [[0.7531983499655835, 0.003618193903031667, 0.0, 0.006574385031351544, -6.87181191015432e-05, 0.0]]

So I am not sure if we want to introduce this complexity here if the 1 speed correction approach is something we may revisit in the future.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
shorowit: What do you think about this? It increments the capacity by speed rather than setting them all to the same minimum capacity. It seems to work for this test file at least.

```
      if not hp_sys.heating_detailed_performance_data.empty?
        hp_sys.heating_detailed_performance_data.each do |dp|
          speed = dp.capacity_description == 'minimum' ? 1 : 2
          dp.capacity = [dp.capacity, min_capacity * speed].max
        end
      end
      if not hp_sys.cooling_detailed_performance_data.empty?
        hp_sys.cooling_detailed_performance_data.each do |dp|
          speed = dp.capacity_description == 'minimum' ? 1 : 2
          dp.capacity = [dp.capacity, min_capacity * speed].max
        end
      end
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2842,30 +2747,253 @@ def self.create_curve_quint_linear(model, coeff, name)
...
+      gross_cap_btu_hr = UnitConversions.convert(gross_cap_watts, 'w', 'Btu/hr')
+      return gross_cap_btu_hr
+    end
+  end
+
+  def self.setup_neep_detailed_performance_data(detailed_performance_data)
+    data_array = Array.new(2) { Array.new }
+    detailed_performance_data.each do |data_point|
+      # Only process min and max capacities at each outdoor drybulb
+      next unless ['minimum', 'maximum'].include? data_point.capacity_description
```
shorowit, line 2786: Please make these HPXML constants (and check the code for other places they are used).

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2842,30 +2747,253 @@ def self.create_curve_quint_linear(model, coeff, name)
     return curve
   end
 
+  def self.convert_net_to_gross_capacity(net_cap, watts_per_cfm, cfm_per_ton, fan_ratio, mode)
```
shorowit, line 2750: This method appears to be unused. Delete?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2842,30 +2747,253 @@ def self.create_curve_quint_linear(model, coeff, name)
...
+    if output_cop
+      net_power = net_cap_watts / net_cop
+      gross_power = net_power - fan_power
+      gross_cop = gross_cap_watts / gross_power
+      return gross_cop
+    else
+      gross_cap_btu_hr = UnitConversions.convert(gross_cap_watts, 'w', 'Btu/hr')
+      return gross_cap_btu_hr
+    end
+  end
```
shorowit, line 2780: It's a bit weird to have a method that can return two different things depending on a boolean argument. How about have it return both values (though the second will be nil when `net_cop` is not provided)? You'd obviously have to update the calls to the method appropriately.

```suggestion
  def self.convert_net_to_gross_capacity_cop(net_cap, fan_power, mode, net_cop = nil)
    net_cap_watts = UnitConversions.convert(net_cap, 'Btu/hr', 'w')
    if mode == :clg
      gross_cap_watts = net_cap_watts + fan_power
    else
      gross_cap_watts = net_cap_watts - fan_power
    end
    if not net_cop.nil?
      net_power = net_cap_watts / net_cop
      gross_power = net_power - fan_power
      gross_cop = gross_cap_watts / gross_power
    end
    gross_cap_btu_hr = UnitConversions.convert(gross_cap_watts, 'w', 'Btu/hr')
    return gross_cap_btu_hr, gross_cop
  end
```

shorowit (commented):
@yzhou601 A few things I noticed while looking into one of your questions...

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1274,6 +1280,38 @@
...
   </sch:pattern>
+  
+  <!-- FIXME: Need to ensure variable speed? Allow more flexibility below? -->
+  <sch:pattern>
+    <sch:title>[HeatingDetailedPerformanceData]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:HeatingDetailedPerformanceData'>
+      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature=47 and h:CapacityDescription="minimum"]) = 1'>Expected 1 element(s) for xpath: PerformanceDataPoint[OutdoorTemperature=47 and CapacityDescription="minimum"]</sch:assert> <!-- See [PerformanceDataPoint] -->
+      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature=47 and h:CapacityDescription="maximum"]) = 1'>Expected 1 element(s) for xpath: PerformanceDataPoint[OutdoorTemperature=47 and CapacityDescription="maximum"]</sch:assert> <!-- See [PerformanceDataPoint] -->
+      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature=17 and h:CapacityDescription="minimum"]) = 1'>Expected 1 element(s) for xpath: PerformanceDataPoint[OutdoorTemperature=17 and CapacityDescription="minimum"]</sch:assert> <!-- See [PerformanceDataPoint] -->
```
yzhou601, line 1290: It hurt my brain to come up with a schematron rule to say, for a OutdoorTemperature other than rated, require both minimum and maximum being specified. Any suggestion? @shorowit

yzhou601: Added these rules, and generated a test file for more flexibilities

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2842,30 +2603,246 @@ def self.create_curve_quint_linear(model, coeff, name)
...
+        rate_dp = data_speed.find { |dp| (dp.indoor_wetbulb == rated_iwb) && (dp.outdoor_temperature == rated_odb) }
+        rated_cap = rate_dp.gross_capacity
+        rated_eir = 1.0 / rate_dp.gross_efficiency_cop
+        clg_ap.cool_rated_eirs << rated_eir
+        clg_ap.cool_rated_capacities_gross << rated_cap
+        clg_ap.cool_rated_capacities_net << rate_dp.capacity
+        cap_ft_output_values = data_speed.map { |dp| dp.gross_capacity / rated_cap }
+        eir_ft_output_values = data_speed.map { |dp| (1.0 / dp.gross_efficiency_cop) / rated_eir }
+        cap_fff_curve = create_table_lookup_constant(model, 1, "Cool-CAP-fFF#{i + 1}")
+        eir_fff_curve = create_table_lookup_constant(model, 1, "Cool-CAP-fFF#{i + 1}")
```
shorowit, line 2827:
```suggestion
        eir_fff_curve = create_table_lookup_constant(model, 1, "Cool-EIR-fFF#{i + 1}")
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2921,52 +2898,103 @@ def self.create_dx_cooling_coil(model, obj_name, cooling_system)
...
+        rate_dp = data_speed.find { |dp| (dp.indoor_temperature == rated_idb) && (dp.outdoor_temperature == rated_odb) }
+        rated_cap = rate_dp.gross_capacity
+        rated_eir = 1.0 / rate_dp.gross_efficiency_cop
+        htg_ap.heat_rated_eirs << rated_eir
+        htg_ap.heat_rated_capacities_net << rate_dp.capacity
+        htg_ap.heat_rated_capacities_gross << rated_cap
+        cap_ft_output_values = data_speed.map { |dp| dp.gross_capacity / rated_cap }
+        eir_ft_output_values = data_speed.map { |dp| (1.0 / dp.gross_efficiency_cop) / rated_eir }
+        cap_fff_curve = create_table_lookup_constant(model, 1, "Heat-CAP-fFF#{i + 1}")
+        eir_fff_curve = create_table_lookup_constant(model, 1, "Heat-CAP-fFF#{i + 1}")
```
shorowit, line 2947:
```suggestion
        eir_fff_curve = create_table_lookup_constant(model, 1, "Heat-EIR-fFF#{i + 1}")
```

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2022,7 +2035,18 @@ def self.process_heat_pump_adjustment(hvac_sizing_values, weather, hvac_heating,
       heating_db = @hpxml.header.manualj_heating_design_temp
     end
 
-    heat_cap_rated = (heating_load / MathTools.biquadratic(@heat_setpoint, heating_db, coefficients)) / capacity_ratio
+    if hvac_heating.compressor_type == HPXML::HVACCompressorTypeVariableSpeed
+      capacity_retention_temp_1speed, capacity_retention_fraction_1speed = HVAC.get_default_heating_capacity_retention(HPXML::HVACCompressorTypeSingleStage)
+      coefficients_1speed = HVAC.get_heat_cap_eir_ft_spec(HPXML::HVACCompressorTypeSingleStage, capacity_retention_temp_1speed, capacity_retention_fraction_1speed)[0][0]
+      heating_capacity_retention_temperature, heating_capacity_retention_fraction = HVAC.get_default_heating_capacity_retention(hvac_heating.compressor_type, hvac_heating.heating_efficiency_hspf)
```
shorowit, line 2041: This doesn't look right. Why are we always using the _default_ capacity retention? If the HPXML file has a user-entered capacity retention factor, shouldn't we use it?

shorowit: Perhaps this could be (untested code)....

```suggestion
      if hvac_heating.heating_capacity_retention_fraction.nil?
        heating_capacity_retention_temperature, heating_capacity_retention_fraction = HVAC.get_default_heating_capacity_retention(hvac_heating.compressor_type, hvac_heating.heating_efficiency_hspf)
      else
        heating_capacity_retention_temperature, heating_capacity_retention_fraction = hvac_heating.heating_capacity_retention_temp, hvac_heating.heating_capacity_retention_fraction 
      end
```

yzhou601: Yes that's here mostly because I used to set retention fraction and temperature to be nil when detailed performance data is provided, and the 17F capacity can also be entered instead of retention fractions. I think there's a better way to handle that now.

shorowit: Also, I'm seeing pretty similar equations (adjustment for indoor conditions) three times in hvac_sizing.rb. It should be consolidated into a method.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1151,7 +1151,7 @@ def self.apply_hvac(runner, hpxml, weather, convert_shared_systems)
         # Calculate heating capacity retention at lowest outdoor drybulb
         detailed_performance_data = heat_pump.heating_detailed_performance_data
         min_odb = detailed_performance_data.map { |dp| dp.outdoor_temperature }.min
-        next if min_odb == 47 # No temperatures below the rating temperature; don't bother to calculate a retention factor
+        min_odb = detailed_performance_data.map { |dp| dp.outdoor_temperature }.max if min_odb == 47 # No temperatures below the rating temperature;
```
yzhou601, line 1154: This will result in a retention fraction >1.0 at higher temperature, is it fine?

shorowit: Rather than calculate the retention factor at the lowest ODB (which was done out of convenience), I will update the code to always calculate the factor at 5F.

yzhou601:
@shorowit I took a look at the installation quality adjustment on hvac sizing, it's being called before apply_hvac_fixed_capacities (and where allow increased fixed capacity being implemented), it's adjusting the autosized capacity with iq, so the more I think of this, the more I feel that we actually don't need to change anything there. The autosized capacity is adjusted and then compared to the fixed capacity, and the allow_increased_fixed_capacity logic will be applied (detailed performance data will be scaled up if needed).

yzhou601:
Results running adaptive step sizes of 1.0, 0.5, 0.1
image
It's about 0.2MBtu difference over 9MBtu (base cooling energy) from step size 1.0 to 0.1.

shorowit:
TODO:
- [ ] Make sure default fanspeed/capacity ratios are not used when detailed performance inputs provided @yzhou601
- [ ] Add more unit tests, particularly around HVAC installation quality @yzhou601 @shorowit

PR #1324, opened 2023-03-17
Improve HP backup heating auto-sizing (e.g., when using MaxLoad)

### LINKED ISSUE #1322: Autosized HP backup capacity when using `MaxLoad` sizing
Currently we always size the HP backup based on the heating design load, regardless of the sizing methodology. For `ACCA` and `HERS`, the backup system will operate on the coldest day so that makes sense. But `MaxLoad` is intended to size the heat pump without needing backup, so we should probably just autosize to zero.

(It shouldn't really make a difference in terms of energy results, but it's confusing to see a large backup system for a HP sized for max load.)

shorowit:
## Pull Request Description

Closes #1322.

## Checklist

### DIFF
--- a/HPXMLtoOpenStudio/measure.xml
+++ b/HPXMLtoOpenStudio/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.0</schema_version>
   <name>hpxm_lto_openstudio</name>
   <uid>b1543b30-9465-45ff-ba04-1d1f85e763bc</uid>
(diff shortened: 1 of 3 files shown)


shorowit:
I assume we should really generalize this approach to size the HP backup by subtracting the HP capacity at the design temperature from the design heating load, and set this as the HP backup capacity. Need to factor in lockout/switchover temperatures. Should also consult with Manual S to see what it says about this, if anything.

PR #1325, opened 2023-03-17
Geb HVAC - variable speed systems

yzhou601:
## Pull Request Description

Add a csv path input for maximum capacity ratios. The variable speed system speed level will be capped by the ratio specified.

## Checklist


yzhou601:
@jmaguire1 Output from running base-hvac-air-to-air-heat-pump-var-speed-max-capacity-schedule.xml:

Note:

```
  Max unitary system speed level calculated by schedule (EXPECTED) = Schedule value * 4 (number of speeds)
  Unitary system speed level combined from output (ACTUAL) =
      IF Unitary System DX Coil Speed Level>1: 
          Unitary System DX Coil Speed Level - 1 + Unitary System DX Coil Speed Ratio, 
      ELSE: 
          Unitary System DX Coil Cycling Ratio
```
image
image

Unitary system speed level is well capped by the ratios specified in the csv schedule

yzhou601:
TODO:

- [ ] Limit the backup coil operation during the curtailment
    - General curtailment(70% < ratio < 100%): 4F off setpoint (Both curtailment and backup coil)
    - Critical curtailment(ratio < 70%): 62F indoor temperature
 - [ ] Add warning when the setpoint is below 62F and there's a critical curtailment(ratio < 70%)

yzhou601:
@jmaguire1 @shorowit I had a hard time finding a proper place to better document the changes and controls here. I just added the max capacity ratio to the schedule file path input table. Maybe a note below the table to describe this task will work?
@jmaguire1 I took a second thought on the warning, I think it's not very straightforward how to detect this scenario to throw the error, since both the setpoint and the max capacity ratio are schedules which will be read dynamically in the simulation. I feel that maybe explicitly documenting it to let users aware of this situation will be good enough for now?

yzhou601:
@jmaguire1 For overshooting setpoint, the explanation is that we're using the same sensor and actuator, which may cause circular issue. 
Discussion for solution: Read the predicted load, compare it to the % of maximum capacity and then actuate the speed level if it exceeds.

yzhou601:
@shorowit The CI error here is confusing to me. I dug into this for hours but didn't find a good way to fix it. It failed at the component load key comparison between 1x and 10x for my newly added test files. The 10x file doesn't have any heating natural ventilation component loads, but the 1x does have one data point in a year that reports a heating natural ventilation component loads in summer. It's weird that in August there's heating natural ventilation. I looked into the hourly simulation results and confirmed that there's no heating coil operation, and no natural ventilation cfm at that hour but the heating:energytransfer is non zero (which detects it as heating mode):
image
Lmk if you have any thoughts.

shorowit:
@yzhou601 Would it make sense to just disallow GEB HVAC models for now when unit multipliers are used? It's surely not a high priority to support, and we've done this for a few other areas (see notes/caveats [here](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#whole-sfa-mf-buildings)).

jmaguire1:
I'd vote for Scott's suggestion, I think having this work with unit multipliers is VERY low priority here.

jmaguire1:
Also wanted to include some example results. Here's the logic in action in the first 2 days of the year in one of our test files (base-hvac-mini-split-heat-pump-ducted-max-capacity-schedule.xml):

image
These days aren't all that extreme, but you can still see the logic in action. When you'd have to exceed the power requirement to be able to maintain setpoint (for example the first time the schedule hits 0.4) you see a drop in the HVAC power and a corresponding drop in the space temperature away from setpoint. In this example the HVAC doesn't have to run at more than 70% to maintain the space temperature, so during those events nothing changes.

Here's an example during a little more extreme weather. Same test file, but now looking at behavior on 2/8-2/9:
image
This is basically the coldest time of year, so you can see a little more of the logic in action. During the 40% events the backup coil is locked out, which you can see pretty clearly in the second 40% event. This means the main coil power drops, backup is off, and the space temperature starts dropping. In this case it didn't drop below setpoint - MITO, so the comfort constraint doesn't kick in to bring things back to setpoint.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1392,6 +1393,12 @@ def includes_col_name(col_name)
     return false
   end
 
+  def remove_col_name(col_name)
```
shorowit, line 1396: Doesn't seem like this is used? Delete it.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1350,6 +1350,7 @@ class SchedulesFile
   ColumnBattery = 'battery'
   ColumnBatteryCharging = 'battery_charging'
   ColumnBatteryDischarging = 'battery_discharging'
+  ColumnMaximumPowerRatio = 'maximum_power_ratio'
```
shorowit, line 1353: I think it's important to include HVAC in the name: `hvac_maximum_power_ratio`. Don't forget to update docs.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -7,12 +7,26 @@ class HVAC
...
                                          weather_max_drybulb, weather_min_drybulb,
-                                         control_zone, hvac_unavailable_periods)
+                                         control_zone, hvac_unavailable_periods, schedules_file, hpxml_bldg)
     is_heatpump = false
+    if not schedules_file.nil?
+      max_pow_ratio_sch = schedules_file.create_schedule_file(model, col_name: SchedulesFile::ColumnMaximumPowerRatio, schedule_type_limits_name: Constants.ScheduleTypeLimitsFraction)
+      # Not allowed with unit multiplier for now
+      if not max_pow_ratio_sch.nil?
+        fail 'Maximum power ratio schedules of variable speed hvac systems are only supported if NumberofUnits is 1.' if hpxml_bldg.building_construction.number_of_units > 1
+      end
```
shorowit, line 20: Good check, but please be consistent with other building technologies (see [here](https://github.com/NREL/OpenStudio-HPXML/blob/4f60b2e19cb413803022cbafef31d28e083f820e/HPXMLtoOpenStudio/resources/hvac.rb#L262-L266) and [here](https://github.com/NREL/OpenStudio-HPXML/blob/4f60b2e19cb413803022cbafef31d28e083f820e/HPXMLtoOpenStudio/resources/hvac.rb#L777-L781)). So: "NumberofUnits greater than 1 is not supported for variable speed HVAC systems with maximum power ratio schedules." And add the FUTURE comment and link to github issue like we do for the others.

workflow/tests/util.rb
```diff
@@ -14,6 +14,7 @@ def run_simulation_tests(xmls)
 
     next unless xml.include?('sample_files') || xml.include?('real_homes') # Exclude e.g. ASHRAE 140 files
     next if xml.include? 'base-bldgtype-mf-whole-building' # Already has multiple dwelling units
+    next if xml.include? 'max-power-ratio-schedule' # Unit multiplier not supported yet
```
shorowit, line 17: Note that the 10x case tests two things -- 1) running a whole SFA/MF building (2x) and 2) using unit multipliers (5x), which combine to 10x.

If unit multipliers is not allowed, that's fine, but we should still test running a whole SFA/MF building without them. So please remove this line and handle it the same way we handle [dehumidifiers and GSHPs](https://github.com/NREL/OpenStudio-HPXML/blob/4f60b2e19cb413803022cbafef31d28e083f820e/workflow/tests/util.rb#L42-L47), where we only exclude the unit multipliers from the test.

HPXMLtoOpenStudio/tests/util.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-def get_ems_values(ems_objects, name)
+def get_ems_values(ems_objects, name, parse_more_operators = false)
```
shorowit, line 3: Do we really need this argument? I.e., would things break if we exclude this argument and always parse more operators? (And if so, why?)

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -201,6 +216,13 @@ def self.apply_air_source_hvac_systems(model, cooling_system, heating_system,
 
     apply_installation_quality(model, heating_system, cooling_system, air_loop_unitary, htg_coil, clg_coil, control_zone)
 
+    if not max_pow_ratio_sch.nil?
+      # Only apply heating control when it's a heat pump
+      htg_coil = nil unless is_heatpump
+      htg_supp_coil = nil unless is_heatpump
+      apply_max_power_EMS(model, max_pow_ratio_sch, air_loop_unitary, control_zone, htg_supp_coil, clg_coil, htg_coil)
+    end
```
shorowit, line 224: It'd be nice to clean up the code because the business logic is spread out a bit. Right now we get the schedule and check for an error at the top of the method, provide a warning in the middle of the method, and apply the EMS at the end of the method. Can we just have a single `apply_max_power` method that is self contained and does all of this?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -54,6 +68,7 @@ def self.apply_air_source_hvac_systems(model, cooling_system, heating_system,
       end
     elsif (heating_system.is_a? HPXML::HeatingSystem) && (heating_system.heating_system_type == HPXML::HVACTypeFurnace)
       obj_name = Constants.ObjectNameFurnace
+      max_pow_ratio_sch = nil
```
shorowit, line 71: This seems odd to me. Won't it prevent the EMS program when you have a variable speed AC with an attached furnace? If so, seems like we should replace the sample file that has a standalone variable speed AC with this HVAC configuration.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1800,6 +1822,225 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+        clg_tot_sensor.setKeyName(coil.name.to_s)
+        clg_sens_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Cooling Coil Sensible Cooling Rate')
+        clg_sens_sensor.setName("#{coil.name} sens cooling rate")
+        clg_sens_sensor.setKeyName(coil.name.to_s)
+
+        program.addLine('If clg_mode > 0')
+        program.addLine("  If #{clg_tot_sensor.name} > 0")
+        program.addLine("    Set shr = #{clg_sens_sensor.name} / #{clg_tot_sensor.name}")
+        program.addLine('  Else')
+        program.addLine('    Set shr = 0.0')
```
shorowit, line 1944: Seems like this would be better as:
```suggestion
        program.addLine("    Set #{cap_multiplier} = #{clg_sens_sensor.name} / #{clg_tot_sensor.name}")
        program.addLine('  Else')
        program.addLine('    Set #{cap_multiplier} = 0.0')
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1800,6 +1822,225 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+        # Calculate capacity and eirs for later use of full-load power calculations at each stage
+        program.addLine('If htg_mode > 0')
+        program.addLine("  If #{outdoor_db_sensor.name} < 4.444444,")
+        program.addLine("    Set T_coil_out = 0.82 * #{outdoor_db_sensor.name} - 8.589")
+        program.addLine("    Set delta_humidity_ratio = @MAX 0 (#{outdoor_w_sensor.name} - (@WFnTdbRhPb T_coil_out 1.0 #{outdoor_bp_sensor.name}))")
+        program.addLine('    Set htg_frost_multiplier_cap = 0.909 - 107.33 * delta_humidity_ratio')
+        program.addLine('    Set htg_frost_multiplier_pow = 0.90 - 36.45 * delta_humidity_ratio')
+        program.addLine('  Else')
+        program.addLine('    Set htg_frost_multiplier_cap = 1.0')
+        program.addLine('    Set htg_frost_multiplier_pow = 1.0')
```
shorowit, line 1920: Same suggestion here:
```suggestion
        program.addLine("    Set #{cap_multiplier} = 0.909 - 107.33 * delta_humidity_ratio")
        program.addLine("    Set #{pow_multiplier} = 0.90 - 36.45 * delta_humidity_ratio")
        program.addLine('  Else')
        program.addLine("    Set #{cap_multiplier} = 1.0")
        program.addLine("    Set #{pow_multiplier} = 1.0")
```

Changelog.md
```diff
@@ -19,6 +19,7 @@ __New Features__
 - Adds window and skylight `GlassType` options of "low-e, high-solar-gain" and "low-e, low-solar-gain"; updates U-factor/SHGC lookup tables.
 - Updated water heater installation location defaulting to match ANSI 301-2022
 - Adds more error-checking for inappropriate inputs (e.g., HVAC SHR=0 or clothes washer IMEF=0).
+- Adds capability to model variable speed system power shedding controls in response to commands in the AHRI 1380 standard with maximum power ratio schedule
```
shorowit, line 22: How about making AHRI 1380 more of an example?
```suggestion
- Adds maximum power ratio detailed schedule for variable-speed HVAC systems (e.g., to model shedding controls per AHRI 1380).
```

docs/source/workflow_inputs.rst
```diff
@@ -634,6 +635,7 @@ The column names available in the schedule CSV files are:
   ``hot_water_fixtures``           frac     Fixtures (sinks, showers, baths) hot water use schedule.                           Yes
   ``heating_setpoint``             F        Thermostat heating setpoint schedule.                                              No
   ``cooling_setpoint``             F        Thermostat cooling setpoint schedule.                                              No
+  ``maximum_power_ratio``          frac     Variable speed system maximum power ratio schedule.                                No
```
shorowit, line 638: I suggest adding a footnote that refers to the AHRI standard as an example of what this is for; you could borrow language from the changelog.

shorowit (commented):
I didn't do any simulation tests or thorough investigation of the EMS, mainly focused on the coding implementation. It looks to be in good shape, just a few suggestions below.

Changelog.md
yzhou601: @jmaguire1 Is there an AHRI 1380 example?

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: This condition is for standalone furnace I think. So I added the test case for var speed ac+ attached furnace to make sure the cooling operation is added.

HPXMLtoOpenStudio/tests/util.rb
yzhou601: Yes it will break, basically the airflow tests. My first thought when it happened was that maybe it's because I am setting variables to be 1 if the statement.to_f is 0.0, but sometimes it's a true 0.0.

Changelog.md
shorowit: I'm not saying we need an AHRI 1380 example, just that AHRI 1380 itself is an example of what you can model with this new capability.

workflow/tests/util.rb
```diff
@@ -247,6 +250,9 @@ def _verify_outputs(rundir, hpxml_path, results, hpxml, unit_multiplier)
     if hpxml_path.include? 'base-hvac-multiple.xml'
       next if message.include? 'Reached a minimum of 1 borehole; setting bore depth to the minimum'
     end
+    if hpxml_path.include? 'base-hvac-furnace-gas-central-ac-var-speed-max-power-ratio-schedule.xml'
+      next if message.include? 'Maximum power ratio schedule is only supported for variable speed systems'
+    end
```
shorowit, line 255: I don't understand, why does this file generate a warning? It has a variable speed system.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1800,6 +1804,251 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
     end
   end
 
+  def self.apply_max_power_EMS(model, runner, hpxml_bldg, air_loop_unitary, control_zone, heating_system, htg_supp_coil = nil, clg_coil, htg_coil, schedules_file)
+    if not schedules_file.nil?
+      max_pow_ratio_sch = schedules_file.create_schedule_file(model, col_name: SchedulesFile::ColumnHVACMaximumPowerRatio, schedule_type_limits_name: Constants.ScheduleTypeLimitsFraction)
+      # Not allowed with unit multiplier for now
+      if not max_pow_ratio_sch.nil?
```
jmaguire1, line 1811: Is this the right way to trigger disallowing unit multipliers? I'd have thought you'd have to check for a multiplier here, not just if there's a schedule file and the max power column isn't there. Can we more directly look at the unit multiplier here?

jmaguire1 (commented):
Really only one minor thing I spotted... nice job Yueyue! :)

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: Yeah probably you're missing the "if" statement in the line1812, that's the checking for multiplier.

workflow/tests/util.rb
yzhou601: Probably it's not the best way to implement this, but if either heating/cooling or both is not variable speed, the schedule doesn't apply for that period of time, in this case, heating is not applied. Maybe we just ignore the heating by default w/o throwing any warning?

HPXMLtoOpenStudio/resources/hvac.rb
jmaguire1: Yup I sure did! Sorry about that, makes sense now.

workflow/tests/util.rb
shorowit: I would think you should check whether the **system** (i.e., heating _or_ cooling) has a multispeed coil and give a warning if the system doesn't have one. Don't independently check and warn about heating and cooling w/o multispeed coils.

yzhou601: Sounds fair to me. Addressed.

yzhou601:
> Also wanted to include some example results. Here's the logic in action in the first 2 days of the year in one of our test files (base-hvac-mini-split-heat-pump-ducted-max-capacity-schedule.xml):
> 
> image These days aren't all that extreme, but you can still see the logic in action. When you'd have to exceed the power requirement to be able to maintain setpoint (for example the first time the schedule hits 0.4) you see a drop in the HVAC power and a corresponding drop in the space temperature away from setpoint. In this example the HVAC doesn't have to run at more than 70% to maintain the space temperature, so during those events nothing changes.
> 
> Here's an example during a little more extreme weather. Same test file, but now looking at behavior on 2/8-2/9: image This is basically the coldest time of year, so you can see a little more of the logic in action. During the 40% events the backup coil is locked out, which you can see pretty clearly in the second 40% event. This means the main coil power drops, backup is off, and the space temperature starts dropping. In this case it didn't drop below setpoint - MITO, so the comfort constraint doesn't kick in to bring things back to setpoint.

Thanks Jeff. We talked about capacity schedule/ratios a lot before basically because we used to adopt the approach to actually cut capacity to certain percentage instead of power, but currently we changed to use power cap approach. All the names related are changed in this PR from `capacity_ratio` to `power_ratio`. Just to clarify so that Scott and others are not confused by the terms.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -7,12 +7,14 @@ class HVAC
...
-  def self.apply_air_source_hvac_systems(model, cooling_system, heating_system,
+  def self.apply_air_source_hvac_systems(model, runner, cooling_system, heating_system,
                                          sequential_cool_load_fracs, sequential_heat_load_fracs,
                                          weather_max_drybulb, weather_min_drybulb,
-                                         control_zone, hvac_unavailable_periods)
+                                         control_zone, hvac_unavailable_periods, schedules_file, hpxml_bldg)
     is_heatpump = false
+
     if not cooling_system.nil?
+      clg_ap = cooling_system.additional_properties
```
shorowit, line 17: This can be removed, it's no longer used here.

shorowit:
Also, there's an unchecked TODO above: "Add warning when the setpoint is below 62F and there's a critical curtailment(ratio < 70%)". Do we still need to (or want to) do this?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1804,6 +1808,252 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+    # No variable speed coil
+    if (clg_coil.nil?) && (htg_coil.nil?)
+      runner.registerWarning('Maximum power ratio schedule is only supported for variable speed systems.')
+    end
+
+    if (htg_coil.is_a? OpenStudio::Model::CoilHeatingDXMultiSpeed) && (heating_system.backup_type != HPXML::HeatPumpBackupTypeIntegrated)
+      htg_coil = nil
+      htg_supp_coil = nil
+      runner.registerWarning('Maximum power ratio schedule is only supported for integrated backup system. Schedule is ignored for heating.')
+    end
```
shorowit, line 1835: I suggest we remove this warning or make it more nuanced. It's common that we model MSHPs with backup electric baseboard and so we'll trigger this warning everytime. No one would expect the backup baseboard to use the power ratio schedule, so why produce a warning? It's more confusing than anything.

jmaguire1:
@yzhou601: Yes, that's correct. We're not going to include the hard limit at 62 F because everything else about controls is based on a temperature offset rather than an absolute temperature value, and what to do if you have a low heating setpoint (say 60 F) is super unclear how the standard is written right now. Hopefully future versions of this standard clarify, and if they do we can consider revisiting.

jmaguire1 (approved):

shorowit:
As I recall, there were two outstanding items from our last discussion:
1. We want to treat separate backup systems the same as integrated backup systems, but it's more effort. If we're not going to address it now, we should create a GitHub issue for it. And we should probably mention this limitation in the documentation.
2. There were questions raised around whether to disable the backup during recovery (and/or for a fixed period of time). What is the plan?

jmaguire1:
@shorowit: 

> As I recall, there were two outstanding items from our last discussion:
> 
> 1. We want to treat separate backup systems the same as integrated backup systems, but it's more effort. If we're not going to address it now, we should create a GitHub issue for it. And we should probably mention this limitation in the documentation.
> 2. There were questions raised around whether to disable the backup during recovery (and/or for a fixed period of time). What is the plan?

For #1, I think we should create a GitHub issue and address later, along with documentation. I think it's worth getting this merged in now so people can use the functionality with integrated backups, which is the most relevant use case, then worry about other scenarios.
For #2, I'm voting to not disable backup during recovery and leave things as is. The standard on how to control these units is silent on disabling backup during recovery, so that means it comes down to the manufacturer how they want to recover in that scenario. I actually was planning on highlighting this in the paper as a means of raising this issue for future versions of the standard, there's a few things in there that seem like they ought to be explicit or revised. Open to discussing if anyone strongly disagrees here though.

yzhou601:
Issue created: https://github.com/NREL/OpenStudio-HPXML/issues/1635

Is there anything else left here?

shorowit:
@yzhou601 Get the CI tests passing? 😛

docs/source/workflow_inputs.rst
```diff
@@ -650,11 +658,17 @@ The column names available in the schedule CSV files are:
...
   ================================  =======  =================================================================================  ===============================
 
+  .. [#] This feature is an advanced research capability. This schedule allows modeling shedding controls for variable speed HVAC systems (instead of setpoint changes) to limit the power of HVAC per `AHRI 1380 <https://www.ahrinet.org/search-standards/ahri-1380-i-p-demand-response-through-variable-capacity-hvac-systems-residential-and-small>`_. 
+         While any value can be entered, this is primarily intended to reflect the AHRI capabilities, which has two levels of load shed: "General Curtailment" and "Critical Curtailment". 
+         A "General Curtailment" signal prevents the equipment from running at more than 70% of max power and "Critical Curtailment" limits it to 40% of max power until comfort constraints are violated.
+         Values of 0.4 or below in the schedule are treated as "Critical Curtailment" events and values between 0.4-0.7 are treated as "General Curtailment" events.
```
shorowit, line 670: What happens when a value above 0.7 is provided? Also, should we say anything about the fact that any supplemental heating (integrated or separate) will continue to operate?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1814,6 +1817,259 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+    htg_spt_sensor.setKeyName(control_zone.name.to_s)
+    clg_spt_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Zone Thermostat Cooling Setpoint Temperature')
+    clg_spt_sensor.setName("#{control_zone.name} clg_spt_temp")
+    clg_spt_sensor.setKeyName(control_zone.name.to_s)
+    load_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Unitary System Predicted Sensible Load to Setpoint Heat Transfer Rate')
+    load_sensor.setName("#{air_loop_unitary.name} sens load")
+    load_sensor.setKeyName(air_loop_unitary.name.to_s)
+
+    # global variable
+    temp_offset_signal = OpenStudio::Model::EnergyManagementSystemGlobalVariable.new(model, 'temp_offset_signal')
```
shorowit, line 1866: I was going to leave a comment that this variable presumably needs to be changed from `"temp_offset_signal"` to `"#{air_loop_unitary.name} temp_offset_signal"` so that it doesn't get confused when there are multiple air loops, but now I'm realizing there's only a single detailed schedule column for the entire building. What is the intended behavior if there are multiple variable-speed HVAC systems in the building? If we want to disallow that, we need an error message, unit tests, and update docs.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1814,6 +1817,259 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+        outdoor_bp_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Site Outdoor Air Barometric Pressure')
+        outdoor_bp_sensor.setName('outdoor_bp')
+
+        # Calculate capacity and eirs for later use of full-load power calculations at each stage
+        program.addLine('If htg_mode > 0')
+        program.addLine("  If #{outdoor_db_sensor.name} < 4.444444,")
+        program.addLine("    Set T_coil_out = 0.82 * #{outdoor_db_sensor.name} - 8.589")
+        program.addLine("    Set delta_humidity_ratio = @MAX 0 (#{outdoor_w_sensor.name} - (@WFnTdbRhPb T_coil_out 1.0 #{outdoor_bp_sensor.name}))")
+        program.addLine("    Set #{cap_multiplier} = 0.909 - 107.33 * delta_humidity_ratio")
+        program.addLine("    Set #{pow_multiplier} = 0.90 - 36.45 * delta_humidity_ratio")
```
shorowit, line 1945: There are a lot of hard-coded numbers here... can we describe how they were derived or what they represent?

workflow/tests/base_results/results_bills.csv
```diff
@@ -15,7 +15,6 @@ base-appliances-wood.xml,1816.09,144.0,1220.28,0.0,1364.28,144.0,234.82,378.82,0
...
 base-atticroof-flat.xml,1787.49,144.0,1288.51,0.0,1432.51,144.0,210.98,354.98,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
-base-atticroof-radiant-barrier-location.xml,1545.61,144.0,1213.91,0.0,1357.91,144.0,43.7,187.7,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
```
shorowit, line 18: Weird, why is this row being deleted..?

shorowit (commented):
This looks really close, but I do have a few more comments/questions.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: Sure, these're from E+ source code or docs, we're reproduce the capacity and power calculation in E+ to avoid the lag from EMS programs. I'll add comments for that.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: I would assume we'll use the same schedule for all the var speed systems in the building, yes then the name needs to be specific to the air loop unitary system. I'll change it and maybe add a test file where there're multiple var speed systems in the same building?

workflow/tests/base_results/results_bills.csv
shorowit: Oh, I see what happened. This file was deleted, but some previous PR incorrectly brought it back during a merge. I'll fix it.

docs/source/workflow_inputs.rst
shorowit: > Also, should we say anything about the fact that any supplemental heating (integrated or separate) will continue to operate?

Note that this comment is specific to setpoint recovery. The question at hand is whether we should document how the backup heating should operate either during A) shedding events or B) setpoint recovery.

yzhou601: From the current docs, the `AHRI 1380` seems to only have 0.4 (critical), 0.7 (general) and 1.0 (no shedding) these three scenarios being sent. But our workflow implementation is more general than that, which accepts any fraction value. My interpretation is that Jeff is recommending the standard values that this feature is mainly for, I can add some more explanation here for the actual implementation.

yzhou601:
@jmaguire1 Could you take a look at Scott's comment and my doc changes to see if there's anything else to address?

shorowit (approved):
Works for me. Thanks for addressing all feedback. I will pull in master and then merge this once the CI comes back happy again.

PR #1340, opened 2023-03-31
Manufactured home belly ducts

nmerket:
## Pull Request Description

Adding mobile home belly and wing and ducts in the belly.

## Checklist


nmerket:
@shorowit This is just a start, but I wanted to check in with you that I'm on correct path here.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -1984,8 +1986,9 @@ def from_oga(hpxml)
 
   class Foundation < BaseElement
     ATTRS = [:id, :foundation_type, :vented_crawlspace_sla, :within_infiltration_volume,
-             :attached_to_slab_idrefs, :attached_to_floor_idrefs, :attached_to_foundation_wall_idrefs,
-             :attached_to_wall_idrefs, :attached_to_rim_joist_idrefs]
+             :skirt_present, :attached_to_slab_idrefs, :attached_to_floor_idrefs,
```
shorowit, line 1989: Should probably call this `belly_wing_skirt_present` or similar, to clarify that it only applies to that foundation type.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -2128,6 +2133,9 @@ def to_oga(doc)
         elsif @foundation_type == FoundationTypeCrawlspaceConditioned
           crawlspace = XMLHelper.add_element(foundation_type_el, 'Crawlspace')
           XMLHelper.add_element(crawlspace, 'Conditioned', true, :boolean)
+        elsif @foundation_type == FoundationTypeBellyAndWing
+          belly_and_wing = XMLHelper.add_element(foundation_type_el, 'BellyAndWing')
+          XMLHelper.add_element(belly_and_wing, 'SkirtPresent', @skirt_present, :boolean)
```
shorowit, line 2138: Are you going to make this a required or optional element? If optional, this would become:
```suggestion
          XMLHelper.add_element(belly_and_wing, 'SkirtPresent', @skirt_present, :boolean, @skirt_present_isdefaulted) unless @skirt_present.nil?
```
and then you'd default the value in `hpxml_defaults.rb`.

shorowit (commented):
Looks like a reasonable start. That may be all you have to do to handle the ducts, but you'll obviously have to make some additional changes to handle the belly/wing foundation type. I'd suggest manually creating an HPXML for testing -- you'll probably hit errors or, if it runs correctly, it'll probably have some wrong IDF properties. Do be able to run the HPXML file, you'll at a minimum need to update the schematron (EPvalidator.xml) to allow a belly and wing foundation type and duct location.

HPXMLtoOpenStudio/resources/hpxml.rb
nmerket: I'm leaving it as required unless you have a good reason I shouldn't. It seems like an important piece of information that shouldn't be defaulted.

shorowit: I don't really have a strong opinion. But some software tools may not collect this information, so you'd be forcing them to choose something (or add an input).

HPXMLtoOpenStudio/measure.rb
```diff
@@ -632,7 +632,16 @@ def self.add_floors(runner, model, spaces)
...
       if floor.is_interior
         surface.setSunExposure('NoSun')
-        surface.setWindExposure('NoWind')
+        if floor.exterior_adjacent_to == HPXML::LocationManufacturedHomeUnderBelly
+          foundation = @hpxml.foundations.find {|foundation| foundation.to_location == floor.exterior_adjacent_to}
+          if foundation.belly_wing_skirt_present
+            surface.setWindExposure('NoWind')
+          else
+            surface.setWindExposure('WindExposed')
+          end
```
nmerket, line 641: @shorowit regarding our conversation earlier about not requiring a `Foundation` element for OpenStudio-HPXML: I'm using it here to determine whether the under belly should have exposure to wind. I guess that'd mean we need a default assumption about the skirt if there's no `Foundation` available. What do you think that should be?

shorowit: I'd probably assume there's a skirt.

shorowit (commented):
Nice work, @nmerket. This all looks pretty reasonable to me but I do have some suggestions/questions.

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -424,6 +424,14 @@ def self.get_temperature_scheduled_space_values(location)
...
                ground_weight: 1.0,
                f_regain: 0.83 } # From LBNL's "Technical Background for default values used for Forced Air Systems in Proposed ASHRAE Standard 152P"
+    elsif location == HPXML::LocationManufacturedHomeBelly
+      # From LBNL's "Technical Background for default values used for Forced Air Systems in Proposed ASHRAE Standard 152P"
+      # 3.5 Manufactured House Belly Pan Temperatures
+      return { temp_min: nil,
+               indoor_weight: 1.0,
+               outdoor_weight: 0.0,
+               ground_weight: 0.0,
+               f_regain: 0.62 }
```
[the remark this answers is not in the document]
shorowit: I will send ORNL an email to see if they have any quick thoughts/feedback.

shorowit (commented):
I was comparing the base-foundation-ambient.xml and base-foundation-belly-wing-skirt.xml files and noticed a difference:
image

Looks like [this line](https://github.com/NREL/OpenStudio-HPXML/blob/017aa799faaf0f04c7f54da85fc2797bd988bc62/BuildResidentialHPXML/measure.rb#L3874) needs to be updated.

shorowit (commented):
I'm also seeing a change to gable vs exterior wall areas that doesn't seem right:

image

docs/source/workflow_inputs.rst
```diff
@@ -1003,7 +1009,7 @@ Each floor/ceiling surface that is not in contact with the ground (Slab) nor adj
   ``Insulation/AssemblyEffectiveRValue``  double    F-ft2-hr/Btu  > 0          Yes                 Assembly R-value [#]_
   ======================================  ========  ============  ===========  ========  ========  ============================
 
-  .. [#] ExteriorAdjacentTo choices are "outside", "attic - vented", "attic - unvented", "basement - conditioned", "basement - unconditioned", "crawlspace - vented", "crawlspace - unvented", "crawlspace - conditioned", "garage", "other housing unit", "other heated space", "other multifamily buffer space", or "other non-freezing space".
+  .. [#] ExteriorAdjacentTo choices are "outside", "attic - vented", "attic - unvented", "basement - conditioned", "basement - unconditioned", "crawlspace - vented", "crawlspace - unvented", "crawlspace - conditioned", "garage", "other housing unit", "other heated space", "other multifamily buffer space", "other non-freezing space", or "manufactured home underbelly".
```
shorowit, line 1012: I don't know how best to do this, but it seems like we need to give some guidance on what area and R-value to use. Is it the horizontal floor area adjacent to conditioned space or the surface area of the rounded/rectangular belly adjacent to the underbelly? How does that choice influence the assembly R-value that should be used? I guess I'm just reiterating [this comment](https://github.com/hpxmlwg/hpxml/pull/360#issuecomment-1504059177) because I think it should be called out in the OS-HPXML documentation too.

nmerket:
> I'm also seeing a change to gable vs exterior wall areas that doesn't seem right:
> 
> image

Hmm. I wonder if this one is because I elongated the building to give it an aspect ratio more like a mobile home. Is it possible that the gable is facing the long side? How would I address that?

shorowit:
> Hmm. I wonder if this one is because I elongated the building to give it an aspect ratio more like a mobile home. Is it possible that the gable is facing the long side? How would I address that?

Ah right, the aspect ratio must be the cause. But as far as I know, the gable should always be on the short side, so it sure seems like the gable area should be smaller in your file compared to base-foundation-ambient.xml. @joseph-robertson, can you help?

shorowit (commented):
The new "mobile home belly" enumeration should be added to HPXML's `UnitLocation` too, so it can be used for the heating/cooling system location.

joseph-robertson:
> > Hmm. I wonder if this one is because I elongated the building to give it an aspect ratio more like a mobile home. Is it possible that the gable is facing the long side? How would I address that?
> 
> Ah right, the aspect ratio must be the cause. But as far as I know, the gable should always be on the short side, so it sure seems like the gable area should be smaller in your file compared to base-foundation-ambient.xml. @joseph-robertson, can you help?

The gable area _is_ smaller in `base-foundation-belly-wing-no-skirt.xml` compared to `base-foundation-ambient.xml` (note in the above side-by-side diff `base-foundation-belly-wing-no-skirt.xml` is on the LHS and `base-foundation-ambient.xml` is on the RHS).

`base-foundation-belly-wing-no-skirt.xml`
image

`base-foundation-ambient.xml`
image

nmerket:
@joseph-robertson That looks right to me. Are you using sketchup and the old OpenStudio plugin to do those renderings or something else?

joseph-robertson:
> @joseph-robertson That looks right to me. Are you using sketchup and the old OpenStudio plugin to do those renderings or something else?

I used [OpenStudioApplication](https://github.com/openstudiocoalition/OpenStudioApplication).

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -706,7 +713,7 @@
...
+      <sch:assert role='ERROR' test='h:UnitLocation[text()="living space" or text()="basement - unconditioned" or text()="basement - conditioned" or text()="attic - unvented" or text()="attic - vented" or text()="garage" or text()="crawlspace - unvented" or text()="crawlspace - vented" or text()="crawlspace - conditioned" or text()="other exterior" or text()="other housing unit" or text()="other heated space" or text()="other multifamily buffer space" or text()="other non-freezing space" or text()="roof deck" or text()="unconditioned space" or text()="manufactured home belly"] or not(h:UnitLocation)'>Expected UnitLocation to be 'living space' or 'basement - unconditioned' or 'basement - conditioned' or 'attic - unvented' or 'attic - vented' or 'garage' or 'crawlspace - unvented' or 'crawlspace - vented' or 'crawlspace - conditioned' or 'other exterior' or 'other housing unit' or 'other heated space' or 'other multifamily buffer space' or 'other non-freezing space' or 'roof deck' or 'unconditioned space' or 'manufactured home belly'</sch:assert>
```
shorowit, line 716: Need to update the documentation for these UnitLocation changes.

nmerket:
> The new "mobile home belly" enumeration should be added to HPXML's `UnitLocation` too, so it can be used for the heating/cooling system location.

Do I need to make any modeling code changes to make that work, or will it happen automagically?

shorowit:
> > The new "mobile home belly" enumeration should be added to HPXML's `UnitLocation` too, so it can be used for the heating/cooling system location.
> 
> Do I need to make any modeling code changes to make that work, or will it happen automagically?

I'm not sure. I suggest adding it to one of your base-foundation-belly-wing-foo.xml sample files and try running it.

nmerket:
Heating and cooling system location aren't in the BuildResidentialHPXML measure.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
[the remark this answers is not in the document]
nmerket: This is done here

https://github.com/NREL/OpenStudio-HPXML/blob/b97255638c6f3dd1c7d7d93861db0d9345c8c652/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml#L541-L546

workflow/tests/base_results/results_ashrae_140.csv
```diff
@@ -1,27 +1,27 @@
 Test Case,Annual Heating Load [MMBtu],Annual Cooling Load [MMBtu]
-L100AC.xml,51.41,N/A
```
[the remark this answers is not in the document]
nmerket: Addressed in https://github.com/NREL/OpenStudio-HPXML/pull/1340/commits/d7cb726c5e6aad041f55bb24d829e4a8be744442. I ended up going with option 2 above, even though option 1 was my preference because it seemed to fit the description more correctly. Bringing the house back down to ground level yielded the same results, so I went with that, left the `NoWind` thing in, but commented out.

nmerket:
Agenda for discussion:

- ASHRAE 140 assumptions with no wind under the floor adjacent to ambient. Does my solution work?
- Make a plan for ducts in the belly space. I don't really want to make the mobile home belly a separate zone at this time. I have way too much information from ORNL now and I could use your help making a simple decision based on it. 
- What other checklist items need to be sorted out to get this across the finish line?

docs/source/workflow_inputs.rst
nmerket: We could document for OpenStudio-HPXML that this should include an adjustment and calculate that in OpenStudio-HEScore.

HPXMLtoOpenStudio/resources/geometry.rb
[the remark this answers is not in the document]
nmerket: We're going to punt on this for now and use the duct temperatures and regain we already have. We can circle back with ORNL for a more detailed approach in the future.

workflow/tests/base_results/results_ashrae_140.csv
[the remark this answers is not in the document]
shorowit: I looked at this briefly. The reason that your commented line has no effect is because it isn't changing the IDF. The code is already setting NoWind a few lines further below in the code ([here](https://github.com/NREL/OpenStudio-HPXML/blob/mobile_home_belly_ducts/HPXMLtoOpenStudio/measure.rb#L662)). So you can safely remove the commented line.

Presumably all we are seeing then is a difference in wind speed due to the height of the conditioned zone (and not a difference in the raised floor heat transfer due to wind). I see nothing in ASHRAE 140 that specifies the above-grade height of the raised floor, so I suggest we stick with what we've been using (zero).

docs/source/workflow_inputs.rst
```diff
@@ -1000,14 +1006,18 @@ Each floor/ceiling surface that is not in contact with the ground (Slab) nor adj
...
          See :ref:`hpxmllocations` for descriptions.
   .. [#] FloorType child element choices are ``WoodFrame``, ``StructuralInsulatedPanel``, ``SteelFrame``, or ``SolidConcrete``.
   .. [#] InteriorFinish/Type choices are "gypsum board", "gypsum composite board", "plaster", "wood", "other", or "none".
   .. [#] InteriorFinish/Type defaults to "gypsum board" if InteriorAdjacentTo is living space and the surface is a ceiling, otherwise "none".
-  .. [#] AssemblyEffectiveRValue includes all material layers, interior/exterior air films, and insulation installation grade.
+  .. [#] AssemblyEffectiveRValue includes all material layers, interior/exterior
+    air films, and insulation installation grade. For a manufactured home belly
+    where the area of the belly wrap is different and usually greater than the
+    floor area, the AssemblyEffectiveRValue should be adjusted to account for
+    the surface area of the belly wrap and insulation.
```
nmerket, line 1020: Updated docs for AssemblyRValue to give direction for mobile home bellies.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -602,20 +602,35 @@ def self.apply_attics(hpxml)
...
+      belly_and_wing_foundations = hpxml.foundations.select { |f| f.foundation_type == HPXML::FoundationTypeBellyAndWing }
+      if belly_and_wing_foundations.empty?
+        hpxml.foundations.add(id: 'BellyAndWing',
+                              foundation_type: HPXML::FoundationTypeBellyAndWing)
+        belly_and_wing_foundations << hpxml.foundations[-1]
+      end
+      belly_and_wing_foundations.each do |foundation|
+        if foundation.belly_wing_skirt_present.nil?
+          foundation.belly_wing_skirt_present_isdefaulted = true
+          foundation.belly_wing_skirt_present = true
```
shorowit, line 631: This is an edge case, but need to add a `break` here like is done in the above example for vented crawlspace. If a HPXML file is submitted with two Foundation elements, neither of which have SkirtPresent specified, they would both default to true in the `in.xml`. But then if you tried to run that file, you'd get an error because [we only allow a single SkirtPresent](https://github.com/NREL/OpenStudio-HPXML/blob/mobile_home_belly_ducts/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml#L544).

(Yeah, it's stupid that we can't allow a HPXML file w/ SkirtPresent=true showing up twice. If you can figure out a better way to handle this, I'm all ears. We're obviously trying to protect against the situation where you have SkirtPresent=true and SkirtPresent=false in the same file.)

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -424,6 +424,14 @@ def self.get_temperature_scheduled_space_values(location)
                outdoor_weight: 0.0,
                ground_weight: 1.0,
                f_regain: 0.83 } # From LBNL's "Technical Background for default values used for Forced Air Systems in Proposed ASHRAE Standard 152P"
+    elsif location == HPXML::LocationManufacturedHomeBelly
+      # From LBNL's "Technical Background for default values used for Forced Air Systems in Proposed ASHRAE Standard 152P"
+      # 3.5 Manufactured House Belly Pan Temperatures
```
shorowit, line 429:
```suggestion
      # From LBNL's "Technical Background for default values used for Forced Air Systems in Proposed ASHRAE Standard 152P"
      # 3.5 Manufactured House Belly Pan Temperatures
      # FUTURE: Consider modeling the belly as a separate thermal zone so that we dynamically calculate temperatures.
```

shorowit (commented):
Nice work. This looks pretty close. I have a few comments, but really only two substantive questions (floor exterior_adjacent_to enumeration and water heaters in belly).

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -519,7 +519,7 @@
...
+      <sch:assert role='ERROR' test='h:ExteriorAdjacentTo[text()="outside" or text()="attic - vented" or text()="attic - unvented" or text()="basement - conditioned" or text()="basement - unconditioned" or text()="crawlspace - vented" or text()="crawlspace - unvented" or text()="crawlspace - conditioned" or text()="garage" or text()="other housing unit" or text()="other heated space" or text()="other multifamily buffer space" or text()="other non-freezing space" or text()="manufactured home underbelly"] or not(h:ExteriorAdjacentTo)'>Expected ExteriorAdjacentTo to be 'outside' or 'attic - vented' or 'attic - unvented' or 'basement - conditioned' or 'basement - unconditioned' or 'crawlspace - vented' or 'crawlspace - unvented' or 'crawlspace - conditioned' or 'garage' or 'other housing unit' or 'other heated space' or 'other multifamily buffer space' or 'other non-freezing space'</sch:assert>
```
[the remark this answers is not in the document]
nmerket: Right, the intention is as you describe it. I think it's confusing either way, so there's no real good choice.

docs/source/workflow_inputs.rst
```diff
@@ -1320,7 +1330,7 @@ Each heating system (other than a heat pump) is entered as an ``/HPXML/Building/
   ``FractionHeatLoadServed``         double    frac    0 - 1 [#]_   See [#]_                  Fraction of heating load served
   =================================  ========  ======  ===========  ========  ==============  ===============================
 
-  .. [#] UnitLocation choices are "living space", "basement - unconditioned", "basement - conditioned", "attic - unvented", "attic - vented", "garage", "crawlspace - unvented", "crawlspace - vented", "crawlspace - conditioned", "other exterior", "other housing unit", "other heated space", "other multifamily buffer space", "other non-freezing space", "roof deck", or "unconditioned space".
+  .. [#] UnitLocation choices are "living space", "basement - unconditioned", "basement - conditioned", "attic - unvented", "attic - vented", "garage", "crawlspace - unvented", "crawlspace - vented", "crawlspace - conditioned", "other exterior", "other housing unit", "other heated space", "other multifamily buffer space", "other non-freezing space", "roof deck", "manufactured home belly", or "unconditioned space".
```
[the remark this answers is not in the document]
nmerket: I have updated the docs to indicate that we allow "manufactured home belly" for HVAC systems. While it is allowed in HPXML, I removed allowing it for water heaters in EPvalidator. It hadn't been updated in the docs for water heaters, so no change there. To recap:

| Equipment | Allowed in belly in EPValidator.xml |
| --- | --- |
| HeatingSystem | ✅ |
| CoolingSystem | ✅ |
| HeatPump | ✅ |
| Water Heating System | 🚫 |

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1621,7 +1628,7 @@
...
+      <sch:assert role='ERROR' test='h:Location[text()="living space" or text()="basement - unconditioned" or text()="basement - conditioned" or text()="attic - unvented" or text()="attic - vented" or text()="garage" or text()="crawlspace - unvented" or text()="crawlspace - vented" or text()="crawlspace - conditioned" or text()="manufactured home belly" or text()="other exterior" or text()="other housing unit" or text()="other heated space" or text()="other multifamily buffer space" or text()="other non-freezing space"] or not(h:Location)'>Expected Location to be 'living space' or 'basement - unconditioned' or 'basement - conditioned' or 'attic - unvented' or 'attic - vented' or 'garage' or 'crawlspace - unvented' or 'crawlspace - vented' or 'crawlspace - conditioned' or 'manufactured home belly' or 'other exterior' or 'other housing unit' or 'other heated space' or 'other multifamily buffer space' or 'other non-freezing space'</sch:assert>
```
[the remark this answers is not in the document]
nmerket: This has been disallowed in EPvalidator.xml. (Although, it works in the model now if we ever find out that's a thing people do.)

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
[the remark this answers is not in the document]
nmerket: I'm going to leave this as is for better or for worse.

shorowit (approved):
I fixed the line endings (so that we don't have 350 HPXML files w/ diffs in GitHub, now or in the future) and found one more place where we needed to remove code for setting the water heater location to the belly.

This looks good to me. Will merge in if the CI comes back happy. Thanks, @nmerket!

Issue #1352, opened 2023-04-07
[Feature request] ReportUtilityBills option to also calculate by End Use


jmarrec:
`ReportSimulationOutput` has the capability, when `include_timeseries_end_use_consumptions` is true, to output hourly timeseries by Fuel and End Use.

It would be great to be able to leverage that in the `ReportUtilityBills` measure, so one could see, for each fuel type,  how much each end use contributes to the total, especially when a non flat structure is used (for a flat structure you can get the annual value by end use and multiply by the flat rate, but that's not true for TOU or tiered rates).

shorowit:
I suppose this could be allowed for flat and TOU rates, but I don't see how it would work for tiered rates. If your tiered structure is 10 cents/kWh up to 500 kWh for the month, and then 15 cents/kWh after, and you have End Use A that consumes 400 kWh and End Use B that consumes 200 kWh for the month, how would you attribute dollars to each? Suppose you simply apportioned them by their consumption like so:
Total $ = 500 * 0.1 + 100 * 0.15 = $65
End Use A $ = $65 * 400/600 = $43.33
End Use B $ = $65 * 200/600 = $21.67
That may seem fine, but then if you applied an upgrade that completely eliminated End Use B consumption, your utility bills would be reduced by $25, not $21.67. It seems confusing to me.

That said, it does make me think that we should at least offer monthly utility bill outputs by fuel type. Not sure why we don't do that.

shorowit:
Also tiered rates can _only_ be calculated monthly. Any other rates can be calculated hourly, I think.

jmarrec:
Yeah for tiered I think you'd apportion the monthly values per end use, then roll up to get the annual $/end use.
Getting an annual breakdown of cost per end use would be nice enough, monthly would be great. The hourly stuff is too granular to be meaningful anyways I think

PR #1365, opened 2023-04-19
ReportUtilityBills follow-up

### LINKED ISSUE #1364: Semi-annual run period bill calcs incorrect
Run `base-simcontrol-runperiod-1-month.xml` and the resulting `results_bills.csv` shows zero fixed costs. I believe this is because the `fuel_time_series` indexes [here](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportUtilityBills/resources/util.rb#L61) do not line up with [enumerating actual months Jan - Dec](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportUtilityBills/resources/util.rb#L420).

joseph-robertson:
## Pull Request Description

Closes #1364. Addresses #1352.

## Checklist

### DIFF
--- a/ReportUtilityBills/measure.rb
+++ b/ReportUtilityBills/measure.rb
@@ -238,23 +238,27 @@ class ReportUtilityBills < OpenStudio::Measure::ReportingMeasure
   end
 
   def report_runperiod_output_results(runner, utility_bills, output_format, output_path, bill_scenario_name)
-    segment = utility_bills.keys[0].split(':', 2)[0]
-    segment = segment.strip
+    line_break = nil
 
     results_out = []
     results_out << ["#{bill_scenario_name}: Total (USD)", utility_bills.values.sum { |bill| bill.annual_total.round(2) }.round(2)]
 
     utility_bills.each do |fuel_type, bill|
-      new_segment = fuel_type.split(':', 2)[0]
-      new_segment = new_segment.strip
-      if new_segment != segment
-        segment = new_segment
-      end
-
       results_out << ["#{bill_scenario_name}: #{fuel_type}: Fixed (USD)", bill.annual_fixed_charge.round(2)]
       results_out << ["#{bill_scenario_name}: #{fuel_type}: Energy (USD)", bill.annual_energy_charge.round(2)]
       results_out << ["#{bill_scenario_name}: #{fuel_type}: PV Credit (USD)", bill.annual_production_credit.round(2)] if [FT::Elec].include?(fuel_type)
       results_out << ["#{bill_scenario_name}: #{fuel_type}: Total (USD)", bill.annual_total.round(2)]
+    end
+    results_out << [line_break]
+
+    utility_bills.each do |fuel_type, bill|
+      bill.monthly_fixed_charge.each_with_index do |monthly_fixed_charge, i|
+        results_out << ["#{bill_scenario_name}: #{i + 1}: #{fuel_type}: Fixed (USD)", monthly_fixed_charge.round(2)]
+      end
+      bill.monthly_energy_charge.each_with_index do |monthly_energy_charge, i|
+        results_out << ["#{bill_scenario_name}: #{i + 1}: #{fuel_type}: Energy (USD)", monthly_energy_charge.round(2)]
+      end
+      results_out << [line_break]
     end
 
     if ['csv'].include? output_format
--- a/ReportUtilityBills/measure.xml
+++ b/ReportUtilityBills/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.0</schema_version>
   <name>report_utility_bills</name>
   <uid>ca88a425-e59a-4bc4-af51-c7e7d1e960fe</uid>
-  <version_id>10456a16-b1fc-449f-b7f8-71fbcdabb1c9</version_id>
-  <version_modified>20230401T012721Z</version_modified>
+  <version_id>3f34fe30-c7c0-49ef-b8c9-225651d3895c</version_id>
+  <version_modified>20230419T155744Z</version_modified>
   <xml_checksum>15BF4E57</xml_checksum>
   <class_name>ReportUtilityBills</class_name>
   <display_name>Utility Bills Report</display_name>
@@ -237,7 +237,7 @@
       <filename>measure.rb</filename>
       <filetype>rb</filetype>
       <usage_type>script</usage_type>
-      <checksum>13808DAA</checksum>
+      <checksum>B2ED0FC2</checksum>
     </file>
   </files>
 </measure>


ReportUtilityBills/tests/utility_bills_test.rb
```diff
@@ -1199,8 +1199,11 @@ def _get_actual_bills(bills_csv)
       next if line.strip.empty?
 
       key, value = line.split(',').map { |x| x.strip }
+      next if (1..12).to_a.any? { |month| key.include?(": #{month}:") }
+
       actual_bills[key] = Float(value)
```
shorowit, line 1204: Rather than ignoring monthly outputs in the test, please check that the sum of the monthly results equals the expected annual result too.

Changelog.md
```diff
@@ -20,6 +20,7 @@ __New Features__
   - Msgpack outputs are no longer rounded (since there is no file size penalty to storing full precision).
   - Annual emissions and utility bills now include all fuel/end uses, even if zero.
   - ReportSimulationOutput measure: Allows disabling individual annual output sections.
+  - ReportUtilityBills measure: Outputs monthly fixed/energy charges.
```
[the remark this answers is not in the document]
joseph-robertson: Requires updates to `run_simulation.rb`?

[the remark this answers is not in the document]
shorowit: Yes. I don't think we need to expose cli arguments though, just enable them both.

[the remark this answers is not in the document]
shorowit: Well, if they are optional and default to true, then maybe no updates are needed.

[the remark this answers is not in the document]
joseph-robertson: I'm going to have annual default to true and monthly default to false. So will need to set monthly to true in `run_simulation.rb`?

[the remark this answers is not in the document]
shorowit: I would default them both to true, since there's no runtime penalty for getting the monthly values.

ReportUtilityBills/measure.rb
```diff
@@ -231,37 +257,54 @@ def run(runner, user_arguments)
...
+      report_runperiod_output_results(runner, args, utility_bills, output_path, utility_bill_scenario.name, @hpxml.header)
     end
 
     return true
   end
 
-  def report_runperiod_output_results(runner, utility_bills, output_format, output_path, bill_scenario_name)
-    segment = utility_bills.keys[0].split(':', 2)[0]
-    segment = segment.strip
+  def report_runperiod_output_results(runner, args, utility_bills, output_path, bill_scenario_name, _header)
```
shorowit, line 266: Was `header` supposed to be used? If not, remove it.

shorowit (commented):
Can you make zero outputs consistently use decimal places?

image

ReportUtilityBills/resources/util.rb
```diff
@@ -98,6 +101,12 @@ def self.simple(fuel_type, header, fuel_time_series, is_production, rate, bill,
       if net_elec < 0 # net producer, give credit at user specified rate
         bill.annual_production_credit += -net_elec * rate.net_metering_user_excess_sellback_rate
       end
+      if (bill.annual_production_credit > bill.annual_energy_charge) || (net_elec < 0)
```
shorowit, line 104: Why is this not just `if bill.annual_production_credit  != 0`? I don't really understand what is happening here.

ReportUtilityBills/measure.rb
joseph-robertson: Hmm... maybe I intended to get run period from it at one point? Not sure. I will remove.

ReportUtilityBills/resources/util.rb
joseph-robertson: I tried to pull the two terms from directly above: `bill.annual_production_credit > bill.annual_energy_charge` and `net_elec < 0`. If you had (previously) entered either of these blocks, move all annual production credit to the final month of the year.

joseph-robertson: I'll try your suggested simplification.

shorowit (commented):
I'm looking at the results for `base-misc-bills-pv.xml`:
- The first scenario (Net Metering w/ Wholesale Excess Rate) shows a negative *annual* PV credit
- The other two scenarios (Net Metering w/ Retail Excess Rate; Feed-In Tariff) show negative *monthly* PV credits.

It's not really clear to me when we should show monthly credits vs having the credits rollover until the end of the year, but I'm not confident that the current behavior is the right answer. Maybe @jmaguire1 knows, otherwise we should ask someone more knowledgeable about residential rates.

shorowit:
I also don't really like the format of the monthly outputs, just stringing them along one after another in a single column. It makes it hard to digest or visualize. It'd be better to produce a separate file (results_bills_monthly.csv?) where rows are months and columns are, e.g., Scenario 1: Electricity: Fixed, similar to the results_timeseries.csv file.

shorowit:
@jmaguire1 will try to see what the REopt team does here.

joseph-robertson:
> I'm looking at the results for `base-misc-bills-pv.xml`:
> 
> * The first scenario (Net Metering w/ Wholesale Excess Rate) shows a negative _annual_ PV credit
> * The other two scenarios (Net Metering w/ Retail Excess Rate; Feed-In Tariff) show negative _monthly_ PV credits.
> 
> It's not really clear to me when we should show monthly credits vs having the credits rollover until the end of the year, but I'm not confident that the current behavior is the right answer. Maybe @jmaguire1 knows, otherwise we should ask someone more knowledgeable about residential rates.

@ejhw Do you have any guidance here?

ejhw:
> > I'm looking at the results for `base-misc-bills-pv.xml`:
> > 
> > * The first scenario (Net Metering w/ Wholesale Excess Rate) shows a negative _annual_ PV credit
> > * The other two scenarios (Net Metering w/ Retail Excess Rate; Feed-In Tariff) show negative _monthly_ PV credits.
> > 
> > It's not really clear to me when we should show monthly credits vs having the credits rollover until the end of the year, but I'm not confident that the current behavior is the right answer. Maybe @jmaguire1 knows, otherwise we should ask someone more knowledgeable about residential rates.
> 
> @ejhw Do you have any guidance here?

@jmaguire1 were you able to track down what REopt does?

My understanding of the question is that it is just about reporting, not actually calculation of credits, right? If so, my 2 cents would be that it would be more intuitive to report generation credits monthly than at the end of the year.

joseph-robertson:
> > > I'm looking at the results for `base-misc-bills-pv.xml`:
> > > 
> > > * The first scenario (Net Metering w/ Wholesale Excess Rate) shows a negative _annual_ PV credit
> > > * The other two scenarios (Net Metering w/ Retail Excess Rate; Feed-In Tariff) show negative _monthly_ PV credits.
> > > 
> > > It's not really clear to me when we should show monthly credits vs having the credits rollover until the end of the year, but I'm not confident that the current behavior is the right answer. Maybe @jmaguire1 knows, otherwise we should ask someone more knowledgeable about residential rates.
> > 
> > 
> > @ejhw Do you have any guidance here?
> 
> @jmaguire1 were you able to track down what REopt does?
> 
> My understanding of the question is that it is just about reporting, not actually calculation of credits, right? If so, my 2 cents would be that it would be more intuitive to report generation credits monthly than at the end of the year, even if end-of-year true-up is common practice in actual billing.

@ejhw Yes just a question about reporting. So you'd propose, even after annual true-up, apportioning the total annual credits out over the 12 months?

ejhw:
> > > > I'm looking at the results for `base-misc-bills-pv.xml`:
> > > > 
> > > > * The first scenario (Net Metering w/ Wholesale Excess Rate) shows a negative _annual_ PV credit
> > > > * The other two scenarios (Net Metering w/ Retail Excess Rate; Feed-In Tariff) show negative _monthly_ PV credits.
> > > > 
> > > > It's not really clear to me when we should show monthly credits vs having the credits rollover until the end of the year, but I'm not confident that the current behavior is the right answer. Maybe @jmaguire1 knows, otherwise we should ask someone more knowledgeable about residential rates.
> > > 
> > > 
> > > @ejhw Do you have any guidance here?
> > 
> > 
> > @jmaguire1 were you able to track down what REopt does?
> > My understanding of the question is that it is just about reporting, not actually calculation of credits, right? If so, my 2 cents would be that it would be more intuitive to report generation credits monthly than at the end of the year, even if end-of-year true-up is common practice in actual billing.
> 
> @ejhw Yes just a question about reporting. So you'd propose, even after annual true-up, apportioning the total annual credits out over the 12 months?

When you put it that way, maybe it isn't a good idea or even possible to apportion to months?

shorowit:
@joseph-robertson Can you improve the monthly CSV file? It is weird to have column headers show up multiple times in the file...

image

Rather, just add the scenarios as additional columns. So the first XX columns are scenario 1, the next XX columns are scenario 2, etc. The file will always have 13 rows (header + 12 rows) for an annual simulation.

shorowit:
It would also be nice to make the file consistent with the results_timeseries.csv. I see two differences:
1. results_timeseries.csv puts units on the second row while results_bills_monthly.csv has units in the first row in parantheses.
2. results_timeseries.csv as a Time column while results_bills_monthly.csv has a month column.

image

image

shorowit:
It would be nice if the ReportUtilityBills measure allows `csv_dview` as an output format too, but if that's not straightforward, it could be a future enhancement.

joseph-robertson:
> It would also be nice to make the file consistent with the results_timeseries.csv. I see two differences:
> 
> 1. results_timeseries.csv puts units on the second row while results_bills_monthly.csv has units in the first row in parentheses.
> 2. results_timeseries.csv has a Time column while results_bills_monthly.csv has a month column.
> 
> image
> 
> image

For the second item, should we consider the timestamp convention (start or end)? If so, would we need to get ep timestamps from the msgpack data?

shorowit:
> > Rather, just add the scenarios as additional columns. So the first XX columns are scenario 1, the next XX columns are scenario 2, etc. The file will always have 13 rows (header + 12 rows) for an annual simulation.
> 
> I'm not totally sure I understand how this is an improvement. Won't the column headers still show up multiple times in the file (where the scenario name prefix is the only difference)? Also, I figured this format would be preferred because it sort of aligns with the annual bills format.

Maybe you're not understanding my suggestion? It would look like this, with one set of column headers:

Time | Scenario A: Type 1 | Scenario A: Type 2 | Scenario B: Type 1 | Scenario B: Type 2 | Scenario C: Type 1 | Scenario C: Type 2
-- | -- | -- | -- | -- | -- | --
  | USD | USD | USD | USD | USD | USD
2007-01-01T00:00:00 | … | … | … | … | … | …
2007-02-01T00:00:00 | … | … | … | … | … | …

ReportUtilityBills/measure.rb
```diff
@@ -150,6 +200,11 @@ def energyPlusOutputRequests(runner, user_arguments)
       end
     end
 
+    if args[:include_monthly_bills]
+      # force a eplusout_monthly.msgpack from which we get timestamps
+      result << OpenStudio::IdfObject.load('Output:Variable,*,Site Outdoor Air Drybulb Temperature,monthly;').get
+    end
```
shorowit, line 206: Why do you need to do this? The measure is already requesting monthly fuel outputs.

shorowit: Oh, I see. The monthly fuel requests are output meter requests and this is an output variable request. You should be able to remove this output variable request and obtain the timestamps in the eplusout.msgpack file, which is where output meter results end up. (Don't ask me why it's like this.)

joseph-robertson: Isn't it possible that eplusout.msgpack doesn't have monthly timestamps though? I.e., when a building consumes only electricity and has detailed electric rates defined.

ReportUtilityBills/tests/utility_bills_test.rb
```diff
@@ -1110,23 +1115,36 @@ def _load_timeseries(pv_size_kw, use_hourly_electricity)
...
+    @measure.get_monthly_output_results(args, utility_bills, utility_bill_scenario.name, monthly_data)
+
+    # Check written values exist and are registered
+    actual_monthly_bills = _get_actual_monthly_bills(monthly_data)
+
+    # Check sum of monthly equal to annual
+    actual_bills.keys.each do |bill|
+      assert(actual_monthly_bills.keys.include?(bill))
+      assert_in_delta(actual_bills[bill], actual_monthly_bills[bill].sum, 0.1) # within 10 cents
+    end
```
shorowit, line 1146: I ran various test files and noticed that the `base-simcontrol-runperiod-1-month.xml file` has all zeroes in the monthly bills output file. Can you make sure this test catches the error in that file, and then fix it?

shorowit:
@joseph-robertson When I set `include_annual_bills` to false, I still get a results_bills.csv file (albeit empty). When I set `include_monthly_bills` to false, it correctly avoids writing the file altogether.

shorowit (approved):

PR #1379, opened 2023-05-08
Infiltration flue enhancements

### LINKED ISSUE #1350: `HasFlueOrChimney` improvements
We currently have a `HasFlueOrChimney` input that is defaulted based on certain logic:

image

However, there is realistically one other condition that should be met -- that the combustion air comes from conditioned space. So we should probably rename `HasFlueOrChimney` to `HasFlueOrChimneyInConditionedSpace`, and, for defaulting purposes, check the location of the water heater and HVAC. (I don't think we currently ask for the HVAC location, but we could use, e.g., [HeatingSystem/UnitLocation](https://hpxml.nrel.gov/datadictionary/3.1.0/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem/UnitLocation).)

The above logic also says that we look for a non-electric fireplace entered as a `HeatingSystem`, but we should also check for a fireplace entered as a `FuelLoad`.

shorowit:
## Pull Request Description

**Breaking change**: Replaces `BuildingConstruction/extension/HasFlueOrChimney` with `AirInfiltration/extension/HasFlueOrChimneyInConditionedSpace`; defaults now incorporate HVAC/water heater location.

For HPXML files where the presence of flue/chimney was defaulted, some will now obtain lower infiltration rates if the location of the HVAC/water heater is outside conditioned space. Use `HVACPlant/*/UnitLocation` to specify where a given HVAC system is located, otherwise it will be defaulted based on its distribution system inputs.

Closes #1350.

## Checklist


shorowit:
- [ ] Add "roof" to element name, documentation. Remove wall furnace, others?

PR #1380, opened 2023-05-09
Schedule:File relative paths

joseph-robertson:
## Pull Request Description

Test https://github.com/NREL/OpenStudio/pull/4879.

## Checklist


shorowit:
@joseph-robertson Maybe I just don't follow what you're doing, but I wouldn't expect any changes are needed on the OS-HPXML side. OS-HPXML is already creating an OSM that has relative paths in it, they just need to be preserved in the IDF. It shouldn't matter what working directory is being used.

shorowit:
Okay, I partly take that back. OS-HPXML is not currently creating an OSM with a relative path, but doing so just means changing [this line](https://github.com/NREL/OpenStudio-HPXML/blob/4e5996a2318b8d22440b4c08f14f02887b03e7b4/HPXMLtoOpenStudio/resources/schedules.rb#L1509) to be a relative path. Changing the working directory is risky and should be avoided.

joseph-robertson:
I was trying to change the `output_dir` from abs to rel (which, correct, I'm now realizing I just need to make the `@output_schedules_path` relative). But where does the working directory get changed?

On a related note, this gets tricky. We need to pass a valid relative (to current working directory) path into `ScheduleFile`, but `ExternalFile` needs to be written with just the basename of that relative path. Right? Or can OSM's ExternalFile "File Name" be the original rel path, and it's up to FT to write out just the basename?

joseph-robertson:
I think we want both the OSM and IDF to have relative paths to the schedule file (alongside the OSM and IDF)?

joseph-robertson:
@shorowit

Not exactly sure how this should be reviewed. Test locally using installer at the bottom of https://github.com/NREL/OpenStudio/pull/4879? Request that a docker image be created so CI can handle it?

shorowit:
@joseph-robertson Are you trying to get https://github.com/NREL/OpenStudio/pull/4879 into OS 3.7, or is it for the following release?

joseph-robertson:
> @joseph-robertson Are you trying to get [NREL/OpenStudio#4879](https://github.com/NREL/OpenStudio/pull/4879) into OS 3.7, or is it for the following release?

@shorowit Sounds like OpenStudio will have a v3.7.0-rc2. And because https://github.com/NREL/OpenStudio/pull/4879 seems to be working, I think it makes sense to try getting it in this upcoming release.

shorowit:
Got it. If you're happy with how things are working, then it sounds like you can proceed w/ getting the OS PR merged for rc2. I assume you don't need the CI tests here to pass for that to happen. When rc2 is available, we can update the CI for this PR.

shorowit (approved):
Looks good. Will merge once the CI comes back green.

PR #1385, opened 2023-05-11
#1384 - Fallback to a WWR calculation when window placement fails

### LINKED ISSUE #1384: BuildResidentialHPXML: error fitting a window
While running ResStock with a certain buildstock.csv, I end up with this model, with a 3 car garage situation.

image

Due to the nature of the model, the only surface left to assign the wwr of 0.09 for the front facade is Surface 13.

https://github.com/NREL/OpenStudio-HPXML/blob/7bf7086324fe2bad934f6ab4ea498d8364cd68a8/BuildResidentialHPXML/resources/geometry.rb#L1267-L1284

```
window_area=17.124272146816143, window_gap_y=1.0, window_gap_x=0.2, window_aspect_ratio=1.333, max_single_window_area=12.0, facade=front
```

Surface 13 is 3.6 m², 38.75 ft². It tries to add a window area of 17 ft², so there's enough room to do so, but it fails because it tries to enforce a given window dimension / aspect.

I'm not sure what's the preferred way to handle it, but I could see:
* Gracefully letting it slide
* Fall back to using a Window to Wall Ratio approach, ignoring the requested window_aspect_ratio and all
    * **Note:** I've implemented that as a workaround for now, not sure whether you would want that upstreamed or not? A draft PR is at https://github.com/NREL/OpenStudio-HPXML/pull/1385

I'm not sure which files I need to provide for you to be able to reproduce minimally? There's the `./resstock/osw/Baseline/1.osw` that is produced, along with the initial buildstock.csv and resstock_project.yaml : https://gist.github.com/jmarrec/e62983549c0699405fc2c72411ab2599

jmarrec:
## Pull Request Description

* Potential Fix #1384
* When placing the Windows fails because BuildResidentialHPXML tries to enforce a certain aspect of window, fall back to using a given WWR: take the surfaces vertices, and scale them towards the surface centroid.

## Checklist

### DIFF
--- a/BuildResidentialHPXML/resources/geometry.rb
+++ b/BuildResidentialHPXML/resources/geometry.rb
@@ -1265,8 +1265,8 @@ class Geometry
   end
 
   def self.add_windows_to_wall(surface, window_area, window_gap_y, window_gap_x, window_aspect_ratio, max_single_window_area, facade, model, runner)
-    wall_width = get_surface_length(surface)
-    average_ceiling_height = get_surface_height(surface)
+    wall_width = get_surface_length(surface) # ft
+    average_ceiling_height = get_surface_height(surface) # ft
 
     # Calculate number of windows needed
     num_windows = (window_area / max_single_window_area).ceil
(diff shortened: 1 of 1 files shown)


BuildResidentialHPXML/resources/geometry.rb
```diff
@@ -1278,9 +1278,44 @@ def self.add_windows_to_wall(surface, window_area, window_gap_y, window_gap_x, w
...
     if width_for_windows > wall_width
-      runner.registerError("Could not fit windows on #{surface.name}.")
-      return false
+      surface_area = UnitConversions.convert(surface.grossArea, 'm^2', 'ft^2')
+      runner.registerWarning("Surface #{surface.nameString} has an area of #{surface_area}, params: window_area=#{window_area} ft^2, window_gap_y=#{window_gap_y} ft, window_gap_x=#{window_gap_x} ft, window_aspect_ratio=#{window_aspect_ratio}, max_single_window_area=#{max_single_window_area} ft^2, facade=#{facade}")
+      runner.registerWarning("Could not fit windows on #{surface.name}. Fall back to WWR")
+      wwr = window_area / surface_area
+      if wwr >= 0.90
+        wwr = 0.90
+      end
```
shorowit, line 1289: My reaction to this is:
1. We probably don't need a warning just because we are relaxing constraints around aspect ratio, window gap, etc. to fit the window area. If we are preserving the window area, the user is probably fine.
2. But we _do_ need a warning if we are going to reduce the window area.

So how about this instead?
```suggestion
      wwr = window_area / surface_area
      if wwr > 0.90
        runner.registerWarning("Could not fit windows on #{surface.name}; reducing window area to 90% WWR.")
        wwr = 0.90
      end
```

BuildResidentialHPXML/resources/geometry.rb
```diff
@@ -1278,9 +1278,44 @@ def self.add_windows_to_wall(surface, window_area, window_gap_y, window_gap_x, w
...
+
+      # Instead of using this
+      # ss_ = surface.setWindowToWallRatio(wwr, offset, true)
+      # We offset the vertices towards the centroid
```
shorowit, line 1293: I assume this is to maximize the likelihood of fitting the window area on the surface?

```suggestion
      # Instead of using this
      # ss_ = surface.setWindowToWallRatio(wwr, offset, true)
      # We offset the vertices towards the centroid to maximize the likelihood of fitting the window area on the surface
```

shorowit (commented):
Thanks for the contribution, @jmarrec. I'm generally fine with the approach you're taking here, just a couple comments.

Also, can you show what [this building](https://user-images.githubusercontent.com/5479063/237670058-e770992e-eb00-49e9-9987-c6a2fdd43839.png) looks like with your fix?

shorowit:
@joseph-robertson Any concerns with this?

joseph-robertson:
@jmarrec Curious if your sample came from an unmodified "certain buildstock csv" generated by ResStock, or from a modified one? Perhaps we should test this fix against ResStock's integration tests...

BuildResidentialHPXML/resources/geometry.rb
jmarrec: I'm made the change somewhere else and got the exact same feedback from someone else, so I guess we do need a warning :)

I didn't add it on purpose originally:

* Either you don't care, because it has no impact overall since you managed to add more window area to another wall for that facade
* Or you do care, but you'll get https://github.com/NREL/OpenStudio-HPXML/blob/9528831dd85ec80027ec93e0d994a7a76f86ea4f/BuildResidentialHPXML/resources/geometry.rb#L1151

jmarrec:
> @jmarrec Curious if your sample came from an unmodified "certain buildstock csv" generated by ResStock, or from a modified one? Perhaps we should test this fix against ResStock's integration tests...

We generate the buildstock CSV, this isn't sampled. Having a 3 car garage with this specific geometry might never happen in ResStock's sampling. (I hope I'm answering your question...?)

shorowit (approved):

PR #1387, opened 2023-05-11
Average resilience hours, try 2

joseph-robertson:
## Pull Request Description

Create a separate `results_resilience.csv` file (always hourly) that is optionally reported independent of the value entered for `timeseries_frequency`.

Add:
* "Resilience: Battery (avg-hr)" to `results_annual.csv`
* "Resilience: Battery (hr)" to `results_resilience.csv`

The idea is that this will accommodate future "Resilience: xxx (xxx)" output metrics.

ReportSimulationOutput:
* Add `include_annual_resilience` argument
* Add `include_resilience_hours` argument
* Add `resilience_output_file_name` argument
* Update ReportSimulationOutput test file
* Add `test_timeseries_hourly_resilience` test

`run_simulation.rb`:
* Update with `--add-resilience-hours` boolean argument
* Update template OSW files

## Checklist


jmaguire1:
Some interesting results here with different timesteps. I tried running base-pv-battery-scheduled (so we can be sure we're calculating resilience hours correctly off the net load if PV is present) at 3 different timesteps: 60 minutes (our default), 10 minute, and 1 minute (which we'll use when HVAC-GEB is enabled). There's some surprising differences in the timeseries resilience hours calculations:

image
The average resilience hours timeseries gets some pretty different results when looking at hourly timesteps vs. shorter timesteps, but 10 minute vs. 1 minute are identical. Looking into WHY this happens, it seems like the battery SoC as reported by E+ is the main difference:

image
I'm wondering if this has to do with EMS timestep lag? We're accounting for conversion losses using an EMS program, since the input to the SAM battery model wasn't working correctly and that'd take E+ changes. On an annual basis the results are pretty close:

<head>

<link rel=File-List
<style>

</style>
</head>

60 minute | 3.703
-- | --
10 minute | 3.697
1 minute | 3.697



</body>

</html>


If we think this is down to the EMS approach, does that make the necessary E+ changes a higher priority? Our immediate need is to use this with HVAC-GEB, so 1 minute timesteps, which I'm more confident are correct. But it's a little weird still.

shorowit:
@jmaguire1 will try re-running without the EMS program to see if that's the cause of the difference.

joseph-robertson:
Superseded by #1212.

PR #1404, opened 2023-05-24
Improve foundation translation, alternative

### LINKED ISSUE #976: Revisit foundation translation
The code related to translating HPXML foundation information into EnergyPlus is complicated and can result in slow simulations in more complex situations.

Some things to consider:
- Rather than using [every combination of Slab and FoundationWall](https://github.com/NREL/OpenStudio-HPXML/blob/17acfbcfe29bd77ddadabf7cabab8b6b7d3d2958/HPXMLtoOpenStudio/measure.rb#L2805) to map to a Foundation:Kiva, can't we just create one Foundation:Kiva per Slab? For a HPXML file with 2 Slabs and 4 FoundationWalls, that would result in two Foundation:Kiva objects instead of eight.
- There is currently no way to associate FoundationWalls with a given Slab, so we currently have to [apportion part of each FoundationWall to each slab](https://github.com/NREL/OpenStudio-HPXML/blob/17acfbcfe29bd77ddadabf7cabab8b6b7d3d2958/HPXMLtoOpenStudio/measure.rb#L1004-L1010). A couple ways to address this:
  - Require Slab/DepthBelowGrade for every foundation type (not just slab on grade). Require that every FoundationWall/DepthBelowGrade input match a Slab/DepthBelowGrade input and vice versa (except for `Slab/DepthBelowGrade`=0). Or...
  - Require the use of HPXML [Foundation](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Enclosure/Foundations/Foundation) elements with attached `Slab` and `FoundationWall` surfaces.

shorowit:
## Pull Request Description

Alternative to https://github.com/NREL/OpenStudio-HPXML/pull/1403.

Refactors/simplifies code translating HPXML foundation walls & slabs to OpenStudio; should result in faster runtime (fewer Kiva:Foundation objects) for some homes. **Breaking change**: Requires `DepthBelowGrade` for every slab and requires that every exterior foundation wall's `DepthBelowGrade` matches a corresponding Slab's `DepthBelowGrade`.

Closes #976.

## Checklist


shorowit:
This was a useful exercise, but I've come to realize that this isn't a complete solution. It can address runtime performance for some HPXMLs, but not others. A lot of the runtime performance is ultimately dictated by how a user or software tool constructs the HPXML file -- e.g., whether they describe a walkout basement using one foundation wall, or three, or ten. Also, it may not be practical to require software tools, or their users, to construct surfaces in a very rigid way (i.e., every exterior foundation wall's `DepthBelowGrade` must match a corresponding Slab's `DepthBelowGrade`).

A better solution would be to take multiple foundation walls that have all the same properties (height, insulation configuration, etc.) _other than below-grade depth_ and combine them into a single foundation wall with an effective below-grade depth that gives equivalent heat transfer.

I will pursue that approach in https://github.com/NREL/OpenStudio-HPXML/pull/1403, as well as move a few other things from this PR into that one.

PR #1420, opened 2023-06-23
Fixes schedules with mixed timesteps

### LINKED ISSUE #1419: Error if using schedules w/ mixed timesteps
Found by @jmaguire1 using BEopt ([BEopt issue](https://github.com/NREL/BEopt/issues/215)).

If you, say, run a 10-minute simulation where one schedule has 10-minute values and another schedule has hourly values, you receive a ruby error. It'd be nice to support this rather than working around the issue in BEopt.

The error is:
```
Error: element size differs (52560 should be 8760)
/mnt/c/git/openstudio-hpxml/HPXMLtoOpenStudio/resources/schedules.rb:1475:in `transpose'
/mnt/c/git/openstudio-hpxml/HPXMLtoOpenStudio/resources/schedules.rb:1475:in `block in export'
:/ruby/2.7.0/csv.rb:658:in `open'
/mnt/c/git/openstudio-hpxml/HPXMLtoOpenStudio/resources/schedules.rb:1473:in `export'
/mnt/c/git/openstudio-hpxml/HPXMLtoOpenStudio/resources/schedules.rb:1396:in `initialize'
```

It seems like the solution is to "expand" the schedules with fewer elements such that all the schedules have the same number of elements when exporting to CSV.

shorowit:
## Pull Request Description

Closes #1419.

## Checklist


HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1673,7 +1685,7 @@ def set_unavailable_periods(unavailable_periods)
           if schedule_name == ColumnWaterHeaterSetpoint
             # Temperature of tank < 2C indicates of possibility of freeze.
             @tmp_schedules[schedule_name][i] = UnitConversions.convert(2.0, 'C', 'F') if @tmp_schedules[column_name][i] == 1.0
-          else
+          elsif ![SchedulesFile::ColumnHeatingSetpoint, SchedulesFile::ColumnCoolingSetpoint].include?(schedule_name)
```
joseph-robertson, line 1688: We don't want to zero out heating/cooling setpoints (even though they get effectively ignored since outage is enforced using availability schedules).

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1641,6 +1642,17 @@ def create_column_values_from_periods(col_name, periods)
...
 
+  def expand_schedules
+    # Expand schedules with fewer elements such that all the schedules have the same number of elements
+    num_hrs_in_year = Constants.NumHoursInYear(@year)
+    n_ts_per_hr = @tmp_schedules.map { |_k, v| v.size / num_hrs_in_year }.uniq.max
+    @tmp_schedules.each do |col, values|
+      if values.size / num_hrs_in_year < n_ts_per_hr
+        @tmp_schedules[col] = values.map { |v| [v] * n_ts_per_hr }.flatten
+      end
+    end
```
shorowit, line 1653: Admittedly I haven't tested it, but this doesn't seem right. Why are you using `n_ts_per_hr` for all schedule columns regardless of how many elements they have? If there are three columns, one with 50k elements, one with 25k elements, and one with 5k elements, why would the 25k and 5k columns both get expanded by the same number of elements? Also, I don't see why `num_hrs_in_year` is needed, it cancels out in the `if` comparison.

Shouldn't it be something like this?
```suggestion
    max_size = @tmp_schedules.map { |_k, v| v.size }.uniq.max
    @tmp_schedules.each do |col, values|
      if values.size < max_size
        @tmp_schedules[col] = values.map { |v| [v] * max_size / values.size }.flatten
      end
    end
```

shorowit:
Tests look good, thanks @joseph-robertson. Will merge unless the CI comes back with unexpected results.

PR #1433, opened 2023-07-12
Split battery losses between charging/discharging

### LINKED ISSUE #1429: Split battery losses between charging/discharging
Can we update our EMS program to split the battery losses between charging and discharging? Currently we lump them all into charging. Did we already try splitting them between charging/discharging and ran into a limitation or problem? I don't see anything in [the PR](https://github.com/NREL/OpenStudio-HPXML/pull/1223) that suggests we tried it, but we might have.

See https://unmethours.com/question/94250/beopt-v3-bess-modeling/ for a BEopt user confused by this.

We won't need to do this if the [E+ efficiency input is connected](https://github.com/NREL/EnergyPlus/issues/9176), but that isn't going to happen in the immediate future.

@jmaguire1 @joseph-robertson

joseph-robertson:
## Pull Request Description

Closes #1429.

Update the `battery_losses` EMS program to move some of the losses during charging to discharging:
```
battery_losses_program.addLine("Set charge_losses = -1 * #{charge_sensor.name} * (1 - (#{battery.round_trip_efficiency} ^ 0.5))")
battery_losses_program.addLine("Set discharge_losses = -1 * #{discharge_sensor.name} * (1 - (#{battery.round_trip_efficiency} ^ 0.5))")
battery_losses_program.addLine('Set losses = charge_losses + discharge_losses')
```

image

## Checklist


jmaguire1:
Talking to a few people and looking at some other tools, it looks like the more common way to deal with splitting round trip efficiency into charge and discharge efficiency is that charging_efficiency = discharging_efficiency = round_trip_efficiency ** 0.5. See for example HOMER: https://www.homerenergy.com/products/pro/docs/3.11/battery_charge_efficiency.html

I think the original issue that raised this (that if you just sum the battery usage you'll get a positive value) will still occur even with the split losses. That was someone misunderstanding that battery energy includes the inefficiencies of the battery and not just the energy that goes into and out of the battery.

jmaguire1:
You might also want to look at this: https://sam.nrel.gov/forum/forum-general/2754-variation-of-round-trip-efficiency-with-battery-soc-range.html

Someone on the SAM forum pointed out how round trip efficiency changes a lot depending on the depth of discharge, with a very steep drop in efficiency as you get to higher depths of discharge. This is why we have inputs for usable vs. actual capacity, to avoid this level of charging/discharging, but in our workflow if we had round trip efficiency used as an input I'd expect we'd see similar results where the efficiency of each cycle depends on the depth of discharge (at least to some extent).

jmaguire1:
The last change I made to this was related to efficiency, namely that when you're going from round trip efficiency to charging and discharging efficiency what should that relationship be. I changed this to say the charging_efficiency = discharging_efficiency = round_trip_efficiency ^ 0.5. This is based on the assumption made in [HOMER ](https://www.homerenergy.com/products/pro/docs/3.11/battery_charge_efficiency.html), although I haven't found many other people either using this assumption or justifying it.

The reason this is even an issue is that battery efficiency is a rather complex function of how much you charge/discharge the battery. So in practice, the same Tesla powerwall cycled from the min/max SoC daily will have a different round trip efficiency than if that battery cycled between 40% and 70% SoC each day. Efficiency is also a function of battery temperature, which complicates things even further.

If we're trying to keep this pretty simple (at least until we can leverage SAM directly), it seems our two options from going from round trip efficiency to charging/discharging efficiency are to use the square root of round trip efficiency or split the losses evenly between charging and discharging. Turns out it's not that big an impact either way you do it, at least for reasonable values of round trip efficiency:

<head>

<link rel=File-List
<style>

</style>
</head>

Round   Trip Efficiency | Charging   Efficiency (square root) | Charing   Efficiency (Even Split)
-- | -- | --
99.00% | 99.50% | 99.50%
98.00% | 99.00% | 98.99%
97.00% | 98.50% | 98.49%
96.00% | 98.00% | 97.98%
95.00% | 97.50% | 97.47%
94.00% | 97.00% | 96.95%
93.00% | 96.50% | 96.44%
92.00% | 96.00% | 95.92%
91.00% | 95.50% | 95.39%
90.00% | 95.00% | 94.87%
89.00% | 94.50% | 94.34%
88.00% | 94.00% | 93.81%
87.00% | 93.50% | 93.27%
86.00% | 93.00% | 92.74%
85.00% | 92.50% | 92.20%
84.00% | 92.00% | 91.65%
83.00% | 91.50% | 91.10%
82.00% | 91.00% | 90.55%
81.00% | 90.50% | 90.00%
80.00% | 90.00% | 89.44%



</body>

</html>

jmaguire1:
I also talked to Michael Blonsky about this, here was his take on how to split round trip efficiency out:

"the square root is better. That's because the energy out = energy in * charge efficiency * discharge efficiency". Which intuitively makes sense to me, and it's not going to be a huge impact regardless.

@shorowit : if you're comfortable with using the square root of round trip efficiency I think this is ready to go. If you're not, lets talk about it.

joseph-robertson:
This is what the change looks like when we apportion half the calculated losses to charging and the other half to discharging (blue is master, orange is feature):
image

shorowit (approved):

shorowit:
Thanks @joseph-robertson and @jmaguire1.

PR #1456, opened 2023-08-01
Allow simulating whole SFA/MF buildings

### LINKED ISSUE #1376: Allow modeling whole SFA/MF buildings
Currently OS-HPXML only allows modeling individual dwelling units of a SFA/MF building (including workarounds for MF common spaces adjacent to the unit or units with shared HVAC/DHW/ventilation systems, per ANSI 301).

However, there are use cases where it would be highly beneficial to create whole MF building models. The easiest path from here to there would be to allow a HPXML file where each dwelling unit is described as an individual `Building` element. The HPXMLtoOpenStudio measure could be updated to create an individual model for each `Building` element (as it does now) and then merge all the models into a single whole-building model (see [example code from URBANopt](https://github.com/urbanopt/urbanopt-example-geojson-project/blob/develop/example_project/measures/BuildResidentialModel/measure.rb#L160-L336)) at the end. Ideally, the BuildResidentialHPXML measure would also be updated to produce HPXML files like this.

(There could be another use case where someone wants to provide an HPXML in which a single `Building` element describes the entire MF building -- e.g., 30 units, 3 stories, 60 bedrooms, 20000 sqft, etc. -- but that would be a greater departure from how OS-HPXML works and far more effort.)

Some things to consider for this approach:
- Common spaces and surface adjacency
- Shared/central systems
- Unit multipliers
- Coordination w/ openstudio-standards and other existing products
- Output reporting

### LINKED ISSUE #1450: Use `AdditionalProperties` instead of object name comparisons
OS-HPXML originally had to rely on object names (e.g., `Constants.ObjectNameFoo`) in many situations to understand what building component a, e.g., `ElectricEquipment` or `InfiltrationDesignFlowRate` object represents. But this comes with some risk because object names may end up with various string prefixes/suffixes (e.g., if there are multiple objects of the same type), or get converted to snake_case, etc.

Now that the OpenStudio `AdditionalProperties` object exists, we should audit our code and start using it. It will eliminate this risk once and for all.

### LINKED ISSUE #1492: Revisit HPXML header inputs for whole MF building modeling
Certain inputs get stored in the HPXML header -- i.e., outside of the individual `Building` elements. Now that we're going to allow modeling whole MF buildings, that means we can't vary those inputs for each `Building` (dwelling unit).

We should review those inputs and see which should be moved into the individual `Building` elements so they can vary across dwelling units. For example, simulation control settings (run period, timestep, etc.) should stay where they are, but most everything else should move.

shorowit:
## Pull Request Description

Closes #1376. Closes #1450 and closes #1492.

If an HPXML file is submitted where there are multiple `Building` elements, each representing an individual dwelling unit, you can now tell OS-HPXML that you want to run ALL units, which creates a single combined OS/E+ model. The ability to just run one of the `Building` IDs is also still available.

Also allows dwelling unit multipliers via `BuildingConstruction/NumberofUnits` (as described [here](https://github.com/hpxmlwg/hpxml/pull/316)) for runtime performance reasons. When provided, simulation outputs will reflect this multiplier. When modeling individual dwelling units, there isn't that much of a benefit since all the OS-HPXML outputs could simply be multiplied by this value by the end user/software during a post-processing. But when [modeling whole SFA/MF buildings](https://github.com/NREL/OpenStudio-HPXML/pull/1446), different units in the building may have different multipliers, so it cannot be performed during a post-processing step.

**Breaking change**: Multiple elements move from `SoftwareInfo/extension` to `BuildingDetails/BuildingSummary/extension` to allow variation across MF dwelling units:
    - `HVACSizingControl`
    - `ShadingControl`
    - `SchedulesFilePath`
    - `NaturalVentilationAvailabilityDaysperWeek`

Finally, the HPXML class has been updated to support multiple buildings. Previously:
```
hpxml = HPXML.new(...)
hpxml.windows.each do |window|
   ...
end
```
Now it will look like:
```
hpxml = HPXML.new(...)
hpxml_bldg = hpxml.buildings[0]
hpxml_bldg.windows.each do |window|
   ...
end
```

TODO:
- [x] Update HPXML class to support multiple Buildings. Follow [this example](https://code.ornl.gov/weatherization/wa_hpxml_translator/-/blob/master/wahpxml/hpxml.py).
- [x] Update code so that `in.xml` echoes the entire HPXML.
- [x] Update ReportSimOutput measure where code uses the HPXML object
- [x] Add simulation tests for NumberofUnits=X (compare results (annual and timeseries) to NumberofUnits=1)
- [x] Provide a warning when unit multipliers used
- [x] Add error-checking if OSM unique objects differ across unit models
- [x] Improve docs
- [x] Address FIXMEs
- [x] Add base-multiple-units.xml sample file

## Checklist


HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1501,10 +1465,9 @@ def get_col_index(col_name:)
     return col_num
   end
 
-  def create_schedule_file(col_name:,
-                           rows_to_skip: 1,
+  def create_schedule_file(model, col_name:, rows_to_skip: 1,
```
joseph-robertson, line 1468: 👍

Changelog.md
```diff
@@ -3,6 +3,14 @@
 __New Features__
 - **Breaking change**: Updates to newer proposed HPXML v4.0:
   - Replaces `PortableHeater` and `FixedHeater` with `SpaceHeater`.
+- Allows simulating whole multifamily (MF) buildings, where each dwelling unit is described by a separate `Building` element.
+  - Allows `NumberofUnits` to be used as a multiplier on dwelling unit simulation results to reduce simulation runtime.
+  - Adds a `building_id` argument to the HPXMLtoOpenStudio, BuildResidentialHPXML, and BuildResidentialScheduleFile measures for whole MF HPXML files.
```
joseph-robertson, line 8: There is no `building_id` argument for `BuildResidentialHPXML`, right?

shorowit: Oops. Good point.

shorowit (commented):
👍

PR #1472, opened 2023-08-23
Update GSHP curves

jmaguire1:
## Pull Request Description

Since our curves are based on a product that's > 20 years old, seems like it's time for a refresh of the curves. I've been taking the spec sheet from the latest single speed product in the ClimateMaster product line: https://www.climatemaster.com/download/18.274be999165850ccd5b5b73/1535543867815/lc377-climatemaster-commercial-tranquility-20-single-stage-ts-series-water-source-heat-pump-submittal-set.pdf 

And trying to use this data along with the EnergyPlus spreadsheet tools for generating appropriate curves that represent the performance of this product. The curves weren't able to be completely renormalized to 1.0 at rated conditions as the spreadsheet tool doesn't let users easily constrain the curves to this, although they are closer to 1.0 at rated conditions.

The overall impact here is a slight INCREASE in heating energy consumption (11% in Denver) and no substantial change in cooling (<1%, although in other climate zones it can be up to a 5% DECREASE in energy consumption), with some differences in impact depending on climate. It was pretty tricky to get the heating curves updated as the curve generation process is very sensitive to the rated conditions entered. The trick is to actually enter the conditions in the point with maximum heat output rather than what's truly a rated condition (which for water to air HP there's actually a few sets of rated conditions depending on the application).

## Checklist


shorowit:
@jmaguire1 If this addresses the warning, you should be able to delete [this line](https://github.com/NREL/OpenStudio-HPXML/blob/e0ac2d2963e63435ddec2d7423687d5f97ab2a90/workflow/tests/hpxml_translator_test.rb#L615) where we were exempting this warning. (The CI tests will throw an error if there's any warning that we don't exempt.)

joseph-robertson:
master -> fan/pump adjustment -> ground_temperatures  -> geothermal loop  -> gshp curves

afontani:
@prsh5175 : Please reach out about TSV changes. I will have some guidance around how to make sure the tests pass and we get the comparisons to run.

@joseph-robertson : For the comparisons approach "a) ensure samples are the same across the two branches" , we could base the PR off another branch (not develop) that has the TSV changes already in the branch. This way the comparisons would have the same samples.

jmaguire1:
Impact of these changes on our GSHP annual energy consumption/COP predictions:

<head>

<link rel=File-List
<style>

</style>
</head>

  |   | Heating Energy | Heating COP | Cooling Energy | Cooling COP
-- | -- | -- | -- | -- | --
Denver | Original | 5.33 | 3.67 | 2.74 | 4.62
Denver | Updated | 5.96 | 3.41 | 2.69 | 4.77
Baltimore | Original | 3.92 | 3.79 | 4.88 | 4.05
Baltimore | Updated | 4.25 | 3.52 | 4.70 | 4.23
Dallas | Original | 0.66 | 3.76 | 10.96 | 3.70
Dallas | Updated | 0.70 | 3.64 | 10.53 | 3.88
ALL | Nominal | - | 3.64 | - | 5.36



</body>

</html>

Overall impact across a few climate zones is that COP during heating slightly decreased (6%) and during cooling it slightly increased (-4%). However, the performance maps aren't strictly a bias in this fashion, it has more to do with where on the performance map we land:

image

image

The heating performance map is closer to a biased result, I've spent a LONG time messing with how to generate the curves (using different rated conditions in the spreadsheet, looking at expanding the dataset a few different ways based on the approach in Tang's thesis). It's worth point out that the COP curves aren't direct inputs here: there's curves for heating capacity and power, which you can use to calculate COP given rated conditions. It's also worth pointing out that since the incoming water temperature depends on the deep ground temperature more than anything, we wouldn't expect to see that we're hitting the full range I'm plotting unless we went to some rather extreme climates.

jmaguire1:
One more thing I forgot to mention: it wasn't possible to easily get rid of the warnings around curves not being centered at the nominal point. We're closer than we were, but there's still several curves that are > 5% different than the nominal at rated conditions. Since there's no way to force the curve generation to set the 1.0 value at rated conditions, I can't easily remove those warnings as much as I'd like to.

shorowit (approved):

Issue #1435, opened 2023-07-13
Allow radiant barrier for attic floor


shorowit:
Looks like the [HPXML schema already allows it](https://hpxml.nrel.gov/datadictionary/3.1.0/Building/BuildingDetails/Enclosure/Roofs/Roof/RadiantBarrierLocation), just need to start using it. Currently we ignore the field and assume the radiant barrier is on the underside of the roof.

shorowit:
Review current assumption about properties with respect to soiling. @afontani can help.

PR #1473, opened 2023-08-23
Adding flexibility to specify location of the radiant barrier

prsh5175:
## Pull Request Description

Allowing Radiant Barrier for Attic Floor

Addresses: https://github.com/NREL/OpenStudio-HPXML/issues/1435

## Checklist


afontani:
Here is the ASTM Standard Specification for Radiant Barriers: [ASTM C1313](https://compass.astm.org/document/?contentCode=ASTM%7CC1313_C1313M-13R19%7Cen-US&proxycl=https%3A%2F%2Fsecure.astm.org&fromLogin=true)

I would like to see a default value of 0.1 emissivity with a flag for soiling that increases the emissivity as over time dust will accumulate and increase the amount of radiative heat transfer.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -270,6 +270,10 @@ class HPXML < Object
   PVTrackingType1Axis = '1-axis'
   PVTrackingType1AxisBacktracked = '1-axis backtracked'
   PVTrackingType2Axis = '2-axis'
+  RadiantBarrierLocationUndersideOfRafters = 'underside of rafters'
```
afontani, line 273: How is "underside of rafters" different than "top side of truss under sheathing"?

prsh5175: Green line Vs Orange line in this image? 
image

shorowit:
It looks like we are currently assuming a radiant battier emittance of 0.05: https://github.com/NREL/OpenStudio-HPXML/blob/4fd2701e4f70734a17dd7bd3306f351d1f52e19f/HPXMLtoOpenStudio/resources/materials.rb#L302-L311

As far as I can tell, we have been using this number since at least 2010 (in BEopt).

afontani:
> It looks like we are currently assuming a radiant barrier emittance of 0.05:
> 
> https://github.com/NREL/OpenStudio-HPXML/blob/4fd2701e4f70734a17dd7bd3306f351d1f52e19f/HPXMLtoOpenStudio/resources/materials.rb#L302-L311
> 
> As far as I can tell, we have been using this number since at least 2010 (in BEopt).

ASTM C1313 3.2.1 defines a radiant barrier as <= 0.1. 
Screen Shot 2023-09-05 at 12 55 17 PM

HPXMLtoOpenStudio/resources/hpxml.rb
shorowit: (It didn't use to be an HPXML enumeration choice but a [HPXML user requested it](https://github.com/hpxmlwg/hpxml/issues/16).)

afontani:
For soiling of radiant barriers on the attic floor, ORNL published an article published that gives an anecdotal estimate of reduced performance [RADIANT BARRIER:
Effect of Radiant Barriers on Heating and Cooling Bills](https://web.ornl.gov/sci/buildings/tools/radiant/rb2/).
Screen Shot 2023-10-03 at 11 58 02 AM

There is probably a technical report somewhere with actual reductions in performance. But if an emittance required by ASTM C1313 is 0.1 and wood emittance is 0.9, then half the difference would be 0.5.

workflow/sample_files/base-appliances-coal.xml
```diff
@@ -113,7 +113,6 @@
             <SolarAbsorptance>0.7</SolarAbsorptance>
             <Emittance>0.92</Emittance>
             <Pitch>6.0</Pitch>
-            <RadiantBarrier>false</RadiantBarrier>
```
joseph-robertson, line 116: Do we want to preserve `false` here?

shorowit: I'm okay with these optional elements being omitted.

HPXMLtoOpenStudio/resources/constructions.rb
```diff
@@ -681,6 +835,10 @@ def self.apply_wood_frame_floor_ceiling(model, surfaces, constr_name, is_ceiling
           addtl_thick_in = rigid_r / 3.0 # Assume roughly R-3 per inch of loose-fill above cavity
           mat_addtl_ins = Material.new(name: 'ceiling loosefill ins', thick_in: addtl_thick_in, mat_base: BaseMaterial.InsulationGenericLoosefill, k_in: addtl_thick_in / rigid_r)
         end
+        mat_rb = nil
+        if has_radiant_barrier
+          mat_rb = Material.RadiantBarrier(radiant_barrier_grade, true)
+        end
```
shorowit, line 841: Why is this code inside the `else`? That means it will only apply if the ceiling has insulation. It should be applicable even if the ceiling is uninsulated. Comment applies to all ceiling methods.

HPXMLtoOpenStudio/resources/constructions.rb
```diff
@@ -700,6 +858,9 @@ def self.apply_wood_frame_floor_ceiling(model, surfaces, constr_name, is_ceiling
       if not mat_int_finish_or_covering.nil?
         constr.add_layer(mat_int_finish_or_covering)
       end
+      if not mat_rb.nil?
+        constr.add_layer(mat_rb)
+      end
```
shorowit, line 863: This doesn't look like the right place. The order of construction layers here is insulation > framing > interior finish > radiant barrier. That means the radiant barrier is below the drywall, whereas it should be above the insulation. Comment applies to all ceiling methods.

shorowit (changes requested):
I ran the HPXML test files and inspected the IDFs; the attic floor case is definitely not correct. I pointed out a couple places where the code looks wrong. Please diff the IDF for the attic floor case against another similar file and verify that you see the expected changes.

HPXMLtoOpenStudio/resources/constructions.rb
prsh5175: Now moved.

HPXMLtoOpenStudio/resources/constructions.rb
prsh5175: Now moved.

shorowit:
Attic temperatures in Dallas (blue=no RB, red=RB on attic floor, orange=RB on attic roof/walls):

image

Cooling energy for same cases:

image

shorowit (approved):
Looks good to me. Will double-check final results from the CI but this should be good to merge. Nice work, @prsh5175!

PR #1475, opened 2023-08-24
Update undisturbed ground temperatures

joseph-robertson:
## Pull Request Description

Use an adjustment, `y = 0.91x + 1.82` on `weather.data.AnnualAvgDrybulb`, for ground temperatures. The adjustment is based on a correlation between mean drybulb temperature (C) and ground temperature (C) from L. Xing 2014 model.

Also, in the `gshp_hxbore_ft_per_ton` method for calculating bore length, use the adjustment instead of the annual average drybulb temperature.

## Checklist


HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2678,8 +2678,9 @@ def self.gshp_hxbore_ft_per_ton(weather, hvac_cooling_ap, bore_spacing, pipe_r_v
...
 
-    nom_length_heat = (1.0 - hvac_cooling_ap.heat_rated_eirs[0]) * (r_value_bore + r_value_ground * rtf_DesignMon_Heat) / (weather.data.AnnualAvgDrybulb - (2.0 * hvac_cooling_ap.design_hw - hvac_cooling_ap.design_delta_t) / 2.0) * UnitConversions.convert(1.0, 'ton', 'Btu/hr')
-    nom_length_cool = (1.0 + hvac_cooling_ap.cool_rated_eirs[0]) * (r_value_bore + r_value_ground * rtf_DesignMon_Cool) / ((2.0 * hvac_cooling_ap.design_chw + hvac_cooling_ap.design_delta_t) / 2.0 - weather.data.AnnualAvgDrybulb) * UnitConversions.convert(1.0, 'ton', 'Btu/hr')
+    ground_temp_f = WeatherProcess.get_undisturbed_ground_temperature(weather, climate_zone_iecc)
+    nom_length_heat = (1.0 - hvac_cooling_ap.heat_rated_eirs[0]) * (r_value_bore + r_value_ground * rtf_DesignMon_Heat) / (ground_temp_f - (2.0 * hvac_cooling_ap.design_hw - hvac_cooling_ap.design_delta_t) / 2.0) * UnitConversions.convert(1.0, 'ton', 'Btu/hr')
+    nom_length_cool = (1.0 + hvac_cooling_ap.cool_rated_eirs[0]) * (r_value_bore + r_value_ground * rtf_DesignMon_Cool) / ((2.0 * hvac_cooling_ap.design_chw + hvac_cooling_ap.design_delta_t) / 2.0 - ground_temp_f) * UnitConversions.convert(1.0, 'ton', 'Btu/hr')
```
shorowit, line 2683: Both `nom_length_heat` and `nom_length_cool` use an annual average value. Would it be better to use the min and max (or perhaps winter and summer) monthly temperatures, respectively?

HPXMLtoOpenStudio/tests/test_weather.rb
```diff
@@ -211,4 +212,58 @@ def test_boulder_amy_with_leap_day
...
+      elsif iz == '7A'
+        gtf = 7.0
+      elsif iz == '7AK'
+        gtf = 5.4
+      elsif iz == '7B'
+        gtf = 6.5
+      elsif iz == '8AK'
+        gtf = 2.3
+      elsif ['1B', '1C', '2C', '5C', '6C', '7', '8'].include?(iz)
+        gtf = UnitConversions.convert(weather.data.AnnualAvgDrybulb, 'F', 'C')
```
shorowit, line 262: This could be improved. For example, CZ 8 should clearly use the 8AK value. 7 could be the average of 7A/7B unless it's also in AK, in which case it'd use 8AK. Etc.

HPXMLtoOpenStudio/resources/data/ground_temperatures.csv
```diff
@@ -0,0 +1,18 @@
+2012 IECC Climate Zone,Annual Average Undisturbed Ground Temperature
```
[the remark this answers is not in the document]
afontani: **Context of where the values came from:** The data is from L. Xing's 2014 [dissertation](https://shareok.org/bitstream/handle/11244/15208/Xing_okstate_0664D_13659.pdf?sequence=1) that created this ground temperature model in E+ (https://bigladdersoftware.com/epx/docs/8-7/engineering-reference/undisturbed-ground-temperature-model-xing.html). Appendix A provides a table of model inputs for many locations. The lookup table values by climate zone were created by tagging each location by climate zone and taking the median value of each climate zone. The result is what you suggest below a correlation but still some variability within each climate zone. 
image

HPXMLtoOpenStudio/resources/weather.rb
```diff
@@ -341,4 +341,21 @@ def self.calc_mains_temperatures(avgOAT, maxDiffMonthlyAvgOAT, latitude, year)
     end
     return mainsAvgTemp, mainsMonthlyTemps, mainsDailyTemps
   end
+
+  def self.get_undisturbed_ground_temperature(weather, climate_zone_iecc)
```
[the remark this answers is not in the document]
afontani: Using TMY3 station data, there are some differences (mainly in cooler climates) between the L. Xing data and the annual drybulb. There is some scatter, but the $r^2$ is pretty good. I would be in favor of using `weather.calc_ground_temperatures()` with an adjustment using the equation below to determine the ground temperature based on the L. Xing Model.

This way we use an established method already implemented in OS-HPXML, only make a small adjustment to the outputs of the method, and ground temperatures change for every location. 

**Drawback:** Different weather files will produce different ground temperatures for a given location (example: an AMY simulation for 2019 and 2020). This might not be bad for foundation heat transfer and shallow GSHP systems but is a little strange for vertical borefields.

drybulb-ground-temperature-correlation

HPXMLtoOpenStudio/resources/hvac_sizing.rb
afontani: @jmaguire1 : any thoughts here. I think you compared our sizing to various established L1-L4 methods.

HPXMLtoOpenStudio/tests/test_weather.rb
joseph-robertson: @afontani @shorowit To be clear, we're not going to be using the new `ground_temperatures.csv` lookup file, correct? Instead, we'll use `weather.calc_ground_temperatures` (C) and then apply a small adjustment based on `y = 0.91x + 1.82` where `x = average(weather.calc_ground_temperatures)`.

Or are we _also_ updating `@data.GroundMonthlyTemps << ...` to use the lookup value instead of `data.AnnualAvgDryBulb`?

HPXMLtoOpenStudio/resources/weather.rb
```diff
@@ -308,9 +308,11 @@ def calc_ground_temperatures
     bo = (data.MonthlyAvgDrybulbs.max - data.MonthlyAvgDrybulbs.min) * 0.5
 
     @data.GroundMonthlyTemps = []
+    adj_annual_avg_drybulb_c = 0.91 * UnitConversions.convert(data.AnnualAvgDrybulb, 'F', 'C') + 1.82
```
shorowit, line 311: If we end up sticking with this, we should add a comment explaining where this equation/regression came from.

HPXMLtoOpenStudio/tests/test_weather.rb
joseph-robertson: Done in https://github.com/NREL/OpenStudio-HPXML/pull/1475/commits/75e61c7ff5b5da2bc4f219321c8d23d184568318.

HPXMLtoOpenStudio/resources/weather.rb
afontani: @shorowit : I added a comment in [d3eb3c4](https://github.com/NREL/OpenStudio-HPXML/pull/1475/commits/d3eb3c47860e8b45532d5ea5530f9e39570a44e8)

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: Let's resolve this question before merging -- do we want to make a change to address this now or not?

afontani: I propose we use the `weather.data.GroundMonthlyTemps` array for the January and July values for `ground_temp_f` and then call this done for now. I will create an issue about how the L. Xing model could be used to update the `weather.data.GroundMonthlyTemps` array.

joseph-robertson: Denver looks like this:
```
0 | 45.59726
1 | 42.4099
2 | 42.15465
3 | 43.66721
4 | 49.73931
5 | 55.78311
6 | 61.02277
7 | 64.34197
8 | 64.62395
9 | 61.92434
10 | 56.8183
11 | 50.95989
```
So use temperatures corresponding to 0 and 6? And not min and max? (Min and max might also be easier for southern hemisphere.)



</body>

</html>

shorowit: I think January/July would be better than min/max, since the design loads will be calculated using the months with the hottest/coldest outdoor air temperatures. You want to know how much borehole depth you need _coincident with when those design loads occur_.

Make sure to handle locations in the southern hemisphere correctly too.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2678,8 +2678,9 @@ def self.gshp_hxbore_ft_per_ton(weather, hvac_cooling_ap, bore_spacing, pipe_r_v
     rtf_DesignMon_Heat = [0.25, (71.0 - weather.data.MonthlyAvgDrybulbs[0]) / @htd].max
     rtf_DesignMon_Cool = [0.25, (weather.data.MonthlyAvgDrybulbs[6] - 76.0) / @ctd].max
```
joseph-robertson, line 2679: Are the 0 and 6 here specific to northern hemisphere? Should we use the same logic here (i.e., switch the indexes based on latitude) as we will with the ground temperatures?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
jmaguire1: I'm on board with this (and I agree with Scott that Jan/July might be better than min/max, since that's consistent with a lot of our other design calculations).

@afontani: Do you remember what depth you used as an input to develop your correlation? The [Xing model](https://bigladdersoftware.com/epx/docs/8-7/engineering-reference/undisturbed-ground-temperature-model-xing.html) is a function of time and depth. I'm a little surprised how much of a seasonal swing we're seeing in the ground temperature here since we're generally talking about the temperature > 100 ft underground, I would have expected it to be a little more stable than what Joe is showing.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
joseph-robertson: I'll take that approach for now. Let me know if it's incorrect.

jmaguire1: What you're doing here is correct

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2675,12 +2675,21 @@ def self.gshp_hxbore_ft_per_ton(weather, hvac_cooling_ap, bore_spacing, pipe_r_v
     r_value_grout = 1.0 / hvac_cooling_ap.grout_conductivity / beta_0 / ((hvac_cooling_ap.bore_diameter / hvac_cooling_ap.pipe_od)**beta_1)
     r_value_bore = r_value_grout + pipe_r_value / 2.0 # Note: Convection resistance is negligible when calculated against Glhepro (Jeffrey D. Spitler, 2000)
 
-    rtf_DesignMon_Heat = [0.25, (71.0 - weather.data.MonthlyAvgDrybulbs[0]) / @htd].max
-    rtf_DesignMon_Cool = [0.25, (weather.data.MonthlyAvgDrybulbs[6] - 76.0) / @ctd].max
+    is_southern_hemisphere = (weather.header.Latitude < 0)
```
afontani, line 2678: Thank you for this addition.

afontani (approved):
Thank you for adding a unit test.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: Nice catch here, @joseph-robertson.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: Based on these changes, total bore lengths are increasing by 25-50% in our tests. That's a lot.

image

@jmaguire1's comment makes me reconsider here. Indeed our monthly ground temperatures represent shallow temperatures primarily used for foundation heat transfer modeling. If we should be using deep ground temperatures here, then using the annual average is probably (?) better than what we've changed to, though we could look at the Xing data to confirm. (If we could develop monthly ground temperatures for deeper depths, that'd be the best possible outcome of course, but that would take more work.)

HPXMLtoOpenStudio/resources/weather.rb
```diff
@@ -308,9 +308,12 @@ def calc_ground_temperatures
     bo = (data.MonthlyAvgDrybulbs.max - data.MonthlyAvgDrybulbs.min) * 0.5
 
     @data.GroundMonthlyTemps = []
+    # The regression coefficients are from L. Xing's simplified design model ground temperatures (Appendix A-3) and the nearest TMY3 station's annual dry bulb.
+    adj_annual_avg_drybulb_c = 0.91 * UnitConversions.convert(data.AnnualAvgDrybulb, 'F', 'C') + 1.82
+    adj_annual_avg_drybulb_f = UnitConversions.convert(adj_annual_avg_drybulb_c, 'C', 'F')
     for i in 0..11
       theta = amon[i] * 24.0
-      @data.GroundMonthlyTemps << UnitConversions.convert(data.AnnualAvgDrybulb - bo * Math::cos(2.0 * Math::PI / p * theta - po - phi) * gm + 460.0, 'R', 'F')
+      @data.GroundMonthlyTemps << UnitConversions.convert(adj_annual_avg_drybulb_f - bo * Math::cos(2.0 * Math::PI / p * theta - po - phi) * gm + 460.0, 'R', 'F')
```
joseph-robertson, line 316: @afontani and I plan to:

Leave this. But in `hvac_sizing.rb`, use `adj_annual_avg_drybulb_f` instead of the average of `@data.GroundMonthlyTemps`.

joseph-robertson: image

shorowit (approved):
Did some minor refactoring/code cleanup. Looks good to me. Will merge if the CI is still happy and results change as expected.

PR #1482, opened 2023-09-13
Testing: fact_sheets_category_6

joseph-robertson:
## Pull Request Description

Add new sample files that demonstrate "weird sizing behavior".

## Checklist


shorowit:
I looked over the sizing calcs and they look reasonable to me. The runs are all in Denver where the design temperature is 6.8F. All runs have an identical heating design load (23640 Btuh) at that design temperature.

- `Electricity - MSHP, SEER 16, 9.2 HSPF, Max Load.xml`: Heating capacity retention of 0.5 at 5F, so the heat pump capacity is quite a bit larger than the heating design load.
- `Electricity - MSHP, SEER 16, 9.5 HSPF, CCHP, Max Load.xml`: Heating capacity retention of 0.7 at 5F, so the heat pump capacity is quite a larger than the heating design load, but not as large as the above case.
- `Electricity - MSHP, SEER 20, 11 HSPF, CCHP, Max Load.xml`: Heating capacity retention of 0.9 at 5F. The heating capacity retention is so high that the heat pump is actually sized based on the cooling capacity.
- `Gas - MSHP`: All of these runs have a minimum compressor temperature of 25F, so that is used to size the heat pump instead of the 6.8F design temperature. Because the heating design load at 25F is lower (than the heating design load at 6.8F), and because the heating capacity retentions at 25F are higher (than the heating capacity retentions at 6.8F), the heat pump again is sized based on the cooling capacity.

shorowit:
Closing, concerns seem to be addressed. Feel free to reopen if anything further comes up.

PR #1489, opened 2023-09-21
Fixes battery resilience output to properly incorporate battery losses

### LINKED ISSUE #1488: Battery resilience calculation missing battery losses
I'm looking at the battery resilience calculation and noticed that the battery losses used in the calculation are always zero. I went back to the original implementation ([here](https://github.com/NREL/OpenStudio-HPXML/pull/1212)) and confirmed it's zero there too.

Looks like the issue is that the code [references an output variable name](https://github.com/NREL/OpenStudio-HPXML/blob/c28a3a350f143206b9060313e7079475ac91c995/ReportSimulationOutput/measure.rb#L991), but that output variable was never created.

shorowit:
## Pull Request Description

Closes #1488.

Previously, if you ran HPXML files with batteries, you would get different results depending on whether you requested hourly resilience outputs too. If you didn't request hourly resilience outputs, _zero_ battery losses were used in the calculation, whereas if you did request hourly resilience outputs, _negative_ battery losses were incorporated in the calculation.

Now you get the same results either way, and it correctly uses _positive_ battery losses.

## Checklist


shorowit:
@joseph-robertson @jmaguire1 See the PR description for how the results were previously wrong compared to now.

My assumption behind the code changes is that the `batt_loss` that is used in the critical load calculation [here](https://github.com/NREL/OpenStudio-HPXML/blob/c28a3a350f143206b9060313e7079475ac91c995/ReportSimulationOutput/measure.rb#L1009) should be a positive value (i.e., battery losses increase the critical load). Previously it was either zero or negative.

ReportSimulationOutput/measure.rb
```diff
@@ -2882,6 +2882,12 @@ def get_object_output_variables_by_key(model, object, sys_id, class_name)
 
       if object.to_ElectricLoadCenterStorageLiIonNMCBattery.is_initialized
         return { RT::Battery => ['Electric Storage Charge Fraction'] }
+
+      elsif object.to_EnergyManagementSystemOutputVariable.is_initialized
+        if object.name.to_s.include? Constants.ObjectNameBatteryLossesAdjustment(nil)
+          return { RT::Battery => [object.name.to_s] }
+        end
```
shorowit, line 2889: This is the primary bugfix. Need to ensure we always create an output variable for the battery losses -- previously the code would _happen_ to create an output variable if you requested hourly timeseries outputs, but not without the request.

ReportSimulationOutput/measure.rb
```diff
@@ -961,69 +961,69 @@ def get_outputs(runner, args)
...
-            next unless (battery_id.is_initialized && battery_id.get == battery.id)
+      vars = ['Electric Storage Charge Fraction']
+      keys = resilience.variables.select { |v| v[2] == vars[0] }.map { |v| v[1] }
+      batt_soc = get_report_variable_data_timeseries(keys, vars, 1, 0, resilience_frequency)
 
-            minimum_storage_state_of_charge_fraction = elcd.minimumStorageStateofChargeFraction
-          end
+      keys = ['EMS']
+      vars = resilience.variables.select { |v| v[1] == keys[0] }.map { |v| v[2] }
+      batt_loss = get_report_variable_data_timeseries(keys, vars, UnitConversions.convert(-1.0, 'J', 'kWh'), 0, resilience_frequency)
```
shorowit, line 979: This is the other part of the bugfix. Previously if you requested hourly output and we obtained a `batt_loss`, it would be a negative number. Now it's a positive number.

joseph-robertson (approved):
I see now that hourly `Output:Variable` "Battery1 li ion losses adjustment outvar" was only created here: https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportSimulationOutput/measure.rb#L418, e.g., when `args[:include_timeseries_end_use_consumptions] = true`.

I'll let @jmaguire1 speak to the "non-zero `batt_loss`" issue.

jmaguire1 (approved):

shorowit:
Thanks for thoroughly testing this, @jmaguire1!

PR #1500, opened 2023-10-03
Fix possible simulation error related to EPW path.

### LINKED ISSUE #1497: Sim error if EPW next to HPXML and relative path used
E.g., put foo.epw next to foo.xml in the same dir, and use "foo.epw" in the HPXML. Gives an error.

shorowit:
## Pull Request Description

Closes #1497.

## Checklist

### DIFF
--- a/HPXMLtoOpenStudio/resources/location.rb
+++ b/HPXMLtoOpenStudio/resources/location.rb
@@ -82,24 +82,26 @@ class Location
   end
 
   def self.get_epw_path(hpxml, hpxml_path)
-    epw_path = File.absolute_path(hpxml.climate_and_risk_zones.weather_station_epw_filepath)
+    epw_filepath = hpxml.climate_and_risk_zones.weather_station_epw_filepath
+    abs_epw_path = File.absolute_path(epw_filepath)
 
-    if not File.exist? epw_path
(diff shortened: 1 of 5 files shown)


HPXMLtoOpenStudio/resources/location.rb
```diff
@@ -82,24 +82,26 @@ def self.get_climate_zone_iecc(wmo)
   end
 
   def self.get_epw_path(hpxml, hpxml_path)
-    epw_path = hpxml.climate_and_risk_zones.weather_station_epw_filepath
+    epw_filepath = hpxml.climate_and_risk_zones.weather_station_epw_filepath
+    abs_epw_path = File.absolute_path(epw_filepath)
```
shorowit, line 86: This is the bugfix, the rest is cleanup. If the HPXML file is in your current working directory, then `epw_path` will remain a relative path and get written to the OSM as a relative path. This causes downstream problems since the OSM ends up in a different directory, so the weather file is no longer relative to it. So the bugfix is to make sure we always convert the `epw_path` to an absolute path for the OSM.

shorowit: Note that in most situations, the relative path was correctly ended up as an absolute path in the OSM. But in this one particular situation, it wasn't.

Issue #1170, opened 2022-09-07
HPWH Improvements


jmaguire1:
Supersedes #932 and #862

Here's the high level TODO list for HPWH modeling, in order of priority (from my perspective):

1. Code cleanup: remove any unnecessary EMS programs, any other simplifications we can make?
2. Split systems (including CO2). This is getting worked on as part of a standards funded project, but it's likely that at the end of that project we might need to do a little more validation/testing and add new HPXML extension elements to properly wrap this up.
3. 120 V HPWHs: Main issue here is lack of data, but when we get data incorporating these appliances is high priority for any electrification study.
4. Confined space impacts: Ben Larson did a ton of additional research on this, waiting to pick his brain on if we ought to change our approach to confined space modeling. This one will require an EMS program.
5. Ducting: On hold until we use AFN in OS-HPXML, we'll revisit then. 
6. Combi applications: Will be used during ICF gas HPWH work, currently no E+ limitations we're aware of.
7. Multifunction devices (single HP for water heating, space heating, and cooling): Not widely available yet, and we'd need a good amount of information about how exactly the system works before we can do much evaluation of whether E+ is up to it or not. Put on hold until either these gain more market adoption or more users request this.

PR #1502, opened 2023-10-04
Add unmet water heater loads EMS program

jmaguire1:
## Pull Request Description
Add an EMS program for calculating unmet water heating loads. Focus is on unmet showers, but we can also calculate this for each end use if we wanted to. Addresses #1170.

## Checklist


jmaguire1:
Right now, this method calculates:

- Unmet shower energy (m*cp*delta T when T_delivered < 105 F)
- Unmet shower time
- Total shower time

Unmet shower energy I feel is the most complete metric, since it accounts for how far off from the desired temperature you are. However, no one really has a sense for if say 2 kWh of unmet showers is too many, so this way you can calculate the % of shower time where loads are unmet.

We could also add metrics related to ALL hot water usage, or all fixtures. We'd want to account for the different desired delivered temperature for appliances, but it wouldn't be hard to add metrics like Unmet Energy for all fixtures.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -1317,6 +1319,8 @@ def self.add_hot_water_and_appliances(runner, model, weather, spaces)
 
     # Add combi-system EMS program with water use equipment information
     Waterheater.apply_combi_system_EMS(model, @hpxml.water_heating_systems, plantloop_map)
+
+    return plantloop_map[sys_id]
```
shorowit, line 1323: This currently just returns the last plant loop, which isn't what you want when there are multiple water heaters in a dwelling unit.

HPXMLtoOpenStudio/resources/hotwater_appliances.rb
```diff
@@ -964,6 +965,89 @@ def self.get_fixtures_gpd(eri_version, nbeds, fixtures_all_low_flow, daily_mw_fr
...
+    # Calculate mixed water fractions
+    t_mix = 105.0 # F, Temperature of mixed water at fixtures
+    avg_setpoint_temp = 0.0 # WH Setpoint: Weighted average by fraction DHW load served
+    hpxml.water_heating_systems.each do |water_heating_system|
+      wh_setpoint = water_heating_system.temperature
+      wh_setpoint = Waterheater.get_default_hot_water_temperature(eri_version) if wh_setpoint.nil? # using detailed schedules
+      avg_setpoint_temp += wh_setpoint * water_heating_system.fraction_dhw_load_served
+    end
+    daily_wh_inlet_temperatures = calc_water_heater_daily_inlet_temperatures(weather, nbeds, hot_water_distribution, fixtures_all_low_flow, hpxml.header.sim_calendar_year)
+    daily_mw_fractions = calc_mixed_water_daily_fractions(daily_wh_inlet_temperatures, avg_setpoint_temp, t_mix)
```
shorowit, line 994: This is all duplicated code from elsewhere. We should rework this code so that the new calculation & EMS program is simply created in the existing code as we process each water heater (rather than trying to do this all at the end of the measure).

shorowit (commented):
Some comments while I think of it. Addressing these comments will make it easier to pull in the master branch.

HPXMLtoOpenStudio/resources/hotwater_appliances.rb
jmaguire1: yeah, at one point I did have most of this integrated in with the fixtures, but I was having trouble figuring out to structure things. Let me take another shot at that and then if I get stuck I'll ask you or Joe for some help.

jmaguire1:
@shorowit: I just did a major refactor of this, no more duplicate code. Thanks for the suggestions. It'd be good to know if you think this is a decent way to organize things at a minimum if/when you have a chance to take a look, we're looking to use this with the 120 V HPWH relatively soon.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -2262,11 +2262,11 @@ def self.unmet_wh_loads_program(model, water_heating_systems, plantloop_map, sho
...
       unmet_wh_loads_program.addLine("If (#{shower_flow_sensor.name} > 0) && (#{wh_temp_sensor.name} < #{mixed_setpoint_sensor.name})")
-      unmet_wh_loads_program.addLine('Set ShowerSagTime=SystemTimeStep')
-      unmet_wh_loads_program.addLine("Set ShowerE=#{shower_flow_sensor.name} * #{showers_peak_flows[water_heating_system.id]} * 4141170 * (#{mixed_setpoint_sensor.name} - #{wh_temp_sensor.name})")
+      unmet_wh_loads_program.addLine('Set ShowerSagTime = SystemTimeStep')
+      unmet_wh_loads_program.addLine("Set ShowerE = #{shower_flow_sensor.name} * #{showers_peak_flows[water_heating_system.id]} * 4141170 * (#{mixed_setpoint_sensor.name} - #{wh_temp_sensor.name})")
       unmet_wh_loads_program.addLine('Else')
-      unmet_wh_loads_program.addLine('Set ShowerSagTime=0')
-      unmet_wh_loads_program.addLine('Set ShowerE=0')
+      unmet_wh_loads_program.addLine('Set ShowerSagTime = 0.0')
+      unmet_wh_loads_program.addLine('Set ShowerE = 0.0')
```
shorowit: Can we revert this change and fix the root cause? We should be able to handle the lack of spaces. It'd be nice to not have a developer hit this issue again in the future.

jmaguire1:
Thanks @joseph-robertson for helping me to get this working for multiple WHs!

I just finished some hand calculations and things look pretty great. There's some weirdness on the first timestep being counted as unmet, I might look into if there's anything I can do about that or not (I suspect not). I also want to do the hand calculations for multiple WHs and shorter timesteps to make sure that works as expected, if it does then this will be ready for review. Hopefully later this week.

joseph-robertson:
> > There's some weirdness on the first timestep being counted as unmet,
> 
> Maybe related to the warmup period? Not too long ago, I addressed a similar issue where EC_adj was off for the first timestep, see [this commit](https://github.com/NREL/OpenStudio-HPXML/pull/1724/commits/eb50a8eeaa67ffc394ad6db88eaf8ebebc1faf66#diff-9316631e12103383267af4b98720e73e228d723d7a0faa552c5cf883e1b9e8b1).

Trying a similar approach in https://github.com/NREL/OpenStudio-HPXML/pull/1502/commits/593d488fcaba9ab5e336e2f703a9835ef70f6d9f.

jmaguire1:
One thing I did realize after trying to use this in OCHRE: without having the shower max flow rate somewhere in HPXML or an unnormalized schedule, it's hard to do the equivalent comfort calculations. I could see shower max gpm being an HPXML extension or switching to an unnormalized schedule (unnormalized is probably easier), but I think I can address this in a follow up after I get at least some of the 120V HPWH capabilities in.

docs/source/workflow_outputs.rst
```diff
@@ -325,6 +325,20 @@ Annual unmet hours are listed below.
...
+Annual unmet loads are listed below.
+
+  ========================================  =====
+  Type                                      Notes
+  ========================================  =====
+  Unmet Loads: Hot Water Shower Energy (J)  TODO
+  Unmet Loads: Hot Water Shower Time (hr)   TODO
+  ========================================  =====
+
+TODO
```
shorowit, line 340: What if we combine these with the above Annual Unmet Hours like so:

- `Unmet Hours: Showers (hr)`
- `Unmet Hours: Showers (%)`

Also document that this accounts for a storage water heater running out of water and occupant discomfort. It does not account for temperature drop due to a distribution system or a tankless water heater's minimum firing rate. Using smooth schedules (rather than detailed/stochastic schedules) may under-represent these values.

tasks.rb
```diff
@@ -2084,6 +2085,12 @@ def apply_hpxml_modification_sample_files(hpxml_path, hpxml)
                                            water_heating_system_idref: nil, # Apply to all water heaters
                                            solar_fraction: 0.65)
     end
+    if ['base-dhw-multiple-undersized.xml'].include? hpxml_file
+      hpxml_bldg.water_heating_systems.each do |water_heating_system|
+        water_heating_system.heating_capacity = 5000 if water_heating_system.water_heater_type == HPXML::WaterHeaterTypeStorage
+        water_heating_system.temperature = 100.0 if [HPXML::WaterHeaterTypeHeatPump, HPXML::WaterHeaterTypeTankless].include?(water_heating_system.water_heater_type)
```
shorowit, line 2091: This is kind of weird, the HPWH/tankless systems are not actually undersized, they just have their setpoint artificially lowered to result in unmet showers. At least for storage/HPWH, can't we just set a small tank volume instead? Maybe we don't need all the sample files. We can probably remove the tankless one, for example.

(Also, hopefully at some point in the future, we'll allow optional capacity inputs for HPWHs and tankless systems.)

Changelog.md
```diff
@@ -34,6 +34,7 @@ __New Features__
 - **Breaking change**: Disaggregates "Walls" into "Above Grade Walls" and "Below Grade Walls" in results_design_load_details.csv output file.
 - Updates `openei_rates.zip` with the latest residential utility rates from the [OpenEI U.S. Utility Rate database](https://apps.openei.org/USURDB/).
 - Adds a warning if the sum of supply/return duct leakage to outside values is very high.
+- Adds metrics for unmet hot water loads.
```
shorowit, line 37: Need a better description.

tasks.rb
joseph-robertson: I tried instead reducing tank volume (e.g., to 1 gal). Water Heater Use Side Outlet Temperature doesn't appear to drop (even when using stochastic schedules).

docs/source/workflow_outputs.rst
joseph-robertson: What should we report for timeseries, just `Unmet Hours: Showers (hr)`?

Also, seems like if we're no longer reporting unmet loads in kBtu, then we'd no longer need to use shower peak flow in the EMS program? If not, this should simplify down to just comparing Use Side temperature to Mixed temperature during shower draws.

tasks.rb
shorowit: Any idea, @jmaguire1?

docs/source/workflow_outputs.rst
shorowit: Good point, yes the % output would only make sense as an annual output.

Second question is for @jmaguire1.

ReportSimulationOutput/measure.rb
```diff
@@ -467,32 +454,21 @@ def energyPlusOutputRequests(runner, user_arguments)
     end
 
     # Unmet Hours
-    @unmet_hours.each do |_key, unmet_hour|
-      result << OpenStudio::IdfObject.load("EnergyManagementSystem:OutputVariable,#{unmet_hour.ems_variable}_annual_outvar,#{unmet_hour.ems_variable},Summed,ZoneTimestep,#{unmet_hours_program.name},hr;").get
+    @unmet_hours.each do |key, unmet_hour|
+      unmet_program = unmet_hours_program
+      if key.include?(UHT::HotWaterShower)
+        unmet_program = unmet_loads_program
```
shorowit, line 460: Rename to unmet_showers_program?

docs/source/workflow_outputs.rst
jmaguire1: Just the hours in the timeseries makes a lot of sense.

If we drop the energy metric, and just track hours of unmet loads, then we could drop the flow rate. The original intent with including flow rate and calculating energy, rather than just hours, was that we might want to account for difference between a shower that's just below setpoint and one that's like 30F below setpoint, or a small amount of flow vs. multiple showers at the same time being below setpoint. I've been proposing energy as the more important metric for a few years now, but the problem (as we discussed) is no one has any context of what 1 kWh of unmet showers means.

All that being said: I'm OK with dropping energy and trying to just report unmet hours of showering, in which case I think it's fine to drop max flow rate from the EMS program. There's no real standard way to report this, and there's advantages to hours in terms of being more intuitive.

tasks.rb
jmaguire1: Definitely a little surprised, but I know we do have some code that scale the max flow rate to ensure that you never draw out more than the whole tank volume in an hour https://github.com/NREL/OpenStudio-HPXML/blob/e770f42307069ba1ed46467ddaee3866d2f70551/HPXMLtoOpenStudio/resources/waterheater.rb#L837

I think this means the water heater might be reducing the flow rate through it with stochastic schedules if you try to specify like 2 gal/min, the WH might only see lower flow rate. E+ will throw an error if you have a flow rate on the use side larger than the tank volume/hour, so we do need this code. If reducing the tank volume and heater capacity doesn't give you unmet loads, I'm not sure that there's a better way to do this without reducing the setpoint.

docs/source/workflow_outputs.rst
joseph-robertson: Sounds good @jmaguire1. I've gone ahead and pulled the peak flow stuff out of the EMS program.

tasks.rb
joseph-robertson: Hmm, I'm definitely confused here. The code you are referencing above is in the `apply_solar_thermal` method. Looks like `storage_tank.setUseSideDesignFlowRate(UnitConversions.convert(storage_volume, 'gal', 'm^3') / 60.1)` is used only for stratified tanks, not mixed. I'm seeing that use side outlet temperature doesn't drop for an HPXML file with storage water heater (i.e., mixed).

joseph-robertson: Here is 40 gal vs 1 gal:
image

Changelog.md
joseph-robertson: https://github.com/NREL/OpenStudio-HPXML/pull/1502/commits/6df3ec6b5f3e095325726e0958801cae2233f1c1

tasks.rb
shorowit: > If reducing the tank volume and heater capacity doesn't give you unmet loads, I'm not sure that there's a better way to do this without reducing the setpoint.

I am confused as well. This new output seems meaningless if it doesn't capture unmet hours associated with not having a large enough storage tank. Isn't that exactly the kind of situation we're trying to evaluate in the context of a (lower heating capacity, potentially without backup electric resistance) HPWH?

joseph-robertson: @jmaguire1 Any ideas here?

jmaguire1: So I did do some runs here looking at super small tank volumes. Here's some of the relevant results for an undersized tank (https://github.com/NREL/OpenStudio-HPXML/blob/unmet_wh_loads/workflow/sample_files/base-dhw-undersized.xml). Relevant WH info:

- 40 gal nominal volume, electric storage water heater
- 1100 Btu/hr burners
- Mixed tank model
- Stochastic draw profile

image

I think this is pretty much what we'd expect to see, that with the super small burner capacity, the tank temperature DOES drop low enough that you get unmet loads, and we see them being tracked when there's a shower draw and the outlet temp is < 105 F (~40.56 C)?

With just dropping the tank volume down, you can still have the WH meet the load for smaller draws. It basically acts like a tiny tankless WH (I think we actually do assume ~1 gal volume for tankless water heaters right now, but a huge burner capacity. See https://github.com/NREL/OpenStudio-HPXML/blob/32d0767c8ad44c548b3cfe6f5e5fc9553a420dea/HPXMLtoOpenStudio/resources/waterheater.rb#L106)

jmaguire1:
I know one of the questions here was why we're not seeing unmet loads if we just drop the tank size to 1 gal, and I think I know why.

If we have a 1 gal WH but 4.5 kW elements, we basically have a small tankless water heater. Pretty much like the under sink units you see some manufacturers make, and the reason the "very small" UEF bin exists. So something like these: https://www.rheem.com/group/rheem-tankless-electric-water-heaters-professional-classic-series-tankless-electric-rtex-3-5-rtex-06-kw/

These units can support flow rates up ~ 1 gal/min, depending on mains. Here's an example from those spec sheets (the number is element capacity in kW):
image

Even with stochastic events but hourly timesteps, you might end up with low enough flow rates that the "tankless" water heater we use for testing can keep up with the loads. So I think it's not actually unexpected that we might not see unmet loads if you just reduce the tank volume, depending on what the draw profile looks like. From our "base-dhw-undersized", here's the fixture flow rate in gal/min:

image

So a test case with 1 gal but 4.5 kW elements might actually keep up with the shower draws, and we'd get 0 unmet loads. I think it's much better for our test cases to just use small elements/burners, rather than small tanks. But to make sure this works right, Joe helped make a new test file that has a 1 gal tank and 4.5 kW elements. With hourly timesteps/events, that still produces 0 unmet loads, but with 10 minute timesteps/events you do see a tiny bit of unmet loads (one hour). So we DO now have a test case for this if we want to make sure both situations will lead to unmet loads.

jmaguire1:
@shorowit: the other big outstanding question here was splitting out SSB from fixtures, since this is sensitive to shower flow rates. I'd vote that we not do that for now, but we could consider it in the future. I'm actually curious what feedback we get about this feature, since there's no real standard way to quantify unmet WH loads, and if people agree with me that we really should just focus on showers.

jmaguire1:
@shorowit: Are there any additional changes that need to be made here? I know this didn't make release, but I think this is largely ready to go? Or are there more changes required before we look at merging this in?

docs/source/workflow_outputs.rst
```diff
@@ -319,11 +319,16 @@ Annual unmet hours are listed below.
   =========================  =====
   Type                       Notes
   =========================  =====
-  Unmet Hours: Heating (hr)  Number of hours where the heating setpoint is not maintained.
+  Unmet Hours: Heating (hr)  Number of hours where the heating setpoint is not maintained. [#]_
   Unmet Hours: Cooling (hr)  Number of hours where the cooling setpoint is not maintained.
+  Unmet Hours: Showers (hr)  Number of hours where the shower water temperature is less than 105 deg-F. [#]_
```
shorowit, line 324: I discussed with @jmaguire1; this should really say water heater tank outlet temperature, not shower water temperature. I will change this.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -2189,4 +2189,103 @@ def self.get_water_heater_solar_fraction(water_heating_system, hpxml_bldg)
...
+      end
+
+      wh_temp_sensor = Model.add_ems_sensor(
+        model,
+        name: "#{tank.name} Outlet Temperature",
+        output_var_or_meter_name: 'Water Heater Use Side Outlet Temperature',
+        key_name: tank.name
+      )
+
+      program.addLine("If (#{shower_flow_sensor.name} > 0) && (#{wh_temp_sensor.name} < #{mixed_setpoint_sensor.name})")
```
shorowit, line 2253: After discussing with @jmaguire1... Since we no longer calculate an unmet *load*, we now only care whether there is a shower for the given timestep. To prevent confusion about how the shower schedule is used (and prevent the need for error-checking if the shower schedule does not correctly jive with the overall fixtures schedule), I'm going to change the shower schedule to be a 0/1 schedule that simply represents the presence of a shower rather than the shower usage.

shorowit (commented):
This is looking good to me. I'm going to address a couple comments I'm leaving below, but it should probably be good to merge after that.

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: I'm second-guessing this. If the detailed schedule is hourly, you'd like to use a fraction for how much of the hour there were showers. Also related: https://github.com/NREL/OpenStudio-HPXML/issues/1890

shorowit:
Need to compare outputs for:
- Hourly. stochastic schedule
- Sub-hourly. stochastic schedule
- Hourly. non-stochastic schedule
- Sub-hourly. non-stochastic schedule

The first two are currently giving very different answers (in hours), I think because of https://github.com/NREL/OpenStudio-HPXML/issues/1890.

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: I also wonder if the `ShowerUnmetTime` and `ShowerTime` variables below should be multiplied by the schedule value, where the schedule value represents how much of the timestep is for showers (stochastic) or the likelihood of showers for a given timestep (non-stochastic).

docs/source/workflow_outputs.rst
```diff
@@ -319,11 +319,16 @@ Annual unmet hours are listed below.
...
+  Unmet Hours: Showers (%)   Number of unmet shower hours as a percent of total hours for showers.
   =========================  =====
 
-These numbers reflect the number of hours during the heating/cooling season when the conditioned space temperature deviates more than 0.2 deg-C (0.36 deg-F) from the heating/cooling setpoint.
+  .. [#] Unmet heating/cooling hours reflect the number of hours during the heating/cooling season when the conditioned space temperature deviates more than 0.2 deg-C (0.36 deg-F) from the heating/cooling setpoint.
+
+  .. [#] Unmet showers hours only account for a storage water heater running out of water and occupant discomfort; they do not account for temperature drop due to a distribution system or a tankless water heater's minimum firing rate.
+         Using simple hourly schedules (rather than detailed/stochastic sub-hourly schedules) may under-represent these values.
```
shorowit, line 331: Need to review this once the results are making sense.

shorowit:
Note: I pulled #1896 into here and changed the base branch.

PR #1519, opened 2023-10-18
Measure documentation

### LINKED ISSUE #1518: Automate measure documentation (readme.md.erb)
Automate the creation of documentation for each measure. Essentially, in addition to the machine-readable `measure.xml` file, there could be a human-readable `README.md` (or similar) file.

shorowit:
## Pull Request Description

Closes #1518. Adds measure README.md documentation. Automatically generated from README.md.erb templates when running `openstudio tasks.rb update_measures`.

## Checklist

- [x] ~No unexpected changes to simulation results of sample files~


ReportSimulationOutput/README.md.erb
```diff
@@ -0,0 +1,41 @@
...
+<br/>
+<% end %>
+
+<% if arguments.size == 0 %>
+<%= "This measure does not have any user arguments" %>
+<% end %>
+
+<% if outputs.size > 0 %>
+## Outputs
+Measure outputs are:
```
joseph-robertson, line 36: Clarify that some of these are optional?

ReportUtilityBills/README.md.erb
```diff
@@ -0,0 +1,41 @@
...
+<br/>
+<% end %>
+
+<% if arguments.size == 0 %>
+<%= "This measure does not have any user arguments" %>
+<% end %>
+
+<% if outputs.size > 0 %>
+## Outputs
+Measure outputs are:
```
joseph-robertson, line 36: This makes me think: should we have an `outputs` method for this measure?

ReportSimulationOutput/README.md.erb
shorowit: Good idea, will address.

ReportUtilityBills/README.md.erb
shorowit: Hmm, I looked at implementing it but I don't think we actually can. All of the outputs are [specific to utility bill scenarios](https://github.com/NREL/OpenStudio-HPXML/blob/dd8a2ce30246ac57e44f21a08113f75d60c5298f/ReportUtilityBills/measure.rb#L333-L340).

joseph-robertson: So, similarly, emissions outputs for ReportSimOutput are excluded from the `outputs` method?

Why is B true?

shorowit: They wouldn't be additional, they'd be changing. We already register outputs with the runner, so these names would change. The only thing we don't do is populate them in the outputs section of the measure. So, for example, you can today in PAT retrieve the utility bill outputs but you have to type them in, whereas if the outputs section is available, you could choose from a dropdown menu.

joseph-robertson: Hmm I'm confused then. Outputs in the `outputs` method couldn't be dynamically populated in the same way they are for the runner?

shorowit: I'll create an issue around this, but I don't want to tackle it as part of this PR.

shorowit: https://github.com/NREL/OpenStudio-HPXML/issues/1522

joseph-robertson (approved):

PR #1523, opened 2023-10-24
Use Xing model for geothermal ground temperatures

joseph-robertson:
## Pull Request Description

Introduce `resources/data/Supplement_Table-Short-Grass-US.csv`. Use lat/lng to assign soil surface temperature amplitude and phase shift of temperature amplitude (1 and 2) on the `Xing` undisturbed ground temperature object. Assign the `Xing` object to the `GHXVertical` object, thus replacing the `Kusuda` object.

## Checklist


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -271,8 +271,17 @@ def self.apply_ground_to_air_heat_pump(model, runner, weather, heat_pump,
     # Supplemental Heating Coil
     htg_supp_coil = create_supp_heating_coil(model, obj_name, heat_pump)
 
+    # Site Ground Temperature Undisturbed
+    ts_amp_1, ts_amp_2, pl_1, pl_2 = Location.get_xing_amplitudes(weather.header.Latitude, weather.header.Longitude)
+    xing = OpenStudio::Model::SiteGroundTemperatureUndisturbedXing.new(model)
+    xing.setSoilDensity(920)
```
joseph-robertson, line 277: Should we use the Xing soil density default value here (962) instead of setting the Kusuda default value?

HPXMLtoOpenStudio/resources/location.rb
```diff
@@ -81,6 +81,37 @@ def self.get_climate_zone_iecc(wmo)
...
+    xing_amplitudes = nil
+    CSV.foreach(supplement_table_short_grass) do |row|
+      v2 = Vector[row[3].to_f, row[4].to_f]
+      new_dist = (v1 - v2).magnitude
+      if new_dist < dist
+        xing_amplitudes = row[6..9].map(&:to_f)
+        dist = new_dist
+      end
+    end
+    return [ts_amp_1, ts_amp_2, pl_1, pl_2] if dist > 1 # return default values; not close enough
```
joseph-robertson, line 110: Return the default values when lat/lon does not match anything close enough in the table? What should the threshold be? For example, I think basically any non-US location would get default values.

HPXMLtoOpenStudio/tests/test_location.rb
```diff
@@ -57,6 +57,48 @@ def test_dst_disabled
...
+    assert_equal(-13, pl_2)
+
+    runner = OpenStudio::Measure::OSRunner.new(OpenStudio::WorkflowJSON.new)
+    weather = WeatherProcess.new(epw_path: File.join(weather_dir, 'ZAF_Cape.Town.688160_IWEC.epw'), runner: runner)
+
+    ts_amp_1, ts_amp_2, pl_1, pl_2 = Location.get_xing_amplitudes(13.4, 0.7, 25, 30, weather.header)
+    assert_equal(13.4, ts_amp_1)
+    assert_equal(0.7, ts_amp_2)
+    assert_equal(25, pl_1)
+    assert_equal(30, pl_2)
```
joseph-robertson, line 90: Non-US location: return default values.

HPXMLtoOpenStudio/resources/location.rb
afontani: Alaska is going to be tough. It spans more than 140W - 165W and 55N - 70N. A guess would be 3? See the map of TMY3 locations (red) in the figure attached.
Screen Shot 2023-10-26 at 12 38 57 PM

afontani (changes requested):
We need to make sure in AK that we are not assigning defaults. Since AK is big, I think we need a larger distance threshold.

HPXMLtoOpenStudio/resources/data/Supplement_Table-Short-Grass-US.csv
shorowit, line 1: Are we just doing the US to save space? We do have users in Canada and internationally. If it really is a huge file to include all international locations, we could consider removing columns that aren't strictly needed.

joseph-robertson: @afontani Is there a version of the file with more than just the US?

joseph-robertson: Sounds like there may be, but the file would be quite large. Are we OK with just providing the US version, and issuing a warning when a user simulates a non-US location?

joseph-robertson:
Replaced by https://github.com/NREL/OpenStudio-HPXML/pull/1537.

PR #1554, opened 2023-11-15
ReportUtilityBills: remove electric storage from electricity produced

joseph-robertson:
## Pull Request Description

For homes with a battery and PV, utility bill calculations based on detailed electric rates should not include electric storage. Additionally, for homes with a battery and no PV (i.e., a scheduled battery), having nonzero electricity production leads to a bug.

## Checklist


joseph-robertson:
Seems like for sure the easier (perhaps short-term) solution is to only compensate for PV production. This would:

(A) maintain consistency in our current HPXML schema,
(B) be consistent with our reporting in ReportSimulationOutput, and
(C) be much easier to code up.

Maybe we leave PVCompensation -> ProductionCompensation as a future TODO?

jmaguire1:
I think then we wouldn't be able to calculate bills for scenarios where a battery is used for energy arbitrage, which was the thought behind allowing scheduled batteries. I still don't think we should allow the combination of feed in tariff, no PV, and battery. More generally we could just error out for FIT and no PV regardless of the battery, the FIT wouldn't change anything and if someone is supplying one they probably expect it to matter.

I also think the combination of FIT + battery with no PV is incredibly unlikely to happen in practice: if you're buying a battery for energy arbitrage, that arbitrage almost certainly won't pay for itself if net metering isn't allowed. If you have the battery following PV generation, then you're using the battery specifically to avoid backfeeding (compensated at the FIT) because that's either more valuable or backfeeding isn't allowed (as in HI).

You can get into this same situation if you have a generator as well. I think if we can the best thing to do here is to error out in any case with a FIT and no PV.

joseph-robertson:
Superseded by #1555.

PR #1574, opened 2023-12-20
BuildResidentialHPXML: make geothermal loop config argument optional

joseph-robertson:
## Pull Request Description

Make `geothermal_loop_configuration` optional. Not providing a value means no geothermal loop is added.
Also, check for a GSHP when trying to add a geothermal loop. Throw a warning, and skip adding the geothermal loop, if a GSHP is not found.

## Checklist


BuildResidentialHPXML/measure.rb
```diff
@@ -1384,10 +1384,9 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     # geothermal_loop_configuration_choices << HPXML::GeothermalLoopLoopConfigurationOther
     geothermal_loop_configuration_choices << HPXML::GeothermalLoopLoopConfigurationVertical
 
-    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('geothermal_loop_configuration', geothermal_loop_configuration_choices, true)
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('geothermal_loop_configuration', geothermal_loop_configuration_choices, false)
```
joseph-robertson, line 1387: Should we just remove the "none" choice altogether? Leaving the choices array with only "vertical" as an option?

shorowit: Either way seems fine. But I just noticed that we can create a `GeothermalLoop` in the HPXML file even if the user didn't select a GSHP for their HVAC system. Shouldn't we check that?

workflow/hpxml_inputs.json
```diff
@@ -105,7 +105,6 @@
     "heat_pump_backup_type": "none",
     "heat_pump_backup_fuel": "electricity",
     "heat_pump_backup_heating_efficiency": 0,
-    "geothermal_loop_configuration": "none",
```
shorowit, line 108: 👍

shorowit (approved):
Looks good to me. Thanks!

PR #1578, opened 2023-12-22
Room by room calculations

yzhou601:
## Pull Request Description

Add room by room sizing methodology from Manual J to workflow.

TODO: 
- [x] Implement peak load fenestration option @yzhou601 
- [x] Unit tests (using new detailed output file) @yzhou601 
- [x] Allow user control over file name @shorowit 
- [x] Allow JSON output format @shorowit 
- [x] Write annual_results.csv even if `--skip-simulation` is used @shorowit 
- [x] Expand tests for Smith, Walker, Cobb, and Bell residences @yzhou601 
  - [x] Add AED curve tests @yzhou601 
- [ ] Report block/space cfms in HPXML/CSV output files @yzhou601 
  - [ ] Double-check interaction w/ HVAC installation quality airflow defects
- [x] Apportion duct loads to rooms @yzhou601 
  - [x] Do we need a zone design load calculation? @shorowit 

## Checklist

- [x] Schematron validator (`EPvalidator.xml`) has been updated @yzhou601 
- [x] Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`) @yzhou601 
- [x] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`) @yzhou601 
- [x] Documentation has been updated @shorowit 
- [x] Changelog has been updated @shorowit 


workflow/tests/ACCA_Examples/Long_Residence.xml
```diff
@@ -167,8 +167,8 @@
           </Window>
           <Window>
             <SystemIdentifier id='Window2'/>
-            <Area>8.0</Area>
```
yzhou601, line 170: Window areas didn't sum up with the same values as manual J example, I inspected each window surfaces and reconfigured a bit here. Not sure if it will break the unit test or reduce the discrepancy. Let's see the ci return.

shorowit:
@yzhou601 Can you pull https://github.com/NREL/OpenStudio-HPXML/pull/1573 into here so that you're branched off that? It will help prevent conflicts to HPXML files and hvac_sizing.rb.

yzhou601:
The more I coded for room-by-room calculations, the more I feel that we need a test file configured in workflow/sample_files with Zone and Spaces. Current available test file is manually configured and it's a bit hard to test new features e.g. overwriting sensible internal loads.

shorowit:
We have tests where we load an HPXML, make some change to it, write a temporary file, and test that. Could that work here?

I'm not against the idea of automating the creation of these files, but it will be a fair amount of effort. The BuildResidentialHPXML measure will only get you so far (it can't handle this nuanced geometry or some of the advanced inputs), so there will be a lot of code to write using the HPXML class.

yzhou601:
> We have tests where we load an HPXML, make some change to it, write a temporary file, and test that. Could that work here?
> 
> I'm not against the idea of automating the creation of these files, but it will be a fair amount of effort. The BuildResidentialHPXML measure will only get you so far (it can't handle this nuanced geometry or some of the advanced inputs), so there will be a lot of code to write using the HPXML class.

Yep temporary test files will be sufficient, where are the tests located?

shorowit:
Here's an example: https://github.com/NREL/OpenStudio-HPXML/blob/f4727069b48760962c872eb704f7bc0a9b7351b5/HPXMLtoOpenStudio/tests/test_hvac.rb#L1292. Search that file for all the references to `@tmp_hpxml_path`.

workflow/tests/ACCA_Examples/Long_Residence.xml
```diff
@@ -52,59 +52,493 @@
           <Spaces>
             <Space>
               <SystemIdentifier id='laundry'/>
-              <FloorArea>77</FloorArea>
+              <FloorArea>77.0</FloorArea>
+              <extension>
+                <DesignLoads dataSource='software'>
+                  <Heating>
+                    <Total/>
+                    <Ducts/>
```
shorowit, line 60: @yzhou601 It looks like the HPXML class is writing these values even when they are empty.

yzhou601: I haven't figured out why the hvac plant doesn't have the same issue yet, but it seems my next commit fixed it

shorowit (commented):
TODO:
- [x] Get CI tests passing
- [ ] ~Review results~ moved to [top PR description]
- [ ] ~Consider implementing peak load fenestration approach (needs discussion)~ moved to [top PR description]

shorowit:
> * Consider implementing peak load fenestration approach (needs discussion)

We will allow an optional input on the HPXML `Space` to specify AED Excursion vs Peak Load. Use this to calculate room fenestration loads. Add room heating/cooling airflow rates. Document that an HPXML `Space` can represent one room or a block of rooms.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -336,6 +336,32 @@ def test_manual_j_residences
...
     assert_in_delta(0, hpxml_bldg.hvac_plant.cdl_lat_vent, default_tol_btuh)
     assert_in_delta(1200, hpxml_bldg.hvac_plant.cdl_lat_intgains, default_tol_btuh)
+    # reduce tolerance for each space
+    default_tol_btuh_space = 50
+    hpxml_bldg.get_conditioned_zone.spaces.each do |space|
+      if space.id.include? 'dining'
+        assert_in_delta(902 + 407, space.hdl_walls, default_tol_btuh_space)
+        assert_in_delta(463, space.hdl_ceilings, default_tol_btuh_space)
+        assert_in_delta(779, space.hdl_infil, default_tol_btuh_space)
+        # assert_in_delta(153 + 69, space.cdl_sens_walls, default_tol_btuh_space) # Area correct, cooling htm in OS-HPXML: 0.51, Manual J: 1.23; Need further investigation
```
yzhou601, line 346: @shorowit It turns out that the tests are passing with 50 btuh tolerance, probably I should add a relative error instead of absolute since the values do vary from 50 to over 1000. 
Two items that I want some attention for now:
1. Cooling wall loads being reported for WallDiningRoomNorth are much smaller than in the manual J, areas correct but the cooling HTM in our calculations are much smaller.
2.  Infiltration shows some diffs, about 50 btuh. I manually checked the wall area ratios, they're in great agreement. So it's the block loads that have some diffs.

shorowit: That's great that we are so close!

1. I looked into WallDiningRoomNorth a bit. In Section A12-13, it says "Table 4B provides wall CLTD values for walls. There are two CLTD values for each group number -- one for exposed walls, regardless of the direction the wall faces and one for partitions." But Figure A12-8 has different CLTDs for north walls vs non-directional walls. And Table 4B Footnote 1 says "Use the North CLTD values for externally shaded walls". I'm tempted to say we are doing it right and their value in the Long example is imperfect. I suggest allowing a larger tolerance for this wall and adding a comment explaining why.

2. I think the difference is because they use the Default Infiltration Table method (Table 5A) to get heating/cooling infiltration values, but we don't allow an input like "average" or "tight" in HPXML, we require an air infiltration measurement value and use the Blower Door Method. I manually calibrated the air infiltration measurement value to give similar results, but it gets heating a bit high and cooling a bit low (or vice versa, I forget). So again, I think we can use a slightly higher tolerance and add a comment. This reminds me that I think we should consider adding qualitative infiltration inputs (average, tight) to OpenStudio-HPXML. We already have an implementation in HEScore, so we could move it here. I'm going to add an issue around this, maybe you can work on that next.

shorowit: Issue added: https://github.com/NREL/OpenStudio-HPXML/issues/1652

yzhou601: > That's great that we are so close!
> 
> 1. I looked into WallDiningRoomNorth a bit. In Section A12-13, it says "Table 4B provides wall CLTD values for walls. There are two CLTD values for each group number -- one for exposed walls, regardless of the direction the wall faces and one for partitions." But Figure A12-8 has different CLTDs for north walls vs non-directional walls. And Table 4B Footnote 1 says "Use the North CLTD values for externally shaded walls". I'm tempted to say we are doing it right and their value in the Long example is imperfect. I suggest allowing a larger tolerance for this wall and adding a comment explaining why.
> 2. I think the difference is because they use the Default Infiltration Table method (Table 5A) to get heating/cooling infiltration values, but we don't allow an input like "average" or "tight" in HPXML, we require an air infiltration measurement value and use the Blower Door Method. I manually calibrated the air infiltration measurement value to give similar results, but it gets heating a bit high and cooling a bit low (or vice versa, I forget). So again, I think we can use a slightly higher tolerance and add a comment. This reminds me that I think we should consider adding qualitative infiltration inputs (average, tight) to OpenStudio-HPXML. We already have an implementation in HEScore, so we could move it here. I'm going to add an issue around this, maybe you can work on that next.

Thanks Scott for the insights. For the north walls, it caused a bit headache to read through the languages in the standard, from the A12-12, it says "For Table 4B, ASHRAE CLTD values for the North exposure are adjusted.... This average is used for partitions, regardless of time of day." So the way I interpreted Table 4B is that partition wall CLTDs are derived from ASHRAE north wall values with adjustment. And the language of "Use the North CLTD values for externally shaded walls" implies that the externally shaded walls should be treated the same as partition wall in Table 4B. But the north walls, I think it makes sense to assume it to be fully shaded, and use North CLTD values (or partition wall CLTD if I read it correctly), or leave it the same as other non-directional exterior walls.
Another thing that we need to pay attention is that the Figure A12-8 group number is ASHRAE Group numbers, not the Manual J Group Numbers, and they're dark color walls. In the section A12-13 it says so and the Figure A12-12 shows the mapping. I looked up the Group B in A12-8, adjusted the color to medium (multiply 0.83), converted it to Group K in Table 4B and they're consistent. If we directly uses the A12-8 with Manual J groups then we're not doing it correctly.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -834,14 +835,15 @@ def self.process_load_walls(mj, bldg_design_loads)
           # Base Cooling Load Temperature Differences (CLTD's) for dark colored sunlit and shaded walls
           # with 95 degF outside temperature taken from MJ8 Figure A12-8 (intermediate wall groups were
           # determined using linear interpolation). Shaded walls apply to north facing and partition walls only.
-          cltd_base_sun = [38.0, 34.95, 31.9, 29.45, 27.0, 24.5, 22.0, 21.25, 20.5, 19.65, 18.8]
-          cltd_base_shade = [25.0, 22.5, 20.0, 18.45, 16.9, 15.45, 14.0, 13.55, 13.1, 12.85, 12.6]
+          cltd_base_sun = { 'G' => 38.0, 'F-G' => 34.95, 'F' => 31.9, 'E-F' => 29.45, 'E' => 27.0, 'D-E' => 24.5, 'D' => 22.0, 'C-D' => 21.25, 'C' => 20.5, 'B-C' => 19.65, 'B' => 18.8 }
```
yzhou601, line 838: This is much more clear! Good job!

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2859,162 +2861,165 @@ def self.get_wall_group(wall)
...
+      table_4a_wall_group = 'K'
 
     end
 
-    # Maximum wall group is K
-    wall_group = [wall_group, 11].min
+    # Mapping from Figure A12-12
+    ashrae_wall_group = { 'A' => 'G', 'B' => 'F-G', 'C' => 'F', 'D' => 'E-F',
+                          'E' => 'E', 'F' => 'D-E', 'G' => 'D', 'H' => 'C-D',
+                          'I' => 'C', 'J' => 'B-C', 'K' => 'B' }[table_4a_wall_group]
```
yzhou601, line 3019: Good mapping!

workflow/tests/ACCA_Examples/Walker_Residence.xml
```diff
@@ -72,65 +116,638 @@
...
+          </Wall>
+          <Wall>
+            <SystemIdentifier id='WallLivingDiningWest'/>
+            <AttachedToSpace idref='living_dining'/>
             <ExteriorAdjacentTo>outside</ExteriorAdjacentTo>
             <InteriorAdjacentTo>conditioned space</InteriorAdjacentTo>
             <WallType>
               <ConcreteMasonryUnit/>
             </WallType>
-            <Area>1165.0</Area>
```
yzhou601, line 96: This wall area was read from the J1 form which is the net area, so the new room by room Walls may yield diffs in block load reports by specifying gross areas.

workflow/tests/ACCA_Examples/Walker_Residence.xml
```diff
@@ -72,65 +116,638 @@
         </AirInfiltration>
         <Roofs>
           <Roof>
-            <SystemIdentifier id='Roof1'/>
+            <SystemIdentifier id='RoofBedroom1'/>
+            <AttachedToSpace idref='bedroom_1'/>
             <InteriorAdjacentTo>attic - vented</InteriorAdjacentTo>
-            <Area>2015.0</Area>
```
yzhou601, line 77: Roof area here I could not track where this came from, but I used the same areas from Smith since they're the same layout on the first floor

shorowit: The roof is over an unconditioned attic, and _it's the attic floor that is associated with different rooms_. So it's weird to have an attic roof surface associated with a space like a bedroom, it's not directly connected to it. I suggest going back to a single `Roof` element and removing the `AttachedToSpace`. (More generally, for any surface that is not adjacent to conditioned space, there's no point splitting it up into multiple surfaces, nor assigning it to a `Space`.)

yzhou601: Sounds reasonable, I used to require space assignments for all surfaces, but no longer a requirement after relaxing the rules. So this is allowed and makes more sense, I'll change it.

workflow/tests/ACCA_Examples/Smith_Residence.xml
```diff
@@ -61,6 +61,72 @@
...
+        <Zone>
+          <SystemIdentifier id='SmithResidenceCrawlSpace'/>
+          <ZoneType>unconditioned</ZoneType>
+          <Spaces>
+            <Space>
+              <SystemIdentifier id='crawlspace'/>
+              <FloorArea>736.0</FloorArea>
+            </Space>
+          </Spaces>
+        </Zone>
```
shorowit, line 128: Why are we bothering to add unconditioned zones/spaces? I suggest removing from any of these files.

workflow/tests/ACCA_Examples/Smith_Residence.xml
```diff
@@ -86,173 +152,762 @@
...
             <DepthBelowGrade>7.0</DepthBelowGrade>
             <Insulation>
-              <SystemIdentifier id='FoundationWall3Insulation'/>
+              <SystemIdentifier id='FoundationWallLaundryEquipmentEastInsulation'/>
+              <AssemblyEffectiveRValue>7.3</AssemblyEffectiveRValue>
+            </Insulation>
+          </FoundationWall>
+          <FoundationWall>
+            <SystemIdentifier id='FoundationWallCrawlSpaceEast'/>
+            <AttachedToSpace idref='crawlspace'/>
```
shorowit, line 593: Remove references too.

workflow/tests/ACCA_Examples/Smith_Residence.xml
yzhou601: Oh it's created when I saw some surfaces that are interior adjacent to crawlspace. I hesitated when I was doing so, I'm fine removing those.

yzhou601:
@jonwinkler @shorowit I got a bit confused here on how to proceed to apportion the duct loads, from our discussion this week, I feel that we want to calculate duct based on space-level loads, which means change the line in [calculate_load_ducts_heating](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L1221) and [calculate_load_ducts_cooling](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L1265). But another thought came to my mind that whether we should keep the methods in duct load calculation and apportion later in [apply_load_ducts](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L1309), where we have the system level duct loads, and assign them to spaces based on its sensible load fractions (sensible load of the space/sum of sensible loads of all the spaces in the zone). I have this thought because I think that it's more reasonable to call the duct calculation at system level, since it's not really each space has its own duct system. We discussed that we may need a zone level calculation for ducts, but actually hpxml requires [hvac to be only attached to one zone](https://hpxml.nrel.gov/datadictionary/3.1.0/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem/AttachedToZone). Does it mean the system level load is equivalent to zone level loads (except the case that two systems are serving the same zone)? If the same system is serving two zones, then they should be combined to one zone, if there're two systems serving the same zone, then after calculating the duct loads for these two systems, we can assign duct loads to the spaces in the zone twice (one for each system).

shorowit:
@yzhou601 I will need to give some thought to your question(s). Also making a note to myself that we'll need to update the documentation (and changelog?) for the AttachedToZone stuff.

shorowit:
We should also consider appropriate error-checking around `AttachedToZone`/`AttachedToSpace` elements to ensure we don't provide incomplete room-by-room design load calculations. Maybe it's as simple as requiring `AttachedToZone` when `AttachedToSpace` is provided? (We use the presence of `AttachedToSpace` elements to decide if we are calculating room-by-room design loads.)

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -2192,6 +2258,123 @@ def from_doc(climate_zone_iecc)
...
+
+      XMLHelper.get_elements(building, 'BuildingDetails/Zones/Zone').each do |zone|
+        self << Zone.new(@parent_object, zone)
+      end
+    end
+  end
+
+  class Zone < BaseElement
+    def initialize(hpxml_bldg, *args)
+      @spaces = Spaces.new(hpxml_bldg)
```
yzhou601, line 2277: @shorowit  I had a hard time tracking the history of this, but the Space initialization here uses hpxml_bldg as parent object, while in from_doc, it uses self (which is zone) as parent object, this looks a bit confusing to me. Is this intended?

shorowit: Can you compare to other similar classes that have been in hpxml.rb for a long time? I would use what they've been using.

yzhou601: Yep I found Ducts attribute at HVACDistribution is quite similar and I changed it to be consistent with that, where we only call from_doc instead of reinitializing.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -433,6 +438,49 @@
...
+      <sch:assert role='ERROR' test='count(h:extension/h:ManualJInputs/h:FenestrationLoadProcedure) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/ManualJInputs/FenestrationLoadProcedure</sch:assert>
+      <sch:assert role='ERROR' test='h:extension/h:ManualJInputs/h:FenestrationLoadProcedure[text()="standard" or text()="peak"] or not(h:extension/h:ManualJInputs/h:FenestrationLoadProcedure)'>Expected extension/ManualJInputs/FenestrationLoadProcedure to be 'standard' or 'peak'</sch:assert>
+    </sch:rule>
+  </sch:pattern>
+
+  <sch:pattern>
+    <sch:title>[AttachedToSpace]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Enclosure/*/*/h:AttachedToSpace'>
+      <sch:assert role='ERROR' test='count(../../../../h:Systems/h:HVAC/h:HVACPlant/*/h:AttachedToZone) = (count(../../../../h:Systems/h:HVAC/h:HVACPlant/h:HeatingSystem) + count(../../../../h:Systems/h:HVAC/h:HVACPlant/h:CoolingSystem) + count(../../../../h:Systems/h:HVAC/h:HVACPlant/h:HeatPump))'>Expected AttachedToZone for each HVAC system</sch:assert>
```
shorowit, line 480: Nice check. Though as your test shows, this currently triggers on every AttachedToSpace element. I think you could instead trigger it once for the building by doing something like this (note: I haven't tested it):
```suggestion
    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails[h:Enclosure/*/*/h:AttachedToSpace]'>
      <sch:assert role='ERROR' test='count(h:Systems/h:HVAC/h:HVACPlant/*/h:AttachedToZone) = (count(h:Systems/h:HVAC/h:HVACPlant/h:HeatingSystem) + count(../../../../h:Systems/h:HVAC/h:HVACPlant/h:CoolingSystem) + count(h:Systems/h:HVAC/h:HVACPlant/h:HeatPump))'>Expected AttachedToZone for each HVAC system</sch:assert>
```

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -414,12 +419,22 @@ def test_manual_j_residences
       ['laundry'] => [2292, 635],
       ['workshop'] => [3057, 192]
     }
-    space_load_results.each do |space_names, sens_loads|
+    total_htg_loads = space_load_results.values.map { |loads| loads[0] }.sum.to_f
+    total_clg_sens_loads = space_load_results.values.map { |loads| loads[1] }.sum.to_f
+    space_load_results.each do |space_names, loads|
+      est_space_duct_loads_htg = (loads[0].to_f / total_htg_loads * 2561.0)
```
yzhou601, line 425: I calculated an estimated weighted duct loads here and made two comparison with and without duct loads. It's hard to split the total loads without space duct values, so this is the only best solution I can think of.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -541,27 +566,36 @@ def test_manual_j_residences
...
+    space_load_results.each do |space_names, loads|
+      est_space_duct_loads_htg = (loads[0].to_f / total_htg_loads * 499.0)
+      est_space_duct_loads_clg = (loads[1].to_f / total_clg_sens_loads * 1631.0)
+      space_load_no_ducts_htg = loads[0].to_f - est_space_duct_loads_htg
+      space_load_no_ducts_clg = loads[1].to_f - est_space_duct_loads_clg
       spaces = hpxml_bldg.conditioned_spaces.select { |space| space_names.any? { |space_name| space.id.include? space_name } }
-      spaces_htg_load = spaces.map { |space| space.hdl_total }.sum
-      spaces_clg_load = spaces.map { |space| space.cdl_sens_total }.sum
-      # For some reason, sum of space loads are much greater than total block load.
-      # Allowing extra tolerance because of this.
```
yzhou601: I took a look at this example home, it had a larger diff, so I spent some time investigating the reason. I found that for infiltration, the partition wall on the east is probably considered as "exposed wall" in the wall area ratio calculations since it's adjacent to the open stairs. In this case, the space infiltration values will go down for most spaces (except for those spaces that have east walls) because a total exposed wall area increases (denominator increases), which is the correct direction.

But we defined this wall as exterior adjacent to multifamily buffer space, so there's actually no way to differentiate "exposed" vs "non exposed" in this case, the same scenario applies to garage. Maybe we can assume garage to be open and partition wall is "exposed", and maybe also allow "other non-freezing space" to be assumed open and specify this open stairwell as "other non-freezing space"?

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -541,27 +566,36 @@ def test_manual_j_residences
...
+    # eyeball observation from figure 14-3
+    block_aed = [2500, 3200, 3900, 4200, 4600, 7500, 10600, 13900, 15200, 16100, 12700, 4100]
+    hpxml_bldg.hvac_plant.cdl_sens_aed_curve.split(', ').map { |s| s.to_f }.each_with_index do |aed_curve_value, i|
+      assert_in_delta(block_aed[i], aed_curve_value, [block_aed[i] * space_tol_frac, block_tol_btuh].max)
+    end
     space_load_results = {
       ['living', 'dining'] => [2534, 7176],
       ['kitchen', 'utility', 'entry'] => [773, 2121],
-      ['bedroom_3'] => [1550, 3572],
+      # ['bedroom_3'] => [1550, 3572], # Bedroom 3 for some reason has much greater heating loads than bedroom 2, which is of similar layout, exclude it for now
```
yzhou601, line 577: I couldn't figure out why the bedroom 3 heating loads are different from bedroom 2. They're almost the same layout and the same load components, internal gains are different but shouldn't apply to heating load calculation.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
yzhou601: And another potential reason is the occupant load assignment, for which I adjusted to pass the ci with the previous tolerances. I didn't find explanations in the example for how they assign occupant loads.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
shorowit: All your comments sound reasonable. There's only so much we can do with the examples. We are mostly looking for obvious outliers that may suggest we have a bug or deficiency. I think we've exhausted what we can learn from the examples and what we can build tests around, so we should call this effort complete and move on. 👍

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
shorowit: Yeah, they have one (or two?) examples where they detail the occupant assignment, but in most cases they didn't. I took some guesses at this based on the differences between their results and ours, but it's all a bit squishy indeed.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -816,7 +816,7 @@ def self.process_load_walls(mj, bldg_design_loads, all_space_design_loads)
       end
       if not wall.attached_to_space_idref.nil?
         space_design_loads = all_space_design_loads[wall.attached_to_space_idref]
-        if wall.is_exterior
+        if wall.is_exposed
```
shorowit, line 819: @yzhou601 We are already only looping through thermal boundary surfaces, what additional surfaces are limited by using `is_exposed`? I can't help but feel that we don't need any if statement at all here. Same for foundation walls below.

yzhou601: So the is_exterior was used to be `exterior_adjacent_to` to be outside (or ground for foundation wall) before. Manual J has a concept of whether the surface contacts outdoor air, which is what this is_exposed tries to handle it better. It's not all the thermal boundary surfaces, but not only exterior walls. For example, garage is assumed to be open so by default the garage wall should be counted as an exposed wall, and in the Cobb example, east walls are connecting to an open stairwell which is also considered exposed walls. Exposed walls are summed up to calculate the wall area ratio for space infiltration load calculations.

yzhou601:
@shorowit @jonwinkler I'm a bit not sure how we are going to handle the cfm output, maybe I forgot something from our last discussion. So the cfm reporting at HPXML system level is after manual s procedure and adjusted by a few methods, but the space load report is only for design loads, are we just going to report the apportioned final airflows ? In that way is it supposed to sit under the Zone design loads PR where we have zone level results?

shorowit (approved):
I'm happy with this. Diffs are pretty minimal, and where they occur, are improvements. This PR has hung out for so long, I'm going to merge it. Thanks for all your contributions here, @yzhou601! Work on the space airflow assignments can occur in another PR.

Issue #1581, opened 2024-01-02
Mains water temperature can go below freezing


jmaguire1:
I'm pretty confident that at one point we were doing this but lost it. The mains temperature should never go below freezing, even if the algorithm we use would suggest a lower mains temperature. We should just set the minimum temperature to be freezing here, should be relatively straightforward to implement but it's not there right now.

We currently provide a schedule for the daily mains temperature like this:
![image](https://github.com/NREL/OpenStudio-HPXML/assets/8376057/ebc13ecb-3d7d-479c-bfd0-aac111fc89a4)
This example is Fairbanks, AK and clearly goes below 0.

shorowit:
Huh. I see that we added this check to BEopt's code, but that must have been after we had already ported the routine over to OpenStudio. I don't see any issues/commits/code related to this since we moved to OpenStudio. Anyway, good to have this fixed up.

jmaguire1:
Yeah, I'm very sure we had this check (I think we added it around the time we noticed that it wouldn't work for the southern hemisphere) in BEopt, and somehow it never made it's way into OS-HPXML. I was surprised to see I could get a negative mains temperature, although I've only seen that happen in Alaska which might be how this escaped notice for so long. Easy to fix at least, glad we had a reason to check this!

PR #1582, opened 2024-01-02
Fix for mains temperature going below freezing.

jmaguire1:
## Pull Request Description

Addresses #1581. Mains temperature CAN go below freezing (0 C) right now, which is clearly physically impossible. Add a check to the mains temperature algorithm to make sure this can't happen, and provide water at 0 C if it would otherwise happen.

As an example, here's the mains water temperature with and without this change for Fairbanks, AK:
image

There is [one spot](https://github.com/NREL/OpenStudio-HPXML/blob/2601c24cecd3c1ef9c71762d18cfa55c50953ac1/HPXMLtoOpenStudio/resources/hotwater_appliances.rb#L888C12-L888C54) in the code where the mains temperature can be updated after the initial creation. This only happens if you have drain water heat recovery (DWHR), and can only increase the mains temperature, never decrease it. Since it can never decrease mains temperature or get anywhere near boiling we don't need an additional change there.

## Checklist

~- [X] Schematron validator (`EPvalidator.xml`) has been updated~
~- [X] Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)~
~- [X] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)~
~- [X] Documentation has been updated~
- [X] Changelog has been updated
- [X] `openstudio tasks.rb update_measures` has been run
- [X] No unexpected changes to simulation results of sample files


HPXMLtoOpenStudio/resources/weather.rb
```diff
@@ -386,12 +386,14 @@ def calc_mains_temperatures(n_days)
     for d in 1..n_days
       data.MainsDailyTemps << data.AnnualAvgDrybulb + 6 + tmains_ratio * maxDiffMonthlyAvgOAT / 2 * Math.sin(deg_rad * (0.986 * (d - 15 - tmains_lag) + sign * 90))
     end
+    data.MainsDailyTemps.map! { |temp| [32.0, temp].max } # ensure mains never gets below freezing. Algorithm will never provide water over boiling without a check
```
shorowit, line 389: Nice use of ruby.

shorowit (approved):
Looks good to me. Thanks, @jmaguire1!

PR #1584, opened 2024-01-04
HVAC Autosizing Limits

### LINKED ISSUE #1530: Expose blower W/cfm
Expose blower W/cfm in BuildResidentialHPXML measure to allow ResStock to specify value. This is needed to account for the baseline building ductwork size for heat pump upgrades in ResStock.

### LINKED ISSUE #1556: Sizing control option to specify max allowed airflow
@joseph-robertson @yzhou601 @shorowit 

It would be useful to implement an argument for passing in the calculated maximum airflow rate for existing ducts  (`hvac_distribution_duct_max_airflow_cfm`). Analogous to system capacities, ResStock could use ApplyUpgrade to assign the values corresponding to the existing building. This could make sizing heat pump upgrades (i.e., a new advanced [HVAC sizing control input](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#hpxml-hvac-sizing-control)) more realistic where the maximum allowed airflow rate could be based on the baseline.

joseph-robertson:
## Pull Request Description

Closes #1530. Closes #1556. Supersedes #1547. 

The PR aims to allow specifying upper limits for autosized capacities.

Adds optional arguments to BuildResidentialHPXML:
* `heating_system_heating_autosizing_limit`
* `cooling_system_cooling_autosizing_limit`
* `heat_pump_heating_autosizing_limit`
* `heat_pump_cooling_autosizing_limit`
* `heat_pump_backup_heating_autosizing_limit`
* `heating_system_2_heating_autosizing_limit`
* `hvac_blower_fan_watts_per_cfm`

Update sample files:
* `base-hvac-central-ac-only-1-speed-autosize-factor.xml`
* `base-hvac-furnace-gas-room-ac-autosize-factor.xml`
* `base-hvac-air-to-air-heat-pump-1-speed-autosize-factor.xml`
* `base-hvac-air-to-air-var-speed-backup-furnace-autosize-factor.xml`
* `base-hvac-mini-split-heat-pump-ductless-autosize-factor.xml`

## Checklist


HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -3097,7 +3102,7 @@ def self.apply_hvac_sizing(runner, hpxml_bldg, weather, cfa)
                 [HPXML::HVACTypeBoiler,
                  HPXML::HVACTypeElectricResistance].include?(htg_sys.heating_system_type))
           htg_sys.heating_airflow_cfm = Float(hvac_sizing_values.Heat_Airflow.round)
-          htg_sys.heating_airflow_cfm_isdefaulted = true
+          htg_sys.heating_airflow_cfm_isdefaulted = hvac_sizing_values.Heat_Airflow_isdefaulted
```
joseph-robertson, line 3105: Now, this may or may not be defaulted. It's defaulted when not provided in the HPXML file.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -47,7 +47,11 @@ def self.calculate(runner, weather, hpxml_bldg, cfa, hvac_systems)
       apply_hvac_heat_pump_logic(hvac_sizing_values, hvac_cooling)
       apply_hvac_equipment_adjustments(runner, hvac_sizing_values, weather, hvac_heating, hvac_cooling, hvac_system)
       apply_hvac_installation_quality(hvac_sizing_values, hvac_heating, hvac_cooling)
+      # apply_hvac_maximum_airflows(hvac_sizing_values, hvac_heating, hvac_cooling)
+      # apply_hvac_airflow_adjustments(hvac_sizing_values, hvac_heating, hvac_cooling)
       apply_hvac_fixed_capacities(hvac_sizing_values, hvac_heating, hvac_cooling)
+      # apply_hvac_blower_fan_adjustment(hvac_sizing_values, hvac_heating, hvac_cooling)
+      apply_hvac_airflow_adjustments(hvac_sizing_values, hvac_heating, hvac_cooling) # <= FIXME should this be before apply_hvac_fixed_capacities?
```
joseph-robertson, line 54: In the PR that this supersedes, we had the following methods called in order:
* apply_hvac_maximum_airflows
* apply_hvac_fixed_capacities
* apply_hvac_blower_fan_adjustment

Now that there is a single flag for max airflows and blower fan adjustment, should we have just a single new method for "airflow adjustments" (that's called before or after apply_hvac_fixed_capacities?), or should we follow the same order as before?

BuildResidentialHPXML/measure.rb
```diff
@@ -5343,13 +5419,20 @@ def self.set_secondary_heating_systems(hpxml_bldg, args)
       fraction_heat_load_served = args[:heating_system_2_fraction_heat_load_served]
     end
 
+    if args[:hvac_distribution_fan_watts_per_cfm].is_initialized
+      if [HPXML::HVACTypeFurnace].include?(heating_system_type)
+        fan_watts_per_cfm = args[:hvac_distribution_fan_watts_per_cfm].get
+      end
+    end
```
joseph-robertson, line 5426: This probably shouldn't apply to the second heating system, right?

shorowit (commented):
I want to make a comment before I forget it. I'm still kind of bothered by having a high-level input (`ExistingDuctworkRestriction`) that essentially changes the entire meaning of the `HeatingAirflowCFM`/`CoolingAirflowCFM` inputs and how they are used. I think it will be very confusing to a user.

My inclination is to replace the high-level input w/ two additional heating/cooling airflow inputs (`MaxHeatingAirflowCFM`/`MaxCoolingAirflowCFM`?). The schematron validation would prevent both e.g. `HeatingAirflowCFM` and `MaxHeatingAirflowCFM` from being provided for a given HVAC system.

It's fine to continue the path you're going for proof-of-concept/testing, but we should consider something like this instead.

joseph-robertson:
How would this work with the adjustment to the blower fan efficiency then? If either `MaxHeatingAirflowCFM` or `MaxCoolingAirflowCFM` provided, make the blower fan efficiency adjustment - something like that?

shorowit:
> How would this work with the adjustment to the blower fan efficiency then? If either `MaxHeatingAirflowCFM` or `MaxCoolingAirflowCFM` provided, make the blower fan efficiency adjustment - something like that?

Yes, that's what I would think. I assume you also check whether the autosized airflow is actually limited by these inputs before making the blower fan adjustment.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
shorowit: A single method sounds good to me. I assume it should be after apply_hvac_fixed_capacities.

joseph-robertson:
> I want to make a comment before I forget it. I'm still kind of bothered by having a high-level input (`ExistingDuctworkRestriction`) that essentially changes the entire meaning of the `HeatingAirflowCFM`/`CoolingAirflowCFM` inputs and how they are used. I think it will be very confusing to a user.
> 
> My inclination is to replace the high-level input w/ two additional heating/cooling airflow inputs (`MaxHeatingAirflowCFM`/`MaxCoolingAirflowCFM`?). The schematron validation would prevent both e.g. `HeatingAirflowCFM` and `MaxHeatingAirflowCFM` from being provided for a given HVAC system.
> 
> It's fine to continue the path you're going for proof-of-concept/testing, but we should consider something like this instead.

If schematron validation was updated to prevent both, then how would you get around both being present in in.xml?

shorowit:
> If schematron validation was updated to prevent both, then how would you get around both being present in in.xml? Edit: trying out a slightly different approach in the latest commit. Omit the rule from EPvalidator (i.e., allow both to be present), but then ignore `HeatingAirflowCFM` if `MaxHeatingAirflowCFM` is present.

Good point. There's no real harm in allowing both, it's just a bit silly to imagine someone submitting an HPXML file with that. But it seems fine to allow both so that the in.xml can be re-simulated.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -134,8 +136,14 @@ def test_hvac_configurations
...
               else
-                assert_in_delta(htg_cap, [htg_load, clg_load].max, 1.0)
-                assert_in_delta(clg_cap, [htg_load, clg_load].max, 1.0)
+                if max_htg_cfm.nil? && max_clg_cfm.nil?
+                  assert_in_delta(htg_cap, [htg_load, clg_load].max, 1.0)
+                  assert_in_delta(clg_cap, [htg_load, clg_load].max, 1.0)
+                else
+                  # Check HP capacity is less than max(htg_load, clg_load)
+                  assert_operator(htg_cap, :<, [htg_load, clg_load].max, 1.0)
+                  assert_operator(clg_cap, :<, [htg_load, clg_load].max, 1.0)
```
joseph-robertson, line 145: For `base-hvac-air-to-air-heat-pump-1-speed-existing-ductwork-restriction.xml`, we've restricted the airflow. So we'd expect that capacity is less than max of heating/cooling load.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -1163,117 +1163,153 @@ def test_central_air_conditioners
...
+    hpxml_bldg.cooling_systems[0].max_cooling_airflow_cfm = 100
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.025, 100, -0.11, 0, 2405, 12.0, 40.0, 1.0)
+
+    # Test defaults - ductwork restriction autosized
+    hpxml_bldg.cooling_systems[0].cooling_capacity = nil
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.008, 100, -0.11, 0, 2405, 12.0, 40.0, 1.0)
```
joseph-robertson, line 1215: For central AC, furnace, and ASHP, we check:
* the autosized airflow with no restriction and fixed capacity
* the autosized airflow with a high restriction and fixed capacity
* the autosized airflow with a low restriction and fixed capacity
* the autosized airflow with a low restriction and autosized capacity

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -48,6 +48,8 @@ def self.calculate(runner, weather, hpxml_bldg, cfa, hvac_systems, update_hpxml:
       apply_hvac_installation_quality(mj, hvac_sizing_values, hvac_heating, hvac_cooling, frac_heat_load_served, frac_cool_load_served)
       apply_hvac_autosizing_factors(hvac_sizing_values, hvac_heating, hvac_cooling)
       apply_hvac_fixed_capacities(hvac_sizing_values, hvac_heating, hvac_cooling)
+      apply_hvac_fixed_airflows(hvac_sizing_values, hvac_heating, hvac_cooling)
```
joseph-robertson, line 51: This intentionally comes after `apply_hvac_autosizing_factors`.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -1282,83 +1318,115 @@ def test_mini_split_air_conditioners
...
+    hpxml_bldg.heating_systems[0].max_heating_airflow_cfm = 1000
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_furnace_values(default_hpxml_bldg.heating_systems[0], 0.66, 273, 0, 12345, true, 999, 1.0)
+
+    # Test defaults - ductwork restriction low
+    hpxml_bldg.heating_systems[0].max_heating_airflow_cfm = 100
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_furnace_values(default_hpxml_bldg.heating_systems[0], 0.007, 53, 0, 2405, true, 999, 1.0)
```
joseph-robertson, line 1405: Shouldn't this 53 be 100?

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -1625,12 +1706,53 @@ def test_air_source_heat_pumps
...
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_air_to_air_heat_pump_values(default_hpxml_bldg.heat_pumps[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.66, 740, 513, -0.11, 0, 12345, 23456, 9876, 34567, 14.0, 8.0, nil, nil, 40.0, 1.0, 1.0, 1.0)
+
+    # Test defaults - ductwork restriction low
+    hpxml_bldg.heat_pumps[0].max_heating_airflow_cfm = 100
+    hpxml_bldg.heat_pumps[0].max_cooling_airflow_cfm = 120
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_air_to_air_heat_pump_values(default_hpxml_bldg.heat_pumps[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.03, 120, 158, -0.11, 0, 3804, 3804, 1602, 34567, 14.0, 8.0, nil, nil, 40.0, 1.0, 1.0, 1.0)
```
joseph-robertson, line 1738: Similarly, shouldn't this 120 and 158 be 100 and 120, respectively?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -1960,6 +1962,127 @@ def self.apply_hvac_fixed_capacities(hvac_sizing_values, hvac_heating, hvac_cool
...
+      hvac_sizing_values.Cool_Airflow = max_airflow_allowed
+      hvac_sizing_values.Cool_Capacity *= hvac_sizing_values.Cool_Airflow / autosized_cooling_airflow
+      hvac_sizing_values.Cool_Capacity_Sens = hvac_sizing_values.Cool_Capacity * @hvac_cooling_shr
+      hvac_sizing_values.Cool_Airflow_isdefaulted = false
+
+      if hvac_sizing_values.Heat_Capacity > 0
+        supply_air_temp = hvac_heating.additional_properties.supply_air_temp
+
+        hvac_sizing_values.Heat_Capacity = hvac_sizing_values.Cool_Capacity
+        hvac_sizing_values.Heat_Airflow = calc_airflow_rate_manual_s(mj, hvac_sizing_values.Heat_Capacity, (supply_air_temp - mj.heat_setpoint), hvac_sizing_values.Heat_Capacity)
```
joseph-robertson, line 2020: There's something about this and the use of `calc_airflow_rate_manual_s` that leads to unexpected results. Same for below. See notes about default tests' expected values around this.

joseph-robertson:
`apply_hvac_autosizing_factors` => `apply_hvac_autosizing_factors_and_limits`
`apply_hvac_fan_power_adjustment` after `apply_hvac_finalize_airflows`

jonwinkler:
**Notes from 4/15 meeting**

@joseph-robertson, We discussed the pivoting to use a `MaxCoolingCapacityAutoSizeLimit` and a `MaxHeatingCapacityAutoSizeLimit` rather than maximum allowed airflow rates. These limits would be applied to all equipment types and likely handled in the same method as the [autosizing factors.](https://github.com/NREL/OpenStudio-HPXML/blob/355ba3a3b4165c4c69cc7108c045bea84a1cd65c/HPXMLtoOpenStudio/resources/hvac_sizing.rb#L1901) 

The maximum allowed capacity should be applied after the autosizing factors. Perhaps something like:
`max_oversize_factor = min(hvac_cooling.cooling_autosizing_factor, MaxCoolingCapacityAutoSizeLimit / hvac_sizing_values.Cool_Capacity)`
`hvac_sizing_values.Cool_Capacity *= max_oversize_factor`
`hvac_sizing_values.Cool_Airflow *= max_oversize_factor`
`hvac_sizing_values.Cool_Capacity_Sens *= max_oversize_factor`

@shorowit, Does the above look right and capture the gist? Code for heating would be similar.

For use in ResStock, an algorithm is needed to set the maximum autosize limits. For ducted heat pump upgrades, this would likely be based on assuming 400 cfm/ton using the baseline airflow rates. We should use changes in this PR [1663](https://github.com/NREL/OpenStudio-HPXML/pull/1663) for consistency.

joseph-robertson:
@jonwinkler When/how should the fan power adjustment be made? Should it happen whenever the max of the new (due to either factor or limit) heating/cooling airflow is greater than the max of the autosized heating/cooling airflow?

jonwinkler:
> @jonwinkler When/how should the fan power adjustment be made? Should it happen whenever the max of the new (due to either factor or limit) heating/cooling airflow is greater than the max of the autosized heating/cooling airflow?

@joseph-robertson I think the fan power adjustment should be made after the `apply_hvac_finalize_airflows` method is called. The fan power adjustment should be made independent of how the max capacity is set/determined. The fan power fix is independent of that and is trying to ensure the baseline and upgrade models have the same duct system curve.

joseph-robertson:
> > @jonwinkler When/how should the fan power adjustment be made? Should it happen whenever the max of the new (due to either factor or limit) heating/cooling airflow is greater than the max of the autosized heating/cooling airflow?
> 
> @joseph-robertson I think the fan power adjustment should be made after the `apply_hvac_finalize_airflows` method is called. The fan power adjustment should be made independent of how the max capacity is set/determined. The fan power fix is independent of that and is trying to ensure the baseline and upgrade models have the same duct system curve.

@jonwinkler So you're saying that for the fan power adjustment, we just need the before (autosized) and after (potentially adjusted) airflow rates. If they are equal, or if autosized is less than adjusted, then no adjustment is made to the fan power. If airflow was reduced, then a fan power adjustment is made.

jonwinkler:
> > > @jonwinkler When/how should the fan power adjustment be made? Should it happen whenever the max of the new (due to either factor or limit) heating/cooling airflow is greater than the max of the autosized heating/cooling airflow?
> > 
> > 
> > @joseph-robertson I think the fan power adjustment should be made after the `apply_hvac_finalize_airflows` method is called. The fan power adjustment should be made independent of how the max capacity is set/determined. The fan power fix is independent of that and is trying to ensure the baseline and upgrade models have the same duct system curve.
> 
> @jonwinkler So you're saying that for the fan power adjustment, we just need the before (autosized) and after (potentially adjusted) airflow rates. If they are equal, or if autosized is less than adjusted, then no adjustment is made to the fan power. If airflow was reduced, then a fan power adjustment is made.

@joseph-robertson, Sorry, but not quite. For the fan power adjustment, we need the baseline max airflow, the upgrade max air flow (after it has been potentially adjusted), and the upgrade W/cfm. The fan power adjustment should be done after the all teh sizing adjustments (meaning the airflow calculation has been finalized) and made regardless if the upgraded airflow is greater or less than the baseline airflow. I think this is being handled correctly in the current implementation.

workflow/tests/base_results/results_simulations1_energy.csv
```diff
@@ -189,7 +191,9 @@ base-hvac-air-to-air-heat-pump-var-speed-autosize-maxload.xml,40.281,40.281,40.2
...
 base-hvac-air-to-air-heat-pump-var-speed-backup-boiler.xml,53.603,53.603,38.663,38.663,14.939,0.0,0.0,0.0,0.0,0.0,5.314,0.371,0.0,0.056,2.328,0.444,9.014,0.0,0.0,4.507,0.0,0.334,0.0,0.0,0.0,0.0,2.071,0.0,0.0,0.319,0.365,1.513,1.529,0.0,2.116,8.384,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,14.939,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
-base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-autosize-factor.xml,51.479,51.479,37.045,37.045,14.434,0.0,0.0,0.0,0.0,0.0,4.724,0.033,0.0,0.052,1.986,0.099,9.014,0.0,0.0,4.507,0.0,0.334,0.0,0.0,0.0,0.0,2.071,0.0,0.0,0.319,0.365,1.513,1.529,0.0,2.116,8.384,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,14.434,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```
joseph-robertson, line 192: Expecting changes here since the sample file was corrected.

BuildResidentialHPXML/measure.rb
```diff
@@ -1623,18 +1676,35 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
     arg.setDisplayName('Heating System 2: Fraction Heat Load Served')
     arg.setDescription('The heat load served fraction of the second heating system. Ignored if this heating system serves as a backup system for a heat pump.')
     arg.setUnits('Frac')
     arg.setDefaultValue(0.25)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('hvac_distribution_fan_watts_per_cfm', false)
+    arg.setDisplayName('HVAC Distribution: Blower Fan Efficiency')
+    arg.setDescription("The blower fan efficiency at maximum fan speed. Applies only to #{HPXML::HVACTypeFurnace} heating system, #{HPXML::HVACTypeCentralAirConditioner} and #{HPXML::HVACTypeMiniSplitAirConditioner} cooling systems, and #{HPXML::HVACTypeHeatPumpAirToAir}, #{HPXML::HVACTypeHeatPumpMiniSplit}, and #{HPXML::HVACTypeHeatPumpGroundToAir} heat pumps. If not provided, the OS-HPXML default (see <a href='#{docs_base_url}#hpxml-heating-systems'>HPXML Heating Systems</a>, <a href='#{docs_base_url}#hpxml-cooling-systems'>HPXML Cooling Systems</a>, <a href='#{docs_base_url}#hpxml-heat-pumps'>HPXML Heat Pumps</a>) is used.")
```
shorowit, line 1704: For mini-splits, it will only apply to ducted systems. Also need to update the logic below to incorporate that.

Changelog.md
```diff
@@ -32,6 +32,8 @@ __New Features__
   - Updates to using G-Functions from the [G-Function Library for Modeling Vertical Bore Ground Heat Exchanger](https://gdr.openei.org/submissions/1325).
   - Updated heating/cooling performance curves to reflect newer equipment.
 - Allows optional `HeatingAutosizingFactor`, `CoolingAutosizingFactor`, `BackupHeatingAutosizingFactor` inputs to scale HVAC equipment autosizing results.
+- Allows optional `HeatingAutosizingLimit`, `CoolingAutosizingLimit`, `BackupHeatingAutosizingLimit` inputs to set capacity ceiling for HVAC equipment autosizing results.
```
shorowit, line 35:
```suggestion
- Allows optional `HeatingAutosizingFactor`, `CoolingAutosizingFactor`, `BackupHeatingAutosizingFactor` inputs to scale HVAC capacities for autosized equipment.
- Allows optional `HeatingAutosizingLimit`, `CoolingAutosizingLimit`, `BackupHeatingAutosizingLimit` inputs to set maximum HVAC capacities for autosized equipment.
```

BuildResidentialHPXML/measure.rb
```diff
@@ -1125,11 +1125,23 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     arg.setUnits('Btu/hr')
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_heating_airflow', false)
+    arg.setDisplayName('Heating System: Heating Airflow')
```
shorowit, line 1129: We only want to expose airflow rate inputs for central systems (air distribution). I think you can replace all the airflow arguments with just two arguments (`hvac_distribution_heating_airflow` and `hvac_distribution_cooling_airflow`).

You'll also need to update the schematron, documentation, etc. to remove the airflow inputs from non-central systems (like wall furnace, etc.).

BuildResidentialHPXML/measure.rb
```diff
@@ -1623,18 +1676,35 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
+    args << arg
+
     arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_2_fraction_heat_load_served', true)
     arg.setDisplayName('Heating System 2: Fraction Heat Load Served')
     arg.setDescription('The heat load served fraction of the second heating system. Ignored if this heating system serves as a backup system for a heat pump.')
     arg.setUnits('Frac')
     arg.setDefaultValue(0.25)
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('hvac_distribution_fan_watts_per_cfm', false)
```
shorowit, line 1702: Move this argument down so that it's right above the duct inputs (which are part of the HVAC distribution).

shorowit (commented):
Some initial feedback for you to address before I dive in further.

BuildResidentialHPXML/measure.rb
```diff
@@ -1125,11 +1125,23 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
     arg.setUnits('Btu/hr')
     args << arg
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('heating_system_heating_airflow', false)
+    arg.setDisplayName('Heating System: Heating Airflow')
+    arg.setDescription("The output heating airflow of the heating system. If not provided, the OS-HPXML autosized default (see <a href='#{docs_base_url}#hpxml-heating-systems'>HPXML Heating Systems</a>) is used.")
```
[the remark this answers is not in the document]
shorowit: We should probably also update the display name and argument name to include "rate".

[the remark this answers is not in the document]
shorowit: Since systems can be multi-speed or variable-speed, I think these should be specified as the "maximum" airflow rate. Is that right, @jonwinkler?

workflow/hpxml_inputs.json
```diff
@@ -1689,6 +1694,15 @@
     "heat_pump_heating_autosizing_factor": 1.4,
     "heat_pump_backup_heating_autosizing_factor": 0.9
   },
+  "sample_files/base-hvac-air-to-air-heat-pump-1-speed-autosize-limit.xml": {
```
shorowit, line 1697: Rather than adding a whole bunch of extra test files and simulations for the autosizing limits, how about just incorporating the new input into the 4 existing `base-hvac-foo-autosize-factor.xml` files? Look at what autosized capacities they are getting and then choose something a bit less to demonstrate they are limiting the capacity.

BuildResidentialHPXML/measure.rb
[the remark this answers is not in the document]
jonwinkler: @shorowit Good question. I think these airflow rates should correspond to the entered cooling and heating capacities, right? For two stage, these will be maximum. But for variable speed, are the entered capacities the nominal speed or the maximum speed?

BuildResidentialHPXML/measure.rb
joseph-robertson: I'm confused why the docs say "FanPowerWattsPerCFM defaults to 0.07 W/cfm for ductless systems and 0.18 W/cfm for ducted systems." for MSHP then?

jonwinkler: I think because MSHPs can be ductless or ducted. For ductless MSHP, the default would 0.07 but ducted MSHP it would 0.18.

BuildResidentialHPXML/measure.rb
joseph-robertson: Why only for central systems, e.g., not floor/wall furnace? These systems output HeatingAirflowCFM...

workflow/hpxml_inputs.json
joseph-robertson: We are therefore ok not checking results when specifying autosizing factors?

BuildResidentialHPXML/measure.rb
shorowit: Oh, good point... Yes, I was wrong, this does apply to ductless MSHPs too.

workflow/hpxml_inputs.json
shorowit: Why wouldn't you be able to check results? I'm not following.

BuildResidentialHPXML/measure.rb
shorowit: Though that raises the question about the name convention. It's weird to call the argument "HVAC Distribution: Blower Fan Efficiency" and allow it to apply to a ductless MSHP, which doesn't have a distribution system.

BuildResidentialHPXML/measure.rb
shorowit: You're right that we do, but it's hard for me to imagine a BuildResHPXML user ever wanting to change the airflow for a room AC of wall furnace, the input will have almost no effect. It feels like overkill to introduce so many arguments when two will do.

workflow/hpxml_inputs.json
joseph-robertson: I just mean that the autosize factor values become arbitrary then, right?

BuildResidentialHPXML/measure.rb
joseph-robertson: I think that was sort of why I (ultimately) went the route of defining airflow rates at the system level -- the arguments map to system level HPXML fields.

workflow/hpxml_inputs.json
shorowit: No...

If autosized value = 30k Btuh, and you set the autosize factor to be 1.5 and the autosize limit to be 40k Btuh, you should end up at 40k Btuh. Both the factor and limit applied.

joseph-robertson: We're not testing simulation at the 1.5x autosized value. That's all I meant.

BuildResidentialHPXML/measure.rb
shorowit: Here's my proposal: call it "HVAC: Blower Fan Efficiency". The description will indicate that it applies to split (i.e., not packaged) HVAC system types. We can do the same for airflow ("HVAC: Blower Heating Airflow Rate" and "HVAC: Blower Cooling Airflow Rate"), and it will apply to the same HVAC types. @jonwinkler, does this make sense to you?

workflow/hpxml_inputs.json
shorowit: They are almost all above 1. And if you want to change them to all be above 1, that's fine.

BuildResidentialHPXML/measure.rb
jonwinkler: @shorowit, yes, that makes sense to me. I think these inputs are most useful for ducted systems.

shorowit: (Or I guess it could be "HVAC Blower: Fan Efficiency", "HVAC Blower: Heating Airflow Rate", etc.)

docs/source/workflow_inputs.rst
```diff
@@ -1662,8 +1663,10 @@ Each central furnace is entered as a ``/HPXML/Building/BuildingDetails/Systems/H
   ``AnnualHeatingEfficiency[Units="AFUE"]/Value``         double   frac       > 0, <= 1        Yes                       Rated efficiency
   ``FractionHeatLoadServed``                              double   frac       >= 0, <= 1 [#]_  See [#]_                  Fraction of heating load served
   ``extension/FanPowerWattsPerCFM``                       double   W/cfm      >= 0             No        See [#]_        Blower fan efficiency at maximum fan speed [#]_
+  ``extension/HeatingAirflowCFM``                         double   cfm        >= 0             No        autosized
```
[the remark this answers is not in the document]
joseph-robertson: Some about this being the design airflow rate (and not the installed)?

BuildResidentialHPXML/measure.rb
```diff
@@ -5828,6 +5901,35 @@ def self.set_hvac_distribution(hpxml_bldg, args)
     end
   end
 
+  def self.set_hvac_blower(hpxml_bldg, args)
+    # Blower fan W/cfm
+    hpxml_bldg.hvac_distributions.each do |hvac_distribution|
+      distribution_system_type = hvac_distribution.distribution_system_type
+
+      hvac_distribution.hvac_systems.each do |hvac_system|
+        next unless (distribution_system_type == HPXML::HVACDistributionTypeAir) || (hvac_system.is_a?(HPXML::HeatPump) && [HPXML::HVACTypeHeatPumpMiniSplit].include?(hvac_system.heat_pump_type))
```
shorowit, line 5910: This logic won't work for a ductless MSHP since there is no distribution system. How about something like this?
```suggestion
    # Blower fan W/cfm
    hpxml_bldg.hvac_systems.each do |hvac_system|
        next unless (!hvac_system.distribution_system.nil? && hvac_system.distribution_system.distribution_system_type == HPXML::HVACDistributionTypeAir) || (hvac_system.is_a?(HPXML::HeatPump) && [HPXML::HVACTypeHeatPumpMiniSplit].include?(hvac_system.heat_pump_type))
```

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -3221,6 +3242,12 @@ def self.assign_to_hpxml_system(htg_sys, clg_sys, hvac_sizing_values)
                HPXML::HVACTypeElectricResistance].include?(htg_sys.heating_system_type))
         htg_sys.heating_airflow_cfm = Float(hvac_sizing_values.Heat_Airflow.round)
         htg_sys.heating_airflow_cfm_isdefaulted = true
+
+        # Blower fan adjustment
+        if not hvac_sizing_values.Adjusted_Fan_Watts_Per_CFM.nil?
+          htg_sys.fan_watts_per_cfm = hvac_sizing_values.Adjusted_Fan_Watts_Per_CFM.round(3)
+          htg_sys.fan_watts_per_cfm_isdefaulted = true
+        end
```
shorowit, line 3250: Was this supposed to be removed?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -3285,6 +3312,12 @@ def self.assign_to_hpxml_system(htg_sys, clg_sys, hvac_sizing_values)
       # Cooling airflow
       clg_sys.cooling_airflow_cfm = Float(hvac_sizing_values.Cool_Airflow.round)
       clg_sys.cooling_airflow_cfm_isdefaulted = true
+
+      # Blower fan adjustment
+      if not hvac_sizing_values.Adjusted_Fan_Watts_Per_CFM.nil?
+        clg_sys.fan_watts_per_cfm = hvac_sizing_values.Adjusted_Fan_Watts_Per_CFM.round(3)
+        clg_sys.fan_watts_per_cfm_isdefaulted = true
+      end
```
shorowit, line 3320: And this?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -3370,5 +3403,6 @@ def initialize
   end
   attr_accessor(:Cool_Load_Sens, :Cool_Load_Lat, :Cool_Load_Tot, :Cool_Capacity, :Cool_Capacity_Sens, :Cool_Airflow,
                 :Heat_Load, :Heat_Load_Supp, :Heat_Capacity, :Heat_Capacity_Supp, :Heat_Airflow,
-                :GSHP_Loop_flow, :GSHP_Bore_Holes, :GSHP_Bore_Depth, :GSHP_G_Functions, :GSHP_Bore_Config)
+                :GSHP_Loop_flow, :GSHP_Bore_Holes, :GSHP_Bore_Depth, :GSHP_G_Functions, :GSHP_Bore_Config,
+                :Adjusted_Fan_Watts_Per_CFM)
```
shorowit, line 3407: And this?

docs/source/workflow_inputs.rst
```diff
@@ -1732,6 +1735,7 @@ Each floor furnace is entered as a ``/HPXML/Building/BuildingDetails/Systems/HVA
   ``FractionHeatLoadServed``                                   double   frac    >= 0, <= 1 [#]_  See [#]_                  Fraction of heating load served
   ``extension/FanPowerWatts``                                  double   W       >= 0             No        0               Fan power
   ``extension/HeatingAutosizingFactor``                        double   frac    > 0              No        1.0             Heating autosizing scaling factor
+  ``extension/HeatingAutosizingLimit``                         double   Btu/hr  > 0              No                        Heating autosizing capacity limit
```
shorowit, line 1738: I know you didn't write it, but can we change all the "autosizing scaling factor" language to "autosizing capacity multiplier"? Here and everywhere below.

tasks.rb
```diff
@@ -1680,7 +1676,9 @@ def apply_hpxml_modification(hpxml_file, hpxml)
                                    location: HPXML::LocationConditionedSpace)
     end
     if ['base-hvac-air-to-air-heat-pump-var-speed-backup-furnace.xml',
-        'base-hvac-autosize-air-to-air-heat-pump-var-speed-backup-furnace.xml'].include? hpxml_file
+        'base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-airflow.xml',
+        'base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-autosize-factor.xml',
+        'base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-autosize-limit.xml'].include? hpxml_file
```
shorowit, line 1681:
```suggestion
        'base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-autosize-factor.xml'].include? hpxml_file
```

workflow/hpxml_inputs.json
```diff
@@ -2217,59 +2226,70 @@
   "sample_files/base-hvac-install-quality-air-to-air-heat-pump-1-speed.xml": {
     "parent_hpxml": "sample_files/base-hvac-air-to-air-heat-pump-1-speed.xml",
     "heat_pump_airflow_defect_ratio": -0.25,
-    "heat_pump_charge_defect_ratio": -0.25
+    "heat_pump_charge_defect_ratio": -0.25,
+    "hvac_blower_fan_watts_per_cfm": 0.365
```
shorowit, line 2230: Nice.

workflow/sample_files/base-hvac-air-to-air-heat-pump-var-speed-backup-furnace-autosize-factor.xml
```diff
@@ -324,17 +324,17 @@
               <SystemIdentifier id='HeatingSystem1'/>
               <DistributionSystem idref='HVACDistribution1'/>
               <HeatingSystemType>
-                <Boiler/>
+                <Furnace/>
```
shorowit, line 327: Nice catch.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -1178,21 +1178,37 @@ def test_central_air_conditioners
...
     hpxml_bldg.cooling_systems[0].cooling_capacity = nil
+    hpxml_bldg.cooling_systems[0].cooling_autosizing_limit = 30000
+    hpxml_bldg.cooling_systems[0].cooling_autosizing_factor = nil
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.66, nil, -0.11, -0.22, nil, 12.0, 40.0, 1.0)
+    cooling_airflow_cfm = default_hpxml_bldg.cooling_systems[0].cooling_airflow_cfm
+    cooling_capacity = default_hpxml_bldg.cooling_systems[0].cooling_capacity
+
+    # Test autosizing with factors
```
shorowit, line 1200: Hmm. I love that you added so much testing, but test_defaults.rb was really intended to check the defaulting logic for simple things, not the HVAC autosizing code. The autosizing factors are included here because they get defaulted to 1 when not provided. Likewise, it's fine to check that the autosizing limits default to nil when not provided if you want (even that's not really necessary).

But the code that checks whether the autosize factors are properly applied is [in test_hvac_sizing.rb](https://github.com/NREL/OpenStudio-HPXML/blob/b692d186bcd2e89656a318598ec8b569d05e98d4/HPXMLtoOpenStudio/tests/test_hvac_sizing.rb#L625). If it's not _too_ much effort, can you move these kinds of tests to there?

shorowit (commented):
This looks pretty close, just a few things I noticed.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -4449,19 +4586,24 @@ def _test_default_furniture_mass_values(furniture_mass, area_fraction, type)
...
+  def _test_default_central_air_conditioner_values(cooling_system, shr, compressor_type, fan_watts_per_cfm, cooling_airflow_cfm, charge_defect_ratio,
                                                    airflow_defect_ratio, cooling_capacity, cooling_efficiency_seer, crankcase_heater_watts, cooling_autosizing_factor)
     assert_equal(shr, cooling_system.cooling_shr)
     assert_equal(compressor_type, cooling_system.compressor_type)
     assert_equal(fan_watts_per_cfm, cooling_system.fan_watts_per_cfm)
+    if cooling_airflow_cfm.nil? # nil implies an autosized value
+      assert(cooling_system.cooling_airflow_cfm > 0)
+    else
+      assert_in_delta(cooling_airflow_cfm, cooling_system.cooling_airflow_cfm, 1.0)
+    end
```
joseph-robertson, line 4598: Leaving these even though we always pass nil into them; we may use them in the future.

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -1178,21 +1178,28 @@ def test_central_air_conditioners
...
     _default_hpxml, default_hpxml_bldg = _test_measure()
-    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.66, -0.11, -0.22, nil, 12.0, 40.0, 1.2)
+    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.66, nil, -0.11, -0.22, nil, 12.0, 40.0, 1.2)
+
+    # Test autosizing with limits
+    hpxml_bldg.cooling_systems[0].cooling_autosizing_limit = 5000
+    hpxml_bldg.cooling_systems[0].cooling_autosizing_factor = nil
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _default_hpxml, default_hpxml_bldg = _test_measure()
+    _test_default_central_air_conditioner_values(default_hpxml_bldg.cooling_systems[0], 0.88, HPXML::HVACCompressorTypeVariableSpeed, 0.66, nil, -0.11, -0.22, 5000, 12.0, 40.0, 1.0)
```
joseph-robertson, line 1202: Let me know if all of these should be removed. Technically they are not "defaults", but they seem pretty harmless.

HPXMLtoOpenStudio/tests/test_defaults.rb
shorowit: This seems fine.

HPXMLtoOpenStudio/tests/test_defaults.rb
shorowit: I'd prefer that we remove them. These tests are covered elsewhere and the default tests already take longer than I'd like, it'd be nice not to add 15 more.

shorowit (commented):
This looks good. Just want to see that one last set of default test changes reverted and then this can merge in.

shorowit (approved):

PR #1592, opened 2024-01-19
Low-E window improvements

### LINKED ISSUE #941: Metal frame double-pane low-e window cannot lookup UFactor and SHGC
@shorowit If a window is specified as `GlassLayers = double-pane, GlassType=low-e, FrameType=Aluminum` with other inputs being defaulted, there'll be an error: `Error: Could not lookup UFactor and SHGC for window 'Window1'.`

In the defaulting, thermal break will be defaulted to be `true` (lack documentation to describe this behavior that [it will be `true` for double-pane metal frame windows, while `false` for single-pane metal frame windows](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hpxml_defaults.rb#L637-L643)), and the gas will be defaulted to be `air` for double-pane windows. 

In this way, there's no valid mapping in the lookup table for this kind of windows because the gas cannot be air for double-pane low-e metal windows: https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#id186 .

Example window specification that will capture the error:

          <Window>
            <SystemIdentifier id='Window1'/>
            <Area>102.8571429</Area>
            <Azimuth>0</Azimuth>
            <Orientation>north</Orientation>
            <FrameType>
              <Aluminum/>
            </FrameType>
            <GlassLayers>double-pane</GlassLayers>
            <GlassType>low-e</GlassType>
            <AttachedToWall idref='Wall1'/>
          </Window>

Questions:
1. Is the window specification itself not allowed? If so, we can simply ignore this error because documentation indicates a warning for improper mapping: OpenStudio-HPXML will return an error if the combination of window properties is not in the above table.
2. Or is the defaulting supposed to be changed so that the air can be defaulted to 'gas' so that the specification can be allowed?
3. Should the documentation be updated to reflect the thermal break defaulting difference for single-pane and double-pane windows?

shorowit:
## Pull Request Description

Closes #941. Adds window and skylight `GlassType` options of "low-e, high-solar-gain" and "low-e, low-solar-gain" (new in HPXML v4). The "low-e, low-solar-gain" option replaces the "reflective" option for low-e windows with solar control in the Ufactor/SHGC lookup tables.

Also now defaults to argon instead of air for double-pane low-e windows, which allows the window specified in [this comment](https://github.com/NREL/OpenStudio-HPXML/issues/941#issue-1085386645) to run successfully.

## Checklist


HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1099,8 +1106,15 @@ def self.apply_skylights(hpxml_bldg)
       end
       if skylight.gas_fill.nil?
         if skylight.glass_layers == HPXML::WindowLayersDoublePane
-          skylight.gas_fill = HPXML::WindowGasAir
-          skylight.gas_fill_isdefaulted = true
+          if [HPXML::WindowGlassTypeLowE,
+              HPXML::WindowGlassTypeLowEHighSolarGain,
+              HPXML::WindowGlassTypeLowELowSolarGain].include? skylight.glass_type
+            skylight.gas_fill = HPXML::WindowGasArgon
```
yzhou601, line 1112: It looks like that if user specifies a `double-pane`, `low-e, low-solar-gain` with Aluminum frame, it fill will cause a break because we default gas fill here and there's no mapping. 

However I am not sure if we can do better. Double pane insulating low-e with metal frames seem to work better with gas fill default but double pane solar control low-e with metal frames works better with air fill default.

yzhou601 (commented):
The changes allow better specification for low-e windows, this is really nice. The default of air fill vs gas fill is a bit tricky in terms of the current mapping. Overall is a great improvement!

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
shorowit: That's a good point. But yeah, I think we need to just live with this limitation. The issue is really that the HEScore table doesn't include double pane solar control low-e with metal frames, but I don't want to go down the rabbit hole of trying to expand the table right now.

We publish the table in our docs and also allow users to provide their own U/SHGC, so they have ways to get around the issue should they hit it.

PR #1605, opened 2024-02-05
Fix hot water piping length for edge cases with both conditioned and unconditioned basements

jmaguire1:
## Pull Request Description

ANSI/RESNET 301-2022 includes a clarification around calculating the reference pipe length (refPipeL). For homes with both conditioned and unconditioned basements, use the length required to connect the conditioned basement (and additional 10 feet), rather than summing the extra length from an unconditioned basement (5 ft) plus the conditioned space adder of 10 ft. This will mean the pipe length is 5 ft shorter for buildings with both conditioned and unconditioned basements.

We also assume pipe length = reference pipe length by default, so this will have some influence on all runs, not just when you're doing ERI calculations (although pipe length isn't directly used in a way that would change energy consumption. What is used is the ratio of pipe length over ref pipe length, since we assume the same equation for both the ratio is always 1). Would also affect any future measures that are costed based on pipe lengths.

## Checklist

- [x] ~Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)~
- [x] ~Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)~
- [X] Documentation has been updated
- [X] Changelog has been updated
- [X] `openstudio tasks.rb update_measures` has been run


shorowit (approved):
Looks good, thanks @jmaguire1. Just pulled the latest master branch in, will merge when the CI is finished.

PR #1615, opened 2024-02-15
Support MF space types when building has no HVAC

### LINKED ISSUE #1614: `Get_space_temperature_schedule` fails when no heating and cooling
[Get_space_temperature_schedule()](https://github.com/NREL/resstock/blob/2b0a7c51a6af71ed07fee2d3e13100c9984592cc/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb#L2724) fails when a building does not have any heating or cooling system. See error log below.

This issue came up from ResStock's [Water Heater Location PR](https://github.com/NREL/resstock/pull/1201), where we are assigning the location of shared water heaters to Conditioned Mechanical Room (i.e., `"LocationOtherHeatedSpace"`). In project_testing, we are hitting a rare building that does not have heating or cooling.

However, the in.xml file of the building does have `<HeatingSetpoint>` and `<CoolingSetpoint>` prepopulated under 
`<HVACSizingControl>`, which probably just doesn't get passed through to the measure. Maybe one solution is to pass those info in even if the HVAC systems are absent.

With @jmaguire1 's help, we tested a version of the same building but with a heating system specified and the problem goes away. Here's the buildstock containing the two buildings for testing. Building 48 fails and 49 passes, using the Water Heater Location PR branch.
[buildstock_test.csv](https://github.com/NREL/OpenStudio-HPXML/files/14301344/buildstock_test.csv)

@shorowit @joseph-robertson @afontani 

Failed log:
```
[11:08:15.365020 ERROR] [openstudio.measure.OSRunner] Optional not initialized
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:2743:in `get'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:2743:in `get_space_temperature_schedule'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:2824:in `get_space_or_schedule_from_location'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:1534:in `block in add_hot_water_and_appliances'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:1533:in `each'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:1533:in `add_hot_water_and_appliances'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:458:in `create_unit_model'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:182:in `block in run'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:174:in `each'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:174:in `each_with_index'
/Users/lliu2/Documents/GitHub/resstock4/resources/hpxml-measures/HPXMLtoOpenStudio/measure.rb:174:in `run'
```

joseph-robertson:
## Pull Request Description

Closes #1614.
Add `base-bldgtype-mf-unit-adjacent-to-multiple-hvac-none.xml`.

## Checklist

- [x] ~Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)~


HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1746,6 +1746,10 @@ def self.apply_detailed_performance_data_for_var_speed_systems(hpxml_bldg)
   end
 
   def self.apply_hvac_control(hpxml_bldg, schedules_file, eri_version)
+    if hpxml_bldg.hvac_controls.size == 0 && hpxml_bldg.has_location(HPXML::LocationOtherHeatedSpace)
+      hpxml_bldg.hvac_controls.add(id: "HVACControl#{hpxml_bldg.hvac_controls.size + 1}")
+    end
```
shorowit, line 1751: Note that any solution will have to address all the MF space types whose temperatures are based on the conditioned zone's temperatures, not just "other heated space".

That said, I don't think this is the right solution. Maybe it causes the simulation to run, but you end up with an "other heated space" that is the average of the conditioned zone temperature and outdoor zone -- except the conditioned zone is not conditioned. So it does not reasonably represent "other **heated** space" in any way.

My immediate reaction is that these space types should simply use default setpoints in the [EMS program](https://github.com/NREL/OpenStudio-HPXML/blob/0200043b5a9ceeb266837fed4e0fd56e2cdb60ad/HPXMLtoOpenStudio/measure.rb#L2786) (similar to what we do for [natural ventilation](https://github.com/NREL/OpenStudio-HPXML/blob/0200043b5a9ceeb266837fed4e0fd56e2cdb60ad/HPXMLtoOpenStudio/resources/airflow.rb#L428-L430)) rather than use a conditioned zone temperature sensor.

jmaguire1: I like Scott's instinct here. Using the setpoint rather than the space temp (which probably won't be that far off from outside if it's truly unconditioned, maybe a little higher from internal + solar gains) doesn't do a good job of representing a heated space.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2745,12 +2745,29 @@ def get_space_temperature_schedule(model, location, spaces)
...
+        else
+          fail 'Unexpected heating setpoints.'
+        end
+        if clg_weekday_setpoints.split(', ').uniq.size == 1 && clg_weekend_setpoints.split(', ').uniq.size == 1 && clg_weekday_setpoints.split(', ').uniq == clg_weekend_setpoints.split(', ').uniq
+          default_clg_sp = clg_weekend_setpoints.split(', ').uniq[0].to_f
+        else
+          fail 'Unexpected cooling setpoints.'
+        end
+        space_values[:temp_min] = (default_htg_sp + default_clg_sp) / 2.0 # avg-F
+      end
```
shorowit, line 2770: It doesn't make sense to use the htg setpoint in the `if` and the average of htg/clg setpoints in the `else`. Don't blindly copy what was being done for natural ventilation, the code needs to be adjusted for the situation.

That said, I don't really remember why the original implementation is like this. We assign the minimum temperature of [68F, per our documentation](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#hpxml-locations), in `Geometry.get_temperature_scheduled_space_values`, but then we override it w/ the heating setpoint. So the code doesn't match the documentation -- either the documentation needs to be updated or we shouldn't be overriding it. It looks like we started overriding it in [this PR](https://github.com/NREL/OpenStudio-HPXML/pull/400), but I don't remember why. @yzhou601, do you remember what issue that PR is trying to address?

Finally, this still doesn't address the primary issue, which is that the zone temperature is being used in the calculation below. Look at `sensor_ia` below.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2760,9 +2771,14 @@ def get_space_temperature_schedule(model, location, spaces)
 
     # Sensors
     if space_values[:indoor_weight] > 0
-      sensor_ia = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Zone Air Temperature')
-      sensor_ia.setName('cond_zone_temp')
-      sensor_ia.setKeyName(spaces[HPXML::LocationConditionedSpace].thermalZone.get.name.to_s)
+      if location == HPXML::LocationOtherHeatedSpace && !spaces[HPXML::LocationConditionedSpace].thermalZone.get.thermostatSetpointDualSetpoint.is_initialized
```
shorowit, line 2774: I'm thinking this should just be:
```suggestion
      if not spaces[HPXML::LocationConditionedSpace].thermalZone.get.thermostatSetpointDualSetpoint.is_initialized
```

so that it is used for any MF space type whose temperature is calculated based on the conditioned space temperature.

workflow/hpxml_inputs.json
```diff
@@ -842,6 +842,11 @@
   "sample_files/base-bldgtype-mf-unit-adjacent-to-other-heated-space.xml": {
     "parent_hpxml": "sample_files/base-bldgtype-mf-unit.xml"
   },
+  "sample_files/base-bldgtype-mf-unit-adjacent-to-other-heated-space-hvac-none.xml": {
+    "parent_hpxml": "sample_files/base-bldgtype-mf-unit-adjacent-to-other-heated-space.xml",
```
shorowit, line 846: We should test all the MF space types:
```suggestion
  "sample_files/base-bldgtype-mf-unit-adjacent-to-multiple-hvac-none.xml": {
    "parent_hpxml": "sample_files/base-bldgtype-mf-unit-adjacent-to-multiple.xml",
```

HPXMLtoOpenStudio/measure.rb
```diff
@@ -2760,9 +2771,14 @@ def get_space_temperature_schedule(model, location, spaces)
...
+        sensor_ia = UnitConversions.convert(space_values[:temp_min], 'F', 'C')
```
shorowit, line 2775: This is the indoor air temperature, so here we should use the average of heating and cooling setpoints.

joseph-robertson:
@shorowit Can you think of other updates/tests I should make here?

shorowit:
I don't think it needs more tests, it's such an edge case. Have you run your test file and made sure the hourly temperatures look right?

Please also update the docs to says "min of heating setpoint" instead of "min of 68F".

PR #1639, opened 2024-03-04
Geb HVAC - Onoff thermostat and realistic staging

yzhou601:
## Pull Request Description

Single Speed System features:

1.  On-off Deadband thermostat
2. Start-up Loss

Two Speed System features:

1. On-off Deadband thermostat
2. Realistic Staging

Integrated backup coil features:

1. Capacity increment and realistic staging (similar to two speed systems)

## Checklist


yzhou601:
@shorowit @jmaguire1 
Remaining question: How to structure the user inputs?
Current approach: adding a numerical input of `extension/BackupHeatingCapacityIncrement` under `HeatPump` to let user to specify how many watts for each stage, the last stage will be the remaining capacity after applying increments.  If not specified, apply the previous approach without stages
Pros: Flexible for any watts to be modeled
Cons: E+ limit of modeling at most 4 stages, which introduced constraints on the combination of increments and capacity.

Proposed approach 1: Use a boolean input instead, and default 5w as increment since that's the smallest capacity in the market.
Pros: Reasonable default number avoids error prone inputs
Cons: Less flexible

Proposed approach x? Any thoughts?

shorowit:
I don't see a problem with the current approach. If you're concerned about bad user inputs, you could add a schematron warning for a value that seems unreasonable.

shorowit:
Be sure to build in a small tolerance into the calculation of the last stage, so that the user doesn't get an additional stage with a tiny capacity simply due to e.g. rounding. For example, say they want 10kW increments (which they might round to 34121 Btuh) and a total backup capacity of 20kW (which they round to 68243 Btuh). You wouldn't want a third stage with 1 Btuh!

yzhou601:
Action items:
- [ ] Resolve the backup actuator conflicts across different features
- [ ] Add test files with multiple geb features

BuildResidentialHPXML/measure.rb
```diff
@@ -124,6 +124,18 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('simulation_control_geb_onoff_thermostat_deadband', false)
+    arg.setDisplayName('Simulation Control: HVAC GEB On-Off Thermostat Deadband')
+    arg.setDescription('On-off thermostat deadband for hvac systems. Only applies to single speed or two speed air source HVAC systems.')
+    arg.setUnits('deg-F')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('simulation_control_geb_backup_heating_capacity_increment', false)
+    arg.setDisplayName('Simulation Control: HVAC GEB Backup Heating Capacity Increment')
+    arg.setDescription("Capacity increment of the multi-staging heat pump backup system. Only applies if Backup Type is '#{HPXML::HeatPumpBackupTypeIntegrated}' and Backup Fuel Type is '#{HPXML::FuelTypeElectricity}'.")
```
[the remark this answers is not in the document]
shorowit: Also, we should mention in the descriptions that these are advanced research features. And say that they require 1 minute timesteps? (Same with the defrost argument above?)

Changelog.md
```diff
@@ -75,6 +76,8 @@ __New Features__
 - Advanced research features:
   - **Breaking change**: Replaces `SimulationControl/TemperatureCapacitanceMultiplier` with `SimulationControl/AdvancedResearchFeatures/TemperatureCapacitanceMultiplier`.
   - Allows an optional boolean input `SimulationControl/AdvancedResearchFeatures/DefrostModelType` for heat pump advanced defrost model.
+  - Allow modeling on/off thermostat deadband with start-up degradation for single speed central ac and single speed ashp systems. Currently only supports homes with at most one cooling system serving 100% cooling loads, and one heating system serving 100% heating loads (including heat pumps). An on/off thermostat deadband temperature is a temperature difference between cut-in and cut-out temperature for HVAC operations, applied to both heating and cooling.
+  - Allow modeling time-based realistic staging for two speed DX systems (central ac and ASHP). Assumes 5 minutes before transitioning to a higher speed.
```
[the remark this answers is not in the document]
shorowit: The PR description says the on/off thermostat can apply to two-stage systems too, but the changelog says single speed only. Which is it?

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -136,7 +136,11 @@ def self.add_zones_spaces_if_needed(hpxml, hpxml_bldg, cfa)
 
   def self.apply_header(hpxml_header, epw_file, hpxml_bldg)
     if hpxml_header.timestep.nil?
-      hpxml_header.timestep = 60
+      if (not hpxml_header.geb_onoff_thermostat_deadband.nil?) || (not hpxml_header.geb_backup_heating_capacity_increment.nil?)
+        hpxml_header.timestep = 1
```
shorowit, line 140: I'm a bit on the fence as to whether we should do this. If we do, it needs to be documented. (Are there other cases, like the advanced defrost model, where we should also default to 1?)

But I lean towards not doing this. Rather, we should make someone specify a timestep of 1 so they know what they're getting themselves into when they choose one of these advanced research features.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -1942,7 +1946,7 @@ def self.apply_detailed_performance_data_for_var_speed_systems(hpxml_bldg)
     end
   end
 
-  def self.apply_hvac_control(hpxml_bldg, schedules_file, eri_version)
+  def self.apply_hvac_control(hpxml_bldg, schedules_file, eri_version, _header)
```
shorowit, line 1949: Revert.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -54,9 +54,27 @@
...
+      <sch:assert role='ERROR' test='count(h:OnOffThermostatDeadbandTemperature) &lt;= 1'>Expected 0 or 1 element(s) for xpath: OnOffThermostatDeadbandTemperature</sch:assert>
+      <sch:assert role='ERROR' test='number(h:OnOffThermostatDeadbandTemperature) &gt; 0 or not(h:OnOffThermostatDeadbandTemperature)'>Expected OnOffThermostatDeadbandTemperature to be greater than 0</sch:assert>
+      <sch:assert role='ERROR' test='count(h:BackupHeatingCapacityIncrement) &lt;= 1'>Expected 0 or 1 element(s) for xpath: BackupHeatingCapacityIncrement</sch:assert>
+      <sch:assert role='ERROR' test='number(h:BackupHeatingCapacityIncrement) &gt; 0 or not (h:BackupHeatingCapacityIncrement)'>BackupHeatingCapacityIncrement should be greater than 0.0</sch:assert>
+      <sch:assert role='ERROR' test='number(../h:Timestep) = 1 or not(../h:Timestep) or not(h:BackupHeatingCapacityIncrement)'>Expected ../Timestep to be 1.0</sch:assert>
+      <sch:assert role='ERROR' test='number(../h:Timestep) = 1 or not(../h:Timestep) or not(h:OnOffThermostatDeadbandTemperature)'>Expected ../Timestep to be 1.0</sch:assert>
```
shorowit, line 62: Advanced defrost model too?

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -54,9 +54,27 @@
...
+      <sch:assert role='ERROR' test='sum(../../../../h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:FractionHeatLoadServed) &gt;= 0.99 or count(../../../../h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:FractionHeatLoadServed) = 0'>Expected sum(FractionHeatLoadServed) to be equal to 1</sch:assert>
+      <sch:assert role='ERROR' test='sum(../../../../h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:FractionCoolLoadServed) &gt;= 0.99 or count(../../../../h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:FractionCoolLoadServed) = 0'>Expected sum(FractionCoolLoadServed) to be equal to 1</sch:assert>
```
shorowit, line 73: A lot going on here...
- Checking across all the `Building` elements in the HPXML file is kind of weird and doesn't really do the right thing. If one `Building` has two cooling systems and another `Building` has zero cooling systems, the error-check will pass. It will make more sense to do these checks within each `Building`. Let's discuss.
- Do we really have to throw an error if there's any non-DX heating/cooling system? I don't really understand why it can't apply to non-DX systems, but if indeed it can't, why do we have to throw an error? Can't we just ignore it for those system types (with a warning if it makes sense)?
- The last two checks are supposed to check that the fractions are equal to 1, but all they check is that they are greater than 0.99. There's no upper limit check. Maybe you did this because of when there are multiple `Building` elements, but we shouldn't be skipping important error-checking just to get CI tests to pass. Seems like it could be easily handled. Again, let's discuss.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -304,6 +322,7 @@
       <sch:assert role='ERROR' test='count(h:extension/h:HVACSizingControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/HVACSizingControl</sch:assert> <!-- See [HVACSizingControl] -->
       <sch:assert role='ERROR' test='count(h:extension/h:ShadingControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/ShadingControl</sch:assert> <!-- See [ShadingControl] -->
       <sch:assert role='ERROR' test='count(h:extension/h:NaturalVentilationAvailabilityDaysperWeek) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/NaturalVentilationAvailabilityDaysperWeek</sch:assert>
+      <sch:assert role='ERROR' test='count(h:extension/h:HVACGEBControl) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/HVACGEBControl</sch:assert>
```
shorowit, line 325: Delete?

docs/source/workflow_inputs.rst
```diff
@@ -2936,18 +2936,22 @@ HPXML HVAC Control
...
+  =======================================================  ========  =====  ===========  ========  =========  ===========================================================
   ``SystemIdentifier``                                     id                            Yes                  Unique identifier
   ``HeatingSeason``                                        element                       No        See [#]_   Heating season        
   ``CoolingSeason``                                        element                       No        See [#]_   Cooling season
   ``extension/CeilingFanSetpointTempCoolingSeasonOffset``  double    F      >= 0         No        0          Cooling setpoint temperature offset [#]_
-  =======================================================  ========  =====  ===========  ========  =========  ========================================
+  ``extension/OnOffThermostatDeadbandTemperature``         double    F      >= 0         No        0          On-off thermostats/deadbands temperature [#]_
+  ``extension/TwospeedRealisticStaging``                   boolean                       No        false      Whether to apply two speed DX system realistic staging [#]_
```
shorowit, line 2947: The documentation is sorely out of date. These elements don't live in `HVACControl` anymore, there's no `TwospeedRealisticStaging` element, etc.

shorowit (commented):
There is a lot to digest in this PR. I am providing some initial review comments below but I only made it through a portion of the PR. I think it's going to be hard to get this into merge-able shape for the upcoming release, and I don't want to hold up the release just for this. We will likely do an incremental OS-HPXML 1.8.1 release that we can target this for.

HPXMLtoOpenStudio/measure.rb
```diff
@@ -180,11 +180,13 @@ def run(model, runner, user_arguments)
         check_schedule_references(hpxml_bldg.header, args[:hpxml_path])
         in_schedules_csv = 'in.schedules.csv'
         in_schedules_csv = "in.schedules#{i + 1}.csv" if i > 0
+        # Fixme: do we need to check system number of speed? Default hasn't been applied yet, so compressor can be unknown.
```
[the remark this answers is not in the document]
yzhou601: No this is a question that I modified the setpoint without knowing what's the number of speed at this execution point, but there's a check of number of speed later in hvac.rb that will throw an error if it doesn't qualify the limitation to applying on-off thermostat, so no worry now.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
yzhou601: The first thought makes sense, we still need to check the systems at building level instead of a sum. Let me think how to do that.
The reason to only enable with DX system is because we're modeling the cycling behavior for start-up degradation, which only applies to DX system if I recall it correctly.
I didn't recall why I just checked it's greater than 0.99, probably forgot to add the other side checking, let me rethink about it with the building level check.

shorowit: I think a building-level check would use a Schematron context that is something like: `/h:HPXML[h:SoftwareInfo/h:extension/h:SimulationControl/h:AdvancedResearchFeatures/h:OnOffThermostatDeadbandTemperature]/h:Building/h:BuildingDetails`

yzhou601: I found that the schema only allows maximum fraction to be 1.0, so if we limit to only allow one cooling system/heat pump then it has to be smaller than 1.0. I was trying to add the other side check but I cannot add test in test_validation.rb due to the schema constraint. So I think the check of 0.99 should be sufficient in this case.

yzhou601: For the second item, I'm thinking maybe we can add a method in measure.rb that checks system type and throw error or warning if needed. There're a few checks in hvac.rb, but they're under specific method call (eg. [this for variable speed system](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac.rb#L1850-L1852) but it won't be called when apply_air_source_hvac_systems is not called. Maybe measure.rb is the right place to check the feature availability for those only apply to specific system types)

yzhou601:
@shorowit @jmaguire1 There's a lot of refactoring happening lately, Jeff and I went through those changes and looked at simulation results together last Friday, and we felt happy with the current implementation. There're a few potential improvements on coding, I've addressed most of the comments from @shorowit , @shorowit let me know if you want to discuss the remaining questions or have any other questions.

yzhou601:
Only one question after addressing the comments from @shorowit 's first round review. 

I removed the rule in the EPvalidator that only allows certain system types, then should we throw a warning when there's no system that is applicable to the new features? I'm thinking maybe we can add a method in measure.rb that checks system type and throw error or warning if needed. There're a few similar checks in hvac.rb, but they're under specific method call (eg. [this for variable speed system](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac.rb#L1850-L1852) but it won't be called when apply_air_source_hvac_systems is not called. Maybe measure.rb is the right place to check the feature availability for those only apply to specific system types)

shorowit:
I'd be okay with a single warning if no HVAC systems are found in the home that would respond to the input(s), but I also don't think it's that important to provide a warning so long as it's documented which system types the input(s) apply to.

(It's kind of like someone specifying the heat pump sizing methodology without the home having a heat pump. We don't issue a warning for that.)

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -30,6 +31,27 @@ def self.apply_air_source_hvac_systems(model, runner, cooling_system, heating_sy
...
+          runner.registerError('On-off thermostat deadband currently is only supported for single speed or two speed air source systems.')
+        end
+        if hpxml_bldg.building_construction.number_of_units > 1
+          # Throw error and stop simulation
+          runner.registerError('NumberofUnits greater than 1 is not supported for on-off thermostat deadband.')
+        end
+        if hpxml_header.timestep > 1
+          # Throw error and stop simulation
+          runner.registerError('On-off thermostat deadband currently is only supported for 1 minute timestep.')
+        end
```
shorowit, line 49: Isn't this caught by EPvalidator? Seems like we can remove this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1130,8 +1162,9 @@ def self.create_setpoint_schedules(runner, heating_days, cooling_days, htg_weekd
   #
   # @param hvac_control [TODO] TODO
   # @param year [TODO] TODO
+  # @param offset_db [Float] On-off thermostat deadband
```
shorowit, line 1165:
```suggestion
  # @param offset_db [Float] On-off thermostat deadband in deg-F
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1170,8 +1206,9 @@ def self.get_heating_setpoints(hvac_control, year)
   # @param has_ceiling_fan [TODO] TODO
   # @param year [TODO] TODO
   # @param weather [WeatherProcess] Weather object
+  # @param offset_db [Float] On-off thermostat deadband
```
shorowit, line 1209:
```suggestion
  # @param offset_db [Float] On-off thermostat deadband in deg-F
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
     end
   end
 
+  # TODO
```
shorowit, line 3367: Update this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+  # @return [Float] Time in minutes to reach full capacity
+  def self.calc_time_to_full_cap(c_d)
+    # assuming a linear relationship between points we have data for: 2 minutes at 0.08 and 5 minutes at 0.23
+    time = (20.0 * c_d + 0.4).round
+    time = [time, get_time_to_full_cap_limits[0]].max
+    time = [time, get_time_to_full_cap_limits[1]].min
+    return time
+  end
+
+  # TODO
```
shorowit, line 3379: And this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+  # @return [2, 5] Minimum and maximum time in minutes to reach full capacity
+  def self.get_time_to_full_cap_limits()
+    return [2, 5]
+  end
+
+  # TODO
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio Model object
+  # @param htg_supp_coil [OpenStudio::Model::CoilHeatingElectric or CoilHeatingElectricMultiStage] OpenStudio Supplemental Heating Coil object
+  # @return [OpenStudio::Model::EnergyManagementSystemActuator, OpenStudio::Model::EnergyManagementSystemGlobalVariable] OpenStudio EMS Actuator and Global Variable objects for supplemental coil availability schedule
```
shorowit, line 3390:
```suggestion
  # @return [Array<OpenStudio::Model::EnergyManagementSystemActuator, OpenStudio::Model::EnergyManagementSystemGlobalVariable>] OpenStudio EMS Actuator and Global Variable objects for supplemental coil availability schedule
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+    global_var_supp_avail_program.setName("#{global_var_supp_avail.name} init program")
+    global_var_supp_avail_program.addLine("Set #{global_var_supp_avail.name} = 1")
+    manager = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    manager.setName("#{global_var_supp_avail_program.name} calling manager")
+    manager.setCallingPoint('BeginZoneTimestepBeforeInitHeatBalance')
+    manager.addProgram(global_var_supp_avail_program)
+    return supp_coil_avail_act, global_var_supp_avail
+  end
+
+  # TODO
```
shorowit, line 3419: And this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+    supp_coil_avail_program.addLine('EndIf')
+
+    # ProgramCallingManagers
+    program_calling_manager = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    program_calling_manager.setName("#{supp_coil_avail_program.name} ProgramManager")
+    program_calling_manager.setCallingPoint('InsideHVACSystemIterationLoop')
+    program_calling_manager.addProgram(supp_coil_avail_program)
+  end
+
+  # TODO
```
shorowit, line 3510: And this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+    cycling_degrad_program.addLine("Set #{ec_actuator.name} = ec_out / cc_mult")
+
+    # ProgramCallingManagers
+    program_calling_manager = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    program_calling_manager.setName("#{cycling_degrad_program.name} ProgramManager")
+    program_calling_manager.setCallingPoint('InsideHVACSystemIterationLoop')
+    program_calling_manager.addProgram(cycling_degrad_program)
+  end
+
+  # TODO
```
shorowit, line 3624: And this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+  # @param runner [OpenStudio::Measure::OSRunner] OpenStudio Runner object
+  # @param air_loop_unitary [TODO] TODO
+  # @param control_zone [OpenStudio::Model::ThermalZone] OpenStudio Thermal Zone object for conditioned space
+  # @param heating_system [TODO] TODO
+  # @param cooling_system [TODO] TODO
+  # @param htg_supp_coil [TODO] TODO
+  # @param clg_coil [TODO] TODO
+  # @param htg_coil [TODO] TODO
+  # @param schedules_file [SchedulesFile] SchedulesFile wrapper class instance of detailed schedule files
+  # @return [TODO] TODO
```
shorowit, line 3760: And these.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3511,6 +3364,766 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+    program.addLine('EndIf')
+
+    # calling manager
+    program_calling_manager = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    program_calling_manager.setName(program.name.to_s + ' calling manager')
+    program_calling_manager.setCallingPoint('InsideHVACSystemIterationLoop')
+    program_calling_manager.addProgram(program)
+  end
+
+  # TODO
```
shorowit, line 4007: And this.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2092,388 +2133,169 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
+    backup_heating_capacity_increment = nil unless fuel == HPXML::FuelTypeElectricity
+    if not backup_heating_capacity_increment.nil?
+      if hpxml_bldg.building_construction.number_of_units > 1
+        # Throw error and stop simulation
+        runner.registerError('NumberofUnits greater than 1 is not supported for multi-staging backup coil.')
+      end
+      if hpxml_header.timestep > 1
+        # Throw error and stop simulation
+        runner.registerError('Multi-staging backup coil currently is only supported for 1 minute timestep.')
+      end
```
shorowit, line 2211: Again, shouldn't need this since it's covered by EPvalidator.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2092,388 +2133,169 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
-    htg_spt_sensor.setKeyName(control_zone.name.to_s)
-    clg_spt_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Zone Thermostat Cooling Setpoint Temperature')
-    clg_spt_sensor.setName("#{control_zone.name} clg_spt_temp")
-    clg_spt_sensor.setKeyName(control_zone.name.to_s)
-    load_sensor = OpenStudio::Model::EnergyManagementSystemSensor.new(model, 'Unitary System Predicted Sensible Load to Setpoint Heat Transfer Rate')
-    load_sensor.setName("#{air_loop_unitary.name} sens load")
-    load_sensor.setKeyName(air_loop_unitary.name.to_s)
+      num_stages = [(capacity / backup_heating_capacity_increment).ceil(), max_num_stages].min
+      # OpenStudio only supports 4 stages for now
+      runner.registerWarning("E+ Currently only supports less than #{max_num_stages} stages for multi-stage electric backup coil. Combined the remaining capacities in the last stage. Simulation continued.") if (capacity / backup_heating_capacity_increment).ceil() > 4
```
shorowit, line 2216:
```suggestion
      runner.registerWarning("EnergyPlus only supports #{max_num_stages} stages for multi-stage electric backup coil. Combined the remaining capacities in the last stage.") if (capacity / backup_heating_capacity_increment).ceil() > 4
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2092,388 +2133,169 @@ def self.disaggregate_fan_or_pump(model, fan_or_pump, htg_object, clg_object, ba
...
-    temp_offset_program.setName("#{air_loop_unitary.name} temp offset init program")
-    temp_offset_program.addLine("Set #{temp_offset_signal.name} = 0")
+      (1..num_stages).each do |stage_i|
+        stage = OpenStudio::Model::CoilHeatingElectricMultiStageStageData.new(model)
+        if stage_i == max_num_stages
+          increment = (capacity - stage_capacity) # Model remaining capacity anyways
+        else
+          increment = backup_heating_capacity_increment
+        end
+        next if increment <= 5 # Tolerance to avoid modeling small capacity stage, Fixme: Is 5 a good tolerance?
```
shorowit, line 2229: Need to address FIXME.

BuildResidentialHPXML/measure.rb
```diff
@@ -127,6 +127,18 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
 
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('simulation_control_onoff_thermostat_deadband', false)
+    arg.setDisplayName('Simulation Control: HVAC On-Off Thermostat Deadband')
+    arg.setDescription('Research feature to model on-off thermostat deadband and start-up degradation for single or two speed central AC/ASHP systems, and realistic time-based staging for two speed central AC/ASHP systems. Currently only supported with 1 min timestep.')
+    arg.setUnits('deg-F')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('simulation_control_heat_pump_backup_heating_capacity_increment', false)
+    arg.setDisplayName('Simulation Control: Heat Pump Backup Heating Capacity Increment')
+    arg.setDescription("Research feature to model capacity increment of the multi-staging heat pump backup system with time-based staging. Only applies if Backup Type is '#{HPXML::HeatPumpBackupTypeIntegrated}' and Backup Fuel Type is '#{HPXML::FuelTypeElectricity}'. Currently only supported with 1 min timestep.")
```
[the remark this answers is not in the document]
yzhou601: The time-based control logic is quite similar between backup multi-stage coil and two speed DX coil. So I referenced the same language, but they're applying to different coil objects. The first input applies to single speed and two speed systems, while the second input applies to backup system, and the second input can be applied to any speed with integrated electric backup.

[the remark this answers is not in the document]
shorowit: Ah right, there is the staging of the HP and the staging of the backup. I got confused in my head. Thanks.

workflow/tests/util.rb
```diff
@@ -382,6 +386,10 @@ def _verify_outputs(rundir, hpxml_path, results, hpxml, unit_multiplier)
     if hpxml_path.include? 'base-location-detailed.xml'
       next if message.include? 'Weather file location will be used rather than entered (IDF) Location object.'
     end
+    # Coil speed level EMS
+    if hpxml_header.hvac_onoff_thermostat_deadband
+      next if message.include?('Wrong coil speed EMS override value, for unit=') && message.include?('Exceeding maximum coil speed level.') # FIXME: speed level actuator throws this error when speed is set to 1 but no load
+    end
```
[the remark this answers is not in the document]
yzhou601: Just an explanation, removed the "FIXME" label

docs/source/workflow_inputs.rst
```diff
@@ -116,6 +118,19 @@ These features may require shorter timesteps, allow more sophisticated simulatio
...
+         Applies to single speed and two speed air-source AC/ASHP only.
+  .. [#] An on/off thermostat deadband temperature is applied to both heating and cooling.
+         Typical values are between 2-4 F, with actual values being specific to the thermostat installed.
+         Note that thermostat deadbands are two sided.
+         As an example, if you had a heating setpoint of 71 F and a 2 F deadband, the heating equipment will turn on when the space temperature hits 70 F and off when it hits 72 F.
+         When this feature is enabled, the model will also explicitly model cycling, such that it will take several minutes for the HVAC to reach full capacity for single and two speed air-source AC/ASHP systems, and time-based realistic staging (stay at low speed for 5 minutes before transitioning to the higher stage, and stay at high speed until cut-out deadband temperature is reached) for two speed air-source AC/ASHP systems.
```
shorowit, line 128: I'm getting confused about when OnOffThermostatDeadbandTemperature applies. First, there are several references to "air-source AC/ASHP", which is redundant. ACs and ASHPs are by definition air-source. Second, I noticed that the changelog and BuildResidentialHPXML argument description both say "central AC/ASHP" instead, but there is a `base-hvac-room-air-conditioner-geb-onoff.xml` example file which suggests it does not apply to just central systems. Can you please clarify the language everywhere?

workflow/hpxml_inputs.json
```diff
@@ -1962,6 +1962,13 @@
     "hvac_control_heating_weekday_setpoint": "64, 64, 64, 64, 64, 64, 64, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 64, 64",
     "hvac_control_heating_weekend_setpoint": "64, 64, 64, 64, 64, 64, 64, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 64, 64"
   },
+  "sample_files/base-hvac-air-to-air-heat-pump-1-speed-lockout-temperatures-geb-multistage-backup.xml": {
```
shorowit, line 1965: Please remove all references to "geb". But more importantly, I would like to see fewer test files, 9 new test files feels overkill. And that's on top of 6 new test files for the defrost model. Can we just add a single "base-hvac-foo-research-features.xml" for each unique HVAC type that enables all the advanced features (defrost, deadband, multistage backup) at the same time? That would hopefully cut the 15 test files roughly in half.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3509,6 +3354,774 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+
+  # Apply EMS program to control back up coil behavior when single speed system is modeled with on-off thermostat feature.
+  # Back up coil is turned on after 5 mins that heat pump is not able to maintain setpoints.
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio Model object
+  # @param htg_supp_coil [OpenStudio::Model::CoilHeatingElectric or OpenStudio::Model::CoilHeatingElectricMultiStage] OpenStudio Supplemental Heating Coil object
+  # @param control_zone [OpenStudio::Model::ThermalZone] Conditioned space thermal zone
+  # @param htg_coil [OpenStudio::Model::CoilHeatingDXSingleSpeed or OpenStudio::Model::CoilHeatingDXMultiSpeed] OpenStudio Heating Coil object
+  # @param is_onoff_thermostat_ddb [Boolean] Whether to apply on off thermostat deadband
+  # @param cooling_system [HPXML::CoolingSystem] HPXML Cooling System
```
shorowit, line 3418: I think a lot of these methods operate on a HPXML::CoolingSystem or a HPXML::HeatPump, isn't that right?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3509,6 +3354,774 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+  # Apply time-based realistic staging EMS program for two speed system.
+  # Observe 5 mins before ramping up the speed level, or enable the backup coil.
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio Model object
+  # @param unitary_system [OpenStudio::Model::AirLoopHVACUnitarySystem] OpenStudio Air Loop HVAC Unitary System object
+  # @param htg_supp_coil [OpenStudio::Model::CoilHeatingElectric or OpenStudio::Model::CoilHeatingElectricMultiStage] OpenStudio Supplemental Heating Coil object
+  # @param control_zone [OpenStudio::Model::ThermalZone] Conditioned space thermal zone
+  # @param is_heatpump [Boolean] True if cooling_system is a heat pump
+  # @param is_onoff_thermostat_ddb [Boolean] Whether to apply on off thermostat deadband
+  # @param cooling_system [HPXML::CoolingSystem] HPXML Cooling System
```
shorowit, line 3627: You shouldn't need `is_heatpump`, you can determine if it's a heat pump based on the type of object that `cooling_system` is. Also need to indicate that `cooling_system` can be a HPXML::HeatPump too.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3509,6 +3354,774 @@ def self.set_cool_rated_shrs_gross(cooling_system)
...
+
+  # Apply maximum power ratio schedule for variable speed system.
+  # Creates EMS program to determine and control the stage that can reach the maximum power constraint.
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio Model object
+  # @param runner [OpenStudio::Measure::OSRunner] OpenStudio Runner object
+  # @param air_loop_unitary [OpenStudio::Model::AirLoopHVACUnitarySystem] Air loop for the HVAC system
+  # @param control_zone [OpenStudio::Model::ThermalZone] Conditioned space thermal zone
+  # @param heating_system [HPXML::HeatingSystem] HPXML Heating System
+  # @param cooling_system [HPXML::CoolingSystem] HPXML Cooling System
```
shorowit, line 3751: or HPXML::HeatPump?

shorowit (commented):
This looks very close, just a few comments left. Also, I did a little cleanup so please look it over and see if you think any changes were not correct. Thanks!

docs/source/workflow_inputs.rst
yzhou601: Sorry for the confusion, it does apply to room ac as well, so I tried to correct it in the docs, missed the changelog and buildresidential arguments. I'm going to fix those too.

workflow/hpxml_inputs.json
```diff
@@ -1962,6 +1958,24 @@
     "hvac_control_heating_weekday_setpoint": "64, 64, 64, 64, 64, 64, 64, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 64, 64",
     "hvac_control_heating_weekend_setpoint": "64, 64, 64, 64, 64, 64, 64, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 70, 64, 64"
   },
+  "sample_files/base-hvac-air-to-air-heat-pump-1-speed-lockout-temperatures-research-features.xml": {
+    "parent_hpxml": "sample_files/base-hvac-air-to-air-heat-pump-1-speed-lockout-temperatures.xml",
+    "simulation_control_timestep": 1,
+    "heat_pump_heating_capacity": 10000,
```
shorowit, line 1964: I assume you are setting a very low heating capacity so that the backup heat triggers a lot? It ends up being a heat pump with 10k heating capacity and 36k cooling capacity, which doesn't reflect what you would possibly see in reality. Would it be better to set the compressor lockout temperature a bit above the heating design temperature? That should similarly cause large backup heating usage during cold periods.

shorowit (commented):
This all looks good to me, thank @yzhou601! I was about to accept and merge it but had one last thought/comment below.

workflow/hpxml_inputs.json
yzhou601: Yes it's set in this weird way to trigger back up heat more. The lockout temperature will help trigger backup coil and see the staging of backup coil, but the behavior to switch to backup coil when main coil is not able to maintain setpoint will be missing. What about keeping 36k heating and cooling but using a more extreme climate?

shorowit: Ah, good point. Yes, you could try using Duluth for the climate. You could also decrease the heat pump capacity a bit if you really need to, but 10k is quite small.

HPXMLtoOpenStudio/tests/test_hvac.rb
```diff
@@ -754,11 +869,11 @@ def test_heat_pump_advanced_defrost
     assert_equal(1, model.getCoilHeatingDXMultiSpeeds.size)
     htg_coil = model.getCoilHeatingDXMultiSpeeds[0]
     # q_dot smaller than backup capacity
-    _check_advanced_defrost(model, htg_coil, 4747.75, 4747.75, backup_fuel, 0.06667, 1215.05)
+    _check_advanced_defrost(model, htg_coil, 4747.75, 4747.75, backup_fuel, 0.06667, 1067.05)
```
yzhou601, line 872: The change here is caused by watts_per_cfm used in the p_dot calculation changed due to the test file location change.

shorowit (approved):

PR #1656, opened 2024-03-21
EIA SEDS Price Data

bpark1327:
## Pull Request Description

Replace individual fuel prices with EIA [State Energy Data System (SEDS) prices](https://www.eia.gov/state/seds/seds-data-fuel.php?sid=US#DataFiles)

About SEDS: https://www.eia.gov/state/seds/

## Checklist

- [x] ~Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)~


ReportUtilityBills/resources/simple_rates/README.md
```diff
@@ -18,3 +18,9 @@ Propane
 * https://www.eia.gov/dnav/pet/xls/PET_PRI_WFR_A_EPLLPA_PRS_DPGAL_W.xls
 * Convert second tab (Data 1) to CSV
 * PET_PRI_WFR_A_EPLLPA_PRS_DPGAL_W.csv
+
+Wood
+* https://www.eia.gov/state/seds/seds-data-fuel.php?sid=US#DataFiles
+* Download Prices (CSV) link
+* Filter the data where "MSN" is labeled as "WDRCD" (wood price in the residential sector)
+* SEDS_price_of_wood.csv
```
shorowit, line 26: Interesting. It looks like this data source has costs for all fuel types. Perhaps we can replace all the individual EPA files we were using before for different fuel types with this single download? It's possible that we just weren't aware of it (or it's a more recently available data source). Can you compare the prices in here to what we downloaded from the other sources and see if they are basically the same?

That would simplify this process of updating costs in the future if we could just download a single new file and drop it in.

shorowit (commented):
Please go through the PR checklist. Also, one initial comment below.

ReportUtilityBills/resources/simple_rates/README.md
bpark1327: While electricity prices are the same, prices for other fuel types are quite different depending on the location. This raises doubts about whether using these wood prices is appropriate or not.

bpark1327:
@nmerket @shorowit I observed a discrepancy in fuel prices between SEDS and each individual fuel page, except for electricity. I reached out to the EIA, and I received a response stating that they were surprised by the discrepancy. The individual mentioned that he would consult their colleagues who work on SEDS and update me if the SEDS team can provide any clarification. I'll keep you informed once I have more information.

bpark1327:
@shorowit The 2024 data we saw yesterday was located in the "state-level profiles" (Not the SEDS). Upon further investigation, it seems to be sourced from the Navigators (the data source we have been using). I confirmed that the most recent data in the SEDS is from 2022. This appears to be the major issue with the SEDS. Do you think we should stick with what we've been using?

shorowit:
@bpark1327 That's unfortunate. I would probably still go with SEDS, it has all the fuel types we need and in a nice standardized format. All EIA data sources have some lag, so having a bit of a larger lag doesn't bother me _too_ much.

shorowit:
And boy, their website is confusing. Thanks for figuring all of this out.

bpark1327:
> And boy, their website is confusing. Thanks for figuring all of this out.

Sure, honestly, it was a bit painful. 😉

HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -105,47 +47,39 @@ def self.marginal_rate_to_average_rate(marginal_rate, fixed_charge, household_co
...
+      HPXML::FuelTypeWoodPellets => 'WDRCD'
+    }
+    unit_conv = {
+      HPXML::FuelTypeElectricity => 0.003412, # convert $/MMBtu to $/kWh
+      HPXML::FuelTypeNaturalGas => 0.1, # convert $/MMBtu to $/therms
+      HPXML::FuelTypeOil => 0.139, # convert $/MMBtu to $/gallons of oil
+      HPXML::FuelTypePropane => 0.0916, # convert $/MMBtu to $/gallons of propane
+      HPXML::FuelTypeCoal => 1, # $/MMBtu
+      HPXML::FuelTypeWoodCord => 1, # $/MMBtu
+      HPXML::FuelTypeWoodPellets => 1 # $/MMBtu
```
shorowit, line 67: Can you use `UnitConversions.convert(...)` here? That way we use consistent unit conversions everywhere. Also, just use MBtu, not MMBtu. We don't use the MMBtu convention anywhere and it will be confusing if we started to use it some places.

For the gal <-> MBtu conversion, since it's not strictly a unit conversion, I suggest adding a new method to the UnitConversions class to handle it.

ReportUtilityBills/measure.rb
```diff
@@ -677,8 +677,9 @@ def setup_utility_outputs()
   def get_outputs(fuels, utility_bill_scenario)
     fuels.each do |(fuel_type, _is_production), fuel|
       unit_conv = UnitConversions.convert(1.0, 'J', fuel.units)
-      unit_conv /= 139.0 if fuel_type == FT::Oil
-      unit_conv /= 91.6 if fuel_type == FT::Propane
+      unit_conv /= 139.0 if fuel_type == FT::Oil # Convert kBtu to gallons of oil
+      unit_conv /= 91.6 if fuel_type == FT::Propane # Convert kBtu to gallons of propane
+      unit_conv /= 1000.0 if [FT::Coal, FT::WoodCord, FT::WoodPellets].include? fuel_type # Convert kBtu to MMBtu
```
shorowit, line 682: Use `UnitConversions` here too. Having the same numbers two different places in the code can easily lead to bugs where someone in the future updates one but not the other. And MBtu instead of MMBtu again.

HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -105,47 +47,39 @@ def self.marginal_rate_to_average_rate(marginal_rate, fixed_charge, household_co
...
-    state_code_to_padd = {}
-    padd_to_state_codes.each do |padd, state_codes|
-      state_codes.each do |state_code|
-        state_code_to_padd[state_code] = padd
+      if fuel_type == HPXML::FuelTypeCoal
+        seds_rate = row[40] # Use 2007 prices for coal. For 2008 forward, EIA assumes there is zero residential sector coal consumption in the United States, and SEDS does not estimate a price.
+      else
+        seds_rate = row[-1]
+        seds_rate = row[-2] if seds_rate.nil? # If the rate for the latest year is unavailable, use the rate from the previous year.
       end
```
shorowit, line 80: When we update the prices in the future, A) the `[40]` for coal may no longer be correct and B) we have to look back more than a single year (`[-2]`) for a non-nil value. Please use more generic, future-proof logic. For example, simply return the last non-nil value for any fuel type.

ReportUtilityBills/resources/simple_rates/Average_retail_price_of_electricity.csv
shorowit, line 1: Yay! Goodbye to these files.

ReportUtilityBills/tests/test_report_utility_bills.rb
```diff
@@ -293,28 +293,6 @@ def test_auto_marginal_rate
     end
   end
 
-  def test_warning_region
```
shorowit, line 296: Awesome.

docs/source/workflow_inputs.rst
```diff
@@ -268,15 +268,9 @@ For each scenario, fuel rates can be optionally entered as an ``/HPXML/SoftwareI
...
-         \- **natural gas**: ``ReportUtilityBills/resources/Data/UtilityRates/NG_PRI_SUM_A_EPG0_PRS_DMCF_A.csv``
-         
-         \- **propane**: ``ReportUtilityBills/resources/Data/UtilityRates/PET_PRI_WFR_A_EPLLPA_PRS_DPGAL_W.csv``
-         
-         \- **fuel oil**: ``ReportUtilityBills/resources/Data/UtilityRates/PET_PRI_WFR_A_EPD2F_PRS_DPGAL_W.csv``
-         
-         or defaults to $0.015/kBtu for other fuels.
+  .. [#] If MarginalRate not provided, defaults to state, regional, or national average based on 2022 EIA SEDS data that can be found at ``ReportUtilityBills/resources/simple_rates/pr_all_update.csv``.
+
+         \- **coal**: based on 2007 EIA SEDS data. For 2008 forward, EIA assumes there is zero residential sector coal consumption in the United States, and SEDS does not estimate a price.
```
shorowit, line 273:
```suggestion
  .. [#] If MarginalRate not provided, defaults to state, regional, or national average based on 2022 EIA SEDS data that can be found at ``ReportUtilityBills/resources/simple_rates/pr_all_update.csv`` (except for coal, which is based on 2007 EIA SEDS data as EIA assumes there is zero residential sector coal consumption in the United States for 2008 forward and  does not estimate a price).
```

shorowit (commented):
Nice work, @bpark1327, looks very comprehensive. I have a few suggestions but this seems close. I also still need to do a little testing.

HPXMLtoOpenStudio/resources/unit_conversions.rb
```diff
@@ -17,6 +17,8 @@ class UnitConversions
     ['gj', 'kwh'] => 277.778,
     ['gj', 'therm'] => 9.48043,
     ['kwh', 'btu'] => 3412.141633127942,
+    ['kwh', 'kbtu'] => 3412.141633127942 / 1000.0,
+    ['kwh', 'mbtu'] => 3412.141633127942 / 1000.0 / 1000.0,
```
shorowit, line 21: These aren't needed, we do these conversions all the time. The code is smart enough to know that `[a, b] => X` is the same as `[b, a] => 1/X`.

HPXMLtoOpenStudio/resources/unit_conversions.rb
```diff
@@ -27,13 +29,18 @@ class UnitConversions
     ['mbtu', 'therm'] => 10.0,
     ['mbtu', 'wh'] => 293071.0701722222,
     ['therm', 'btu'] => 100000.0,
-    ['therm', 'kbtu'] => 100.0,
+    ['therm', 'kbtu'] => 100000.0 / 1000.0,
+    ['therm', 'mbtu'] => 100000.0 / 1000.0 / 1000.0,
```
shorowit, line 33: Not needed either.

HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -1,81 +1,23 @@
...
-        state_name = 'United States'
-      else
-        state_name = 'U.S.'
-      end
-    else
-      state_name = Constants.StateCodesMap[state_code]
-    end
-    return if state_name.nil?
+  def self.get_rates_from_eia_data(_runner, state_code, fuel_type, fixed_charge, marginal_rate = nil)
+    msn_codes = Constants.StateCodesMap.map { |x| x[0] }.uniq
```
shorowit, line 5: I'm not sure why you are ensuring a unique set (isn't it already unique?), but it seems like you just want the hash keys? Untested suggestion:

```suggestion
    msn_codes = Constants.StateCodesMap.keys
```

HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -105,47 +47,39 @@ def self.marginal_rate_to_average_rate(marginal_rate, fixed_charge, household_co
...
+
+    CSV.foreach(File.join(File.dirname(__FILE__), '../../ReportUtilityBills/resources/simple_rates/pr_all_update.csv'), headers: true) do |row|
+      next if row['State'].upcase != state_code.upcase # State
+      next if row['MSN'].upcase != msn_code_map[fuel_type] # EIA SEDS MSN code
+
+      if fuel_type == HPXML::FuelTypeCoal
+        seds_rate = row['2007'] # Use 2007 prices for coal. For 2008 forward, EIA assumes there is zero residential sector coal consumption in the United States, and SEDS does not estimate a price.
+      else
+        seds_rate = row.to_h.values.reverse.find { |rate| !rate.nil? && !rate.empty? } # If the rate for the latest year is unavailable, find the last non-nil rate.
       end
```
shorowit, line 78: Can we not simply collapse these into generic logic? Clearly a value of zero is not an actual fuel price. Untested suggestion:

```suggestion
      seds_rate = row.to_h.values.reverse.find { |rate| rate.to_f != 0 } # If the rate for the latest year is unavailable, find the last non-nil/non-zero rate.
```

shorowit (commented):
Just a couple comments/suggestions.

ReportUtilityBills/measure.rb
```diff
@@ -677,8 +677,9 @@ def setup_utility_outputs()
   def get_outputs(fuels, utility_bill_scenario)
     fuels.each do |(fuel_type, _is_production), fuel|
       unit_conv = UnitConversions.convert(1.0, 'J', fuel.units)
-      unit_conv /= 139.0 if fuel_type == FT::Oil
-      unit_conv /= 91.6 if fuel_type == FT::Propane
+      unit_conv = UnitConversions.convert(unit_conv, fuel.units, 'gal_fuel_oil') if fuel_type == FT::Oil # Convert kBtu to gallons of oil
+      unit_conv = UnitConversions.convert(unit_conv, fuel.units, 'gal_propane') if fuel_type == FT::Propane # Convert kBtu to gallons of propane
+      unit_conv = UnitConversions.convert(unit_conv, fuel.units, 'mbtu') if [FT::Coal, FT::WoodCord, FT::WoodPellets].include? fuel_type # Convert kBtu to MBtu
```
shorowit, line 682: This change doesn't seem right. Why are you converting to `$/MBtu`? We were previously reporting `$/kBtu`, which is [consistent with the documentation](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#fuel-rates).

shorowit (commented):
This is getting close, but I think there's a bug and I see you introduced a new warning.

HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -105,47 +47,41 @@ def self.marginal_rate_to_average_rate(marginal_rate, fixed_charge, household_co
...
+    CSV.foreach(File.join(File.dirname(__FILE__), '../../ReportUtilityBills/resources/simple_rates/pr_all_update.csv'), headers: true) do |row|
+      next if row['State'].upcase != state_code.upcase # State
+      next if row['MSN'].upcase != msn_code_map[fuel_type] # EIA SEDS MSN code
+
+      seds_rate = row.to_h.values.reverse.find { |rate| rate.to_f != 0 } # If the rate for the latest year is unavailable, find the last non-nil/non-zero rate.
+      begin
+        seds_rate = Float(seds_rate)
+      rescue ArgumentError, TypeError
+        seds_rate = 0.0
+        runner.registerWarning("No EIA SEDS rate for #{fuel_type} was found for the state of #{state_code}.") if not runner.nil?
```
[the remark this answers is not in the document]
bpark1327: This is an edge case that occurs specifically with coal in the state of Hawaii. I will add a test.

ReportUtilityBills/measure.rb
bpark1327: I must have been mistaken. I guess I should just delete it.

shorowit (approved):

PR #1661, opened 2024-03-28
Shared water heater updates

joseph-robertson:
## Pull Request Description

Fix some footnotes in water heater docs sections.
Create new `apply_shared_adjustment` method for apportioning shared water heater tank losses. Method is called for tank, tankless, heat pump, and combi water heater types.
Add a new `base-bldgtype-mf-unit-shared-water-heater-heat-pump.xml` sample file, along with a new unit test.

## Checklist


docs/source/workflow_inputs.rst
```diff
@@ -3588,7 +3588,6 @@ Each combination boiler w/ tankless coil is entered as a ``/HPXML/Building/Build
          
          \- **IECC zones 3-8, unknown**: "basement - unconditioned", "basement - conditioned", "conditioned space"
          
-  .. [#] NumberofUnitsServed only required if IsSharedSystem is true, in which case it must be > 1.
   .. [#] The sum of all ``FractionDHWLoadServed`` (across all WaterHeatingSystems) must equal to 1.
   .. [#] FractionDHWLoadServed represents only the fraction of the hot water load associated with the hot water **fixtures**.
          Additional hot water load from clothes washers/dishwashers will be automatically assigned to the appropriate water heater(s).
```
shorowit, line 3593: I can't comment on the exact line, but we should also get rid of the "Tank losses will be apportioned..." text below for combi boilers w/ tankless coil, since we assume no tank losses similar to a tankless system.

shorowit (commented):
These changes to the documentation for HPWHs and combi boilers made me second guess whether we actually are adjusting the tank losses for those system types. And it turns out we're not.

The only adjustment we're making happens [here](https://github.com/NREL/OpenStudio-HPXML/blob/bf4787bc154b4454b114328db0b1d5ade8e20d35/HPXMLtoOpenStudio/resources/waterheater.rb#L1556), but this method is only used for conventional storage tank and tankless systems.

Seems like we also need to make adjustments [here](https://github.com/NREL/OpenStudio-HPXML/blob/bf4787bc154b4454b114328db0b1d5ade8e20d35/HPXMLtoOpenStudio/resources/waterheater.rb#L759) for HPWHs and [here](https://github.com/NREL/OpenStudio-HPXML/blob/bf4787bc154b4454b114328db0b1d5ade8e20d35/HPXMLtoOpenStudio/resources/waterheater.rb#L164) for combi systems. For the combi systems, we need to make sure to do the adjustment after that method (not inside it) because the ERI code calls that method and needs the unadjusted value.

shorowit:
> For the combi systems, we need to make sure to do the adjustment after that method (not inside it) because the ERI code calls that method and needs the unadjusted value.

I take that back. We can (and probably should) do it inside of the method so long as we can pass a `nil` or similar from the ERI caller to indicate we don't want to apply the adjustment there.

HPXMLtoOpenStudio/tests/test_water_heater.rb
```diff
@@ -1061,6 +1061,41 @@ def test_shared_water_heater
...
+    args_hash['hpxml_path'] = File.absolute_path(File.join(sample_files_dir, 'base-bldgtype-mf-unit-shared-water-heater-heat-pump.xml'))
+    model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    # Get HPXML values
+    water_heating_system = hpxml_bldg.water_heating_systems[0]
+
+    # Expected value
+    tank_volume = UnitConversions.convert(water_heating_system.tank_volume * 0.9, 'gal', 'm^3') # convert to actual volume
+    fuel = EPlus.fuel_type(water_heating_system.fuel_type)
+    u =  0.1081
```
joseph-robertson, line 1075: How should we get this expected value? How does `test_tank_heat_pump` get it?

shorowit:
@joseph-robertson Please go through the PR checklist.

joseph-robertson:
> @joseph-robertson Please go through the PR checklist.

Oops. Done.

shorowit (approved):
Looks good, thanks! Before merging this in, I want to see if I can get a quick [docs footnote check](https://github.com/NREL/OpenStudio-HPXML/pull/1673) working and pull it in here to show that this PR fixes the issue.

Issue #1023, opened 2020-08-12
Activities found even when the occupancy is zero (stochastic schedule generator)


hleeNREL:
Rajendra provided a schedule for a single occupancy unit in TX, and I found that human activities exist when occupancy was zero.

![occupancy error](https://user-images.githubusercontent.com/68304805/90048382-4df84280-dc88-11ea-9f46-f30e477326bc.png)

The image shows that sink and range were used.  There are more spots and many water usages of sink were found.  I also found bath and shower intermittently.

rajeee:
I might have found the source of this bug. We randomly shift the generated schedule by +- 30 minutes to avoid synchronization, but apparently this random shift is calculated independently for each schedule. It should have been picked once and reused for all schedule columns. 
https://github.com/NREL/OpenStudio-HPXML/blob/459eeb3ed80e59e6be041e031b3bf9944c29c366/BuildResidentialScheduleFile/resources/schedules.rb#L590

shorowit:
@rajeee I'm a bit confused, I thought you were fixing this as part of the EV schedules PR but when I look at the occupancy-stochastic.csv file in that feature, I am still seeing e.g. hot water usage when there are no occupants. Can you clarify?

![Image](https://github.com/user-attachments/assets/9ae6f2e7-6030-406a-a1db-d2dc4670ca73)

PR #1757, opened 2024-06-20
Generate EV schedules

rajeee:
## Pull Request Description

We leverage the existing American Time Use Survey (ATUS) guided stochastic occupancy generator to generate the EV battery charging and discharging schedule. The schedule generator takes total hours driven per week as an input to the EV schedule generator and produces the charging and discharging schedule. The battery model takes care of the actual power draw during the charging and discharging period. An example of what these schedules look like is presented in Figure 1.  This schedule shows that during an “away period” for the occupant, the vehicle is discharged after the occupant leaves the home. Then later in the day the vehicle is discharged again while the occupant travels home. 

image
 Figure 1 Illustration of EV battery discharge schedule generation. 

 

**Discharging Schedule Generation** 

The guiding theme for the discharge schedule is the assumption that the EV is owned/associated with just one of occupant in the house. We look at the away duration for each occupant and eliminate any occupant whose away hours would not be sufficient to fit the given number of driving hours per year. Then we randomly pick one of the occupant out of the eligible occupant. If no occupant happens to have sufficient away hours then we pick the occupant with the highest away hours and truncate the driving hours to whatever hours the occupant away schedule supports.

Once the occupant is chosen, we assume that the EV battery discharge will start immediately after they leave the home, and immediately before they arrive – simulating the scenario that they left the home driving the EV, stayed somewhere for a while and then returned home. Following that assumption, we sum the total away duration for the occupant, and proportionally assign the driving hours to each away period. The driving hours assigned to each away period is distributed symmetrically at the start and end – always making sure to leave the center 20% as idle to allow for some idle duration at the destination away from home. This algorithm is illustrated in the Figure 1. 

**Charging Schedule Generation** 

Currently, we assume that the occupant plugs in their vehicle as soon as they are home and keeps it plugged in until they leave. So, the charging schedule is identical to the occupancy schedule of the EV owning occupant. This makes the battery eligible to be charged anytime the EV is at home if the state of charge is below 100%. Whether or not the charging happens and at what power level is managed by the battery model. The TEMPO model makes similar assumptions in its “immediate” or “ASAP” charging strategy, as described in Yip et al. (2023). 

**Handling of fractional charging at home** 

Not all EV owners charge exclusively at home. RECS 2020 survey has a question about what fraction of their charging is done at home with options being 100%, 80 to 99%, 60 to 79% etc. This diversity of charging behavior is handled by proportionally reducing the driving hours and calculating effective driving hours. The underlying assumption in using this data is that respondents answered the fraction of their charging in terms of energy, as opposed to time. The "hours driven per week" fed into the schedule generator is assumed to have already applied this adjustment. 

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1016,6 +1016,8 @@ def initialize(name, used_by_unavailable_periods, can_be_stochastic, type)
   # periods CSV (e.g., heating), and/or C) EnergyPlus-specific schedules (e.g., battery_charging).
   Columns = {
     Occupants: Column.new('occupants', true, true, :frac),
+    EVOccupant: Column.new('ev_occupant', true, true, :int),
+    PresentOccupants: Column.new('present_occupants', true, true, :int),
```
shorowit, line 1020: Why do we need these two new columns? As far as I can tell they aren't used, and they seem confusing.

HPXMLtoOpenStudio/resources/schedules.rb
```diff
@@ -1137,7 +1141,11 @@ def import(schedules_paths)
         end
 
         if @schedules.keys.include? col_name
-          fail "Schedule column name '#{col_name}' is duplicated. [context: #{schedules_path}]"
+          if col2path[col_name] == schedules_path
+            fail "Schedule column name '#{col_name}' is duplicated. [context: #{schedules_path}]"
+          else
+            @runner.registerWarning("Schedule column name '#{col_name}' already exist in #{col2path[col_name]}. Overwriting with #{schedules_path}.")
+          end
```
shorowit, line 1148: If a column name shows up twice in the same file, it's an error, but if it shows up in different files, the last file is used? Do I have that right? I don't understand why we would want to treat the two situations differently.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -1074,6 +1123,75 @@ def get_building_america_lighting_schedule(time_zone_utc_offset, latitude, longi
...
+  end
+
+  # TODO
+  #
+  # @param markov_chain_simulation_result [TODO] TODO
+  # @return [TODO] TODO
+  def fill_ev_battery_schedule(markov_chain_simulation_result)
+    if @hpxml_bldg.vehicles.to_a.empty?
+      return
+    end
```
shorowit, line 1182: I'm a bit confused. I see that no `ev_battery_charging` and `ev_battery_charging` columns are created when there's no EV. This is different than other things (like clothes washers), where we create schedules even when the appliance is not present. Why?

It's particularly confusing that when you have no EV, you still get an `ev_occupant` (and `present_occupants`) columns. Why?

Finally, I ran this measure on the `base-battery-ev.xml` file and it's showing 7 occupants?! The HPXML file has 3 bedrooms and doesn't have the number of occupants specified. Where did 7 come from?

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -902,6 +906,52 @@ def sum_across_occupants(all_simulated_values, activity_index, time_index, max_c
     return sum
   end
 
+  # TODO
+  #
+  # @param all_simulated_values [TODO] The Markov-chain activity array
```
shorowit, line 911: You will at least need to replace all these TODOs for newly written code. (Even better would be if you could go through this file and document existing code too so that others can better understand the code.)

shorowit (commented):
Some initial comments/questions. I didn't scrutinize the actual methodology/engineering behind how the EV schedule is created.

HPXMLtoOpenStudio/resources/schedules.rb
rajeee: EVOccupant is the occupancy schedule of the occupant that is driving the EV. 
PresentOccupants is the occupancy of all occupants.
EVOccupant column can be used, for example, to simulate demand response for EV where EV charging is delayed. Without knowing the EV occupant schedule it won't be possible to align charging with the time the occupant is home.
PresentOccupants is mostly for debugging/postprocessing purpose. I have used this to generate graphics like this below where I am looking at the away hours of EV occupants compared to away hours of all occupants.
image

HPXMLtoOpenStudio/resources/schedules.rb
rajeee: Yeah, you got that right. But I am unsure myself why I needed to add this. I will come back to this!

BuildResidentialScheduleFile/resources/schedules.rb
rajeee: Yeah, this is the part we need to decide one way or the other. For other appliance, we can just generate the schedule based on the markov chain. For EV it's not so straight forward since it needs annual driving hours as an input to generate the schedule but that's not available if there is no EV in the HPXML. So, currently, I am not generating the EV schedule. But, yes, this is inconsistent with what we do with other appliances. We do have that input from ResStock though (even if there is no EV) - it's just not available in the HPXML file. We need to find a solution here.

BuildResidentialScheduleFile/resources/schedules.rb
rajeee: Sure, I can do this!

HPXMLtoOpenStudio/resources/schedules.rb
shorowit: I see. Thanks for the description. What it cause any issues for your use case if we only exported these columns when the `debug` flag is used?

rajeee: Yeah, I think we can switch to using debug flag for outputting these. Or, can this be controlled through output variables?

HPXMLtoOpenStudio/resources/schedules.rb
rajeee: I can't recall why I had to add this, and this change is also failing unit test. So, I am going to revert it for now.

shorowit (approved):
This overall looks good to me and can merge in when the CI comes back with a ✅ after the last commit. A couple final thoughts:
1. I wish we also had a stochastic EV charging schedule for when the HPXML has an EV charging `PlugLoad`. It will probably be confusing that a stochastic schedule is only available when the HPXML has an EV `Vehicle`.
2. It looks like a lot of the hot water-related stochastic schedules have changed. I can't immediately tell why. Is it because you tried to address [this](https://github.com/NREL/OpenStudio-HPXML/issues/1023)? Either way, we should probably document it in the changelog.

PR #1760, opened 2024-06-27
Leakiness Description

### LINKED ISSUE #1652: Allow qualitative infiltration inputs (average, tight, etc.)
Allow qualitative infiltration inputs (average, tight, etc.) using HPXML's [LeakinessDescription](https://hpxml.nrel.gov/datadictionary/3.1.0/Building/BuildingDetails/Enclosure/AirInfiltration/AirInfiltrationMeasurement/LeakinessDescription) as an alternative to quantitative blower door values.

We already have the LBNL regression coded into HEScore, so we could move it here and make it more broadly available. And for Manual J, we could implement the Table 5A defaults and update our ACCA test files accordingly.

Related to [the discussion here](https://github.com/NREL/OpenStudio-HPXML/pull/1578#discussion_r1531366554).

cc @yzhou601

yzhou601:
## Pull Request Description

Fix #1652. Allows alternative infiltration input `AirInfiltrationMeasurement/LeakinessDescription`, in which the infiltration level is estimated using age of home, climate zone, foundation type, etc.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -2286,8 +2289,6 @@ def self.apply_hvac_distribution(hpxml_bldg, ncfl, ncfl_ag)
       end
 
       # Calculate FractionDuctArea from DuctSurfaceArea
-      supply_ducts = hvac_distribution.ducts.select { |duct| duct.duct_type == HPXML::DuctTypeSupply }
-      return_ducts = hvac_distribution.ducts.select { |duct| duct.duct_type == HPXML::DuctTypeReturn }
```
shorowit, line 2290: I don't think you can remove these lines of code. The code above potentially adds a new duct, so we need to get the updated list of supply/return ducts again.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -2865,8 +2868,6 @@ def to_location
         return LocationCrawlspaceVented
       elsif @foundation_type == FoundationTypeCrawlspaceConditioned
         return LocationCrawlspaceConditioned
-      elsif @foundation_type == FoundationTypeSlab
-        return LocationConditionedSpace
```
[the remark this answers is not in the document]
yzhou601: It's just a dead code that the Slab is already an option in the first condition so it looks that it will never step here

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -84,6 +84,9 @@ def self.apply(runner, hpxml, hpxml_bldg, eri_version, weather, epw_file: nil, s
     # Do HVAC sizing after all other defaults have been applied
     apply_hvac_sizing(runner, hpxml_bldg, weather, output_format, design_load_details_output_file_path)
 
+    # Set infiltration values for energy simulation after sizing
+    apply_leakiness_description(hpxml_bldg)
```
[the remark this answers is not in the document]
yzhou601: This is not because it needs information from hvac sizing, but I do want hvac sizing uses only the leakiness description information when only leakiness description is provided. And whenever both leakiness and ach/cfm are provided by user, process ach/cfm as a higher priority. So it ends in this way, probably not the best way so let me know if you have thoughts on how to improve.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
yzhou601: Good point, I'm going to add these lines back.

HPXMLtoOpenStudio/resources/hpxml.rb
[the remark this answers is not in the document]
shorowit: Aha, I didn't notice that. Good catch!

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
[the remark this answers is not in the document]
shorowit: Oh wow, interesting issue. I don't think we've ever run into this kind of situation before. I'll give it some thought...

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
```diff
@@ -3696,6 +3697,21 @@ def self.apply_hvac_sizing(runner, hpxml_bldg, weather, output_format, design_lo
...
+  # @param hpxml_bldg [HPXML::Building] HPXML Building object representing an individual dwelling unit
+  # @return [void]
+  def self.apply_leakiness_description(hpxml_bldg)
+    measurement = Airflow.get_infiltration_measurement_of_interest(hpxml_bldg.air_infiltration_measurements)
+    return if measurement.unit_of_measure || measurement.effective_leakage_area
+
+    ach50 = Airflow.calc_ach50(hpxml_bldg)
+    measurement.house_pressure = 50
+    measurement.unit_of_measure = HPXML::UnitsACH
+    measurement.air_leakage = ach50
```
shorowit, line 3711: You'll also want to set `measurement.air_leakage_isdefaulted = true` and ensure that the HPXML class uses it to assign the `dataSource` attribute for in.xml. It is important that a user see that the ACH50 value has been defaulted.

yzhou601: Yes, I just did that today locally and made a few other changes trying to wrap up the development processes. Thanks for bringing this up!

yzhou601:
This is getting really close, last TODO is to update the ACCA test files and add/modify unit test to make sure the Manual J infiltration table look up is correct.

yzhou601:
@shorowit This PR is ready for review.

There're just a few test files that I got confused by the way Manual J used the Table 5A/5B to look up values (eg.  the cfa it used in the look up and the volume it used in the calculation. In the Bob Ross Residence word file, Wes recommends to use 3001 sqft to look up, I infer that it's 2400sqft * 10ft / 8ft(default height in the table), so I implemented in this way instead of directly using cfa specified in the HPXML).

BuildResidentialHPXML/measure.rb
```diff
@@ -1076,6 +1073,18 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
+    leakiness_description_choices << HPXML::LeakinessVeryTight
+    leakiness_description_choices << HPXML::LeakinessTight
+    leakiness_description_choices << HPXML::LeakinessAverage
+    leakiness_description_choices << HPXML::LeakinessLeaky
+    leakiness_description_choices << HPXML::LeakinessVeryLeaky
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('leakiness_description', leakiness_description_choices, false)
+    arg.setDisplayName('Air Leakage: Leakiness Description')
+    arg.setDescription('Qualitative description of infiltration. If provided, Manual J Table 5A/5B look up ACH values will be used for hvac sizing, and the annual average ACH will be calculated based on year built, iecc zone, cfa, infiltration height, foundation type, ducts and leakiness description.')
+    args << arg
```
shorowit, line 1086: A few suggestions:
1. Move this input above the other air leakage inputs, since it's the more generic input.
2. Use a name consistent with the other other air leakage arguments: `air_leakage_leakiness_description`
3. No need for a detailed description of what will happen for Manual J and the energy model here, that should be in the documentation.
4. In the code below, let's use the quantitative input if it's provided, else the qualitative input. I don't think we need to support the unusual use case where someone wants to provide both. And we should update the description of the quantitative input to say that it will override the qualitative input if provided.
5. Since the quantitative defaults were removed, there are now no defaults at all for infiltration. A user of the measure will therefore hit an error unless they specifically enter something. In general, we'd like to have reasonable defaults. I suggest we put back in a default -- probably the qualitative input of "average" is a good default.

Changelog.md
```diff
@@ -2,11 +2,13 @@
 
 __New Features__
 - Adds inputs for modeling skylight curbs and/or shafts.
+- Allow qualitative infiltration inputs with `AirInfiltrationMeasurement/LeakinessDescription` and `BuildingConstruction/YearBuilt`.
 - Central Fan Integrated Supply (CFIS) mechanical ventilation enhancements:
   - Allows modeling systems with no strategy to meet remainder of ventilation target (`CFISControls/AdditionalRuntimeOperatingMode="none"`).
 - HVAC Manual J design load and sizing calculations:
   - Adds optional inputs and outputs for blower fan heat and piping load.
   - Miscellaneous improvements.
+  - Adds optional `InfiltrationMethod` input to specify which infiltration inputs to use for manual J load calculations.
```
shorowit, line 11:
```suggestion
  - Adds optional `HVACSizingControl/ManualJInputs/InfiltrationMethod` input to specify which infiltration method to use for manual J load calculations.
```

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -277,6 +286,109 @@ def self.get_values_from_air_infiltration_measurements(hpxml_bldg, cfa, weather)
     return { sla: sla, ach50: ach50, nach: nach, volume: infil_volume, height: infil_height, a_ext: a_ext }
   end
 
+  # Calculate ACH50 for annual energy simulation when only leakiness description is provided
+  #
+  # @param hpxml_bldg [HPXML::Building] HPXML Building object representing an individual dwelling unit
+  # @return [Float] Calculated ACH50 value based on IECC zone, cfa, infiltration height, year built, foundation type, ducts and leakiness description
+  def self.calc_ach50(cfa, ncfl_ag, year_built, ceil_height, infil_volume, iecc_cz, foundations, ducts, leakiness_description)
```
shorowit, line 293: Need to describe all the arguments, no longer `hpxml_bldg`.

Also, we need an `air_sealed` argument (default to false for OS-HPXML) if we are eventually going to call this from the HEScore code. And it will need to be incorporated it into the calculations below.

shorowit (commented):
Didn't get very far in my review, but a couple initial comments.

BuildResidentialHPXML/measure.rb
yzhou601: I have a question, if quantitative input is overriding the qualitative input, and with default assumptions in the place, then it will always have values in quantitative arguments no matter if it's defaulted or user-provided, how can we get the qualitative input?

shorowit: > I have a question, if quantitative input is overriding the qualitative input, and with default assumptions in the place, then it will always have values in quantitative arguments no matter if it's defaulted or user-provided, how can we get the qualitative input?

Sorry, I don't think I'm understanding. You removed the default quantitative inputs, and I was suggested "average" as the qualitative input. So if they leave everything as is, or if they change the qualitative input (to e.g. "tight"), then the HPXML will only have that value.

HPXMLtoOpenStudio/resources/airflow.rb
```diff
@@ -288,18 +288,30 @@ def self.get_values_from_air_infiltration_measurements(hpxml_bldg, cfa, weather)
...
+  # @param year_built [Integer] Year the dwelling unit is built
+  # @param ceil_height [Double] Average ceiling height (ft2)
+  # @param infil_volume [Double] Infiltration volume (ft2)
+  # @param iecc_cz [String] IECC climate zone
+  # @param foundations [Array<Array>] Array of [fraction, foundation_type] pair for each foundation
+  # @param ducts [Array<Array>] Array of [fraction, duct_location] pair for each duct
+  # @param leakiness_description [String] Leakiness description to qualitatively describe the dwelling unit infiltration
+  # @param air_sealed [Boolean] True if the dwelling unit is air sealed
+  # @return [Double] Calculated ACH50 value based on IECC zone, cfa, infiltration height, year built, foundation type, ducts and leakiness description
+  def self.calc_ach50(cfa, ncfl_ag, year_built, ceil_height, infil_volume, iecc_cz, foundations, ducts, leakiness_description = nil, is_sealed = false)
```
shorowit, line 302: I'm confused, why do you allow for leakiness_description=nil? Don't we only need to call this method if it's not nil? If it's nil, then we should already have the quantitative leakage value.

HPXMLtoOpenStudio/resources/hpxml_defaults.rb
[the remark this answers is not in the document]
shorowit: Just documenting our solution here: we decided to introduce a new optional ManualJ input to specify which infiltration method to use.

HPXMLtoOpenStudio/resources/airflow.rb
yzhou601: Yes that's right for our workflow, but for current HEScore it uses air_sealed instead of leakiness_description, we should only have one being passed between these two.

BuildResidentialHPXML/measure.rb
yzhou601: I see, I thought you mean putting back all the default values (both quantitative and qualitative), then we're in agreement to only allow 'average' as default.

HPXMLtoOpenStudio/resources/airflow.rb
shorowit: Ah, right. Now that you say that, I realize we don't actually need the `air_sealed` input (sorry!) because HEScore uses [a mapping](https://hescore-hpxml.readthedocs.io/en/latest/translation/about.html#air-sealing-present) between `leakiness_description` and `air_sealed`. So, for HEScore, if air_sealed=true, we could pass in tight, and if it's false, we could pass in average. That would keep the method simple.

shorowit: Oh, you're right. Okay, in that case, let's just keep it how you have it. When we go to use this in HEScore, we can decide if we want to simplify/change things or not at that point.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -3166,17 +3172,17 @@ def self.get_ventilation_data(hpxml_bldg)
...
-    q_exh_tot = vent_mech_exh_tot.map { |vent_mech| vent_mech.average_total_unit_flow_rate }.sum(0.0)
-    q_bal_tot = vent_mech_bal_tot.map { |vent_mech| vent_mech.average_total_unit_flow_rate }.sum(0.0)
-    q_erv_hrv_tot = vent_mech_erv_hrv_tot.map { |vent_mech| vent_mech.average_total_unit_flow_rate }.sum(0.0)
-    q_cfis_tot = vent_mech_cfis_tot.map { |vent_mech| vent_mech.average_total_unit_flow_rate }.sum(0.0)
+    # In-unit CFMs (include recirculation from in unit CFMs for shared systems)
+    q_sup_tot = vent_mech_sup_tot.map { |vent_mech| vent_mech.total_unit_flow_rate }.sum(0.0)
+    q_exh_tot = vent_mech_exh_tot.map { |vent_mech| vent_mech.total_unit_flow_rate }.sum(0.0)
+    q_bal_tot = vent_mech_bal_tot.map { |vent_mech| vent_mech.total_unit_flow_rate }.sum(0.0)
+    q_erv_hrv_tot = vent_mech_erv_hrv_tot.map { |vent_mech| vent_mech.total_unit_flow_rate }.sum(0.0)
+    q_cfis_tot = vent_mech_cfis_tot.map { |vent_mech| vent_mech.total_unit_flow_rate }.sum(0.0)
```
shorowit, line 3180: @yzhou601 I'm seeing some diffs that are caused by these changes. I don't really understand why you were proposing these changes. I'm going to revert the changes and we can discuss this when you get back.

shorowit: Oh, I think I see now. The Bob Ross test has 50 cfm of outdoor air supplied through the HVAC system. In the HPXML file, we defined a CFIS system with 50 cfm, but because the default is to run the CFIS system for 8 hrs/day, we weren't actually introducing 50 cfm of outdoor air each hour. So we should really use 150 cfm at 8 hrs/day. I will revert the code changes and update our test files, and then we should be good.

shorowit (approved):
This feature was surprisingly tricky/nuanced, but I think we got there. Results look good. Thanks for your work here, @yzhou601!

PR #1767, opened 2024-07-02
Purge `epw_file` object

shorowit:
## Pull Request Description

Purges the `epw_file` object from the code. Now *all* weather-related data can be found on the `weather` object, so you don't have to pass around both objects or decide which one you need.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/weather.rb
```diff
@@ -251,6 +252,35 @@ def calc_ashrae_622_wsf(rowdata, epw_file)
     return wsf.round(2)
   end
 
+  # TODO
+  #
+  # @param epw_file [OpenStudio::EpwFile] OpenStudio EpwFile object
+  # @return [TODO] TODO
+  def get_header_info_from_epw(epw_file)
+    header.City = epw_file.city
```
shorowit, line 260: This is what allows purging the `epw_file` object. We store any relevant information from that object on the weather object, so that the weather object has *everything*.

BuildResidentialHPXML/measure.rb
```diff
@@ -3772,9 +3772,13 @@ def argument_errors(args)
...
-    epw_file = OpenStudio::EpwFile.new(epw_path)
-    if (args[:hvac_control_heating_season_period].to_s == HPXML::BuildingAmerica) || (args[:hvac_control_cooling_season_period].to_s == HPXML::BuildingAmerica) || (args[:apply_defaults])
-      weather = WeatherProcess.new(epw_path: epw_path, runner: nil)
+    if (args[:hvac_control_heating_season_period].to_s == HPXML::BuildingAmerica) ||
+       (args[:hvac_control_cooling_season_period].to_s == HPXML::BuildingAmerica) ||
+       (args[:solar_thermal_system_type] != 'none' && args[:solar_thermal_collector_tilt].start_with?('latitude')) ||
+       (args[:pv_system_present] && args[:pv_system_array_tilt].start_with?('latitude')) ||
+       (args[:pv_system_2_present] && args[:pv_system_2_array_tilt].start_with?('latitude')) ||
+       (args[:apply_defaults])
+      weather = WeatherFile.new(epw_path: epw_path, runner: nil)
```
joseph-robertson, line 3781: Put this stuff in its own method that returns true or false, and then just call something like `if need_weather_based_on_args(args) ...`?

shorowit: Yeah, I got a little lazy here...

Really it would be better to obtain the `weather` object on an as-need basis (and only if it hasn't already been obtained). Let me take a quick stab at that.

shorowit: Meh, it's getting more complicated than I like. I'll just go with your suggestion.

joseph-robertson (approved):
Looks good. Thanks for the other misc updates as well.

PR #1761, opened 2024-06-27
Update hot water end uses for operational calculations

shorowit:
## Pull Request Description

For operational calculations (where `NumberofResidents` is provided), we now directly use equations from the [FSEC paper](http://www.fsec.ucf.edu/en/publications/pdf/fsec-pf-464-15.pdf) that are a function of number of occupants.

## Checklist

Not all may apply:

- [ ] Schematron validator (`EPvalidator.xml`) has been updated
- [ ] Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)
- [x] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)
- [x] Documentation has been updated
- [x] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [x] No unexpected changes to simulation results of sample files

PR #1775, opened 2024-07-11
Update plug load end uses for operational calculations

### LINKED ISSUE #1773: Use RECS PlugLoads = f(Nocc) equations
Rather than having plug loads solely a function of CFA, we can use the equations from ResStock for operational calculations:
https://github.com/NREL/resstock/blob/061f3370bb53d14018d7c95f7f766d32003dfb11/measures/ResStockArguments/measure.rb#L467-L475

shorowit:
## Pull Request Description

Closes #1773. For operational calculations (where NumberofResidents is provided), we now directly use equations from RECS 2020 that are a function of number of occupants.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/misc_loads.rb
```diff
@@ -279,25 +279,63 @@ def self.apply_pool_or_permanent_spa_pump(runner, model, pool_or_spa, conditione
...
+    else # Operational calculation
+      # RECS 2020
+      if unit_type == HPXML::ResidentialTypeSFD
+        annual_kwh = 786.9 + 241.8 * num_occ + 0.33 * cfa
+      elsif unit_type == HPXML::ResidentialTypeSFA
+        annual_kwh = 654.9 + 206.5 * num_occ + 0.21 * cfa
+      elsif unit_type == HPXML::ResidentialTypeApartment
+        annual_kwh = 706.6 + 149.3 * num_occ + 0.10 * cfa
+      elsif unit_type == HPXML::HPXML::ResidentialTypeManufactured
+        annual_kwh = 1795.1 # No good relationship found in RECS, so just using a constant value
```
shorowit, line 302: Bringing attention to this. If anyone has a better idea, I'm all ears.

HPXMLtoOpenStudio/resources/misc_loads.rb
```diff
@@ -279,25 +279,63 @@ def self.apply_pool_or_permanent_spa_pump(runner, model, pool_or_spa, conditione
...
+  # @param unit_type [String] HPXML::ResidentialTypeXXX type of dwelling unit
+  # @return [Array<Double, Double, Double>] Television annual use (kWh), sensible fraction, and latent fraction
+  def self.get_televisions_default_values(cfa, nbeds, num_occ, unit_type)
+    if num_occ.nil? # Asset calculation
+      # ANSI/RESNET/ICC 301
+      annual_kwh = 413.0 + 69.0 * nbeds
+    else # Operational calculation
+      # RECS 2020
+      # Note: If we know # of televisions, we could use these better relationships instead:
+      # - SFD: 67.7 + 243.4 * num_tv
```
shorowit, line 325: Here too.

HPXMLtoOpenStudio/resources/misc_loads.rb
afontani: I guess I am not surprised by the lack of relationship for `num_occ` and `cfa`. What about the number of bedrooms? I know it is a little strange to introduce a separate variable just for Manufactured homes. I do like the idea of the value changing based on some variable and not constant, but if we don't have a clear trend then maybe we should keep it constant.

HPXMLtoOpenStudio/resources/misc_loads.rb
afontani: When @joseph-robertson was looking at regression equations, number of TVs was by far the variable to include in a regression. I would be in favor of using `num_tv`.

docs/source/workflow_inputs.rst
```diff
@@ -578,7 +578,7 @@ Building occupancy is entered in ``/HPXML/Building/BuildingDetails/BuildingSumma
   =======================================================  ========  =====  ===========  ========  ========  ========================
 
   .. [#] If NumberofResidents not provided, an *asset* calculation is performed assuming standard occupancy, in which occupant-driven end uses (e.g., plug loads, appliances, hot water) are calculated based on NumberofBedrooms.
-         If NumberofResidents is provided, an *operational* calculation is instead performed, in which hot water end uses are calculated based on NumberofResidents from `Estimating Daily Domestic Hot-Water Use in North American Homes <http://www.fsec.ucf.edu/en/publications/pdf/fsec-pf-464-15.pdf>`_ and all other occupant-driven end uses are adjusted using the following relationship from `RECS 2015 <https://www.eia.gov/consumption/residential/reports/2015/overview/>`_:
+         If NumberofResidents is provided, an *operational* calculation is instead performed, in which *some* occupant-driven end uses, as described throughout the documentation, are adjusted using the following relationship from `RECS 2015 <https://www.eia.gov/consumption/residential/data/2015/>`_:
```
afontani, line 581: Is there a plan to transition to RECS 2020 for number of bedrooms?

HPXMLtoOpenStudio/resources/misc_loads.rb
shorowit: Right. We currently don't have an input for # tvs in OS-HPXML. If we did add one, it would have to be optional, and we would need to have a default (based on # occupants and/or # bedrooms?).

I guess the idea you're proposing is that, if we use the equations based on # tvs, it might give a similar result to the equations I implemented when the number of TVs is not known, but if someone who _does_ know the number of TVs gets a better result.

HPXMLtoOpenStudio/resources/misc_loads.rb
shorowit: @joseph-robertson's analysis shows that all the different combinations of independent variables (CFA, # bedrooms, # occupants) have very low R2 values for MH. Worse, all the coefficients are negative. So, for example, the equation based on # bedrooms would be:
`annual_kwh = 2299.72 - 191.7 * num_br`
and would have lower electricity use when there are more bedrooms. Which I don't think makes sense.

docs/source/workflow_inputs.rst
shorowit: Yes, @prsh5175 was working on that [here](https://github.com/NREL/OpenStudio-HPXML/pull/1690), but we ran into issues with hot water use. Thus we ended up pursuing an [alternative hot water use approach](https://github.com/NREL/OpenStudio-HPXML/pull/1761). Assuming testing looks good and it gets merged, we should be able to revisit and wrap up Prateek's PR after that.

shorowit:
Create issues for @afontani's review comments before merging.

Issue #1421, opened 2023-06-26
Model total water usage


shorowit:
Water usage is increasingly of interest. Could use [ANSI/RESNET/ICC 850](https://codes.iccsafe.org/content/RESNET8502020P1/chapter-4-home-water-rating-calculation-procedures) to model the missing water end uses (toilet, outdoor irrigation, etc).

joseph-robertson:
After modeling the missing water end uses, we can then revive something like https://github.com/NREL/OpenStudio-HPXML/pull/1786.

PR #1786, opened 2024-07-25
Total water use outputs

### LINKED ISSUE #1785: Add outputs for mixed hot water usage along with hot water usage
Right now we default the hot water usage for all end uses (fixtures, associated waste, clothes washers, and dishwashers). For mixed end uses (fixtures and associated waste), it'd be nice to have the hot and total (cold + hot) usage, since the inputs are generally the total and then Fmix gets calculated based on the mains water temperature to determine what fraction of the total is hot.

I think this is relatively straightforward to just default to including "Total Water Volume" outputs as well as "Hot Water Volume" Outputs. We could alternately include "Cold Water Volume", but total is generally what we want to look at. Total volume and hot volume will be the same for CW and DW, so we probably only need this for fixtures and waste.

For reference these are the available outputs in E+ form the .rdd:

```
Output:Variable,*,Water Use Equipment Hot Water Volume,hourly; !- HVAC Sum [m3]
Output:Variable,*,Water Use Equipment Cold Water Volume,hourly; !- HVAC Sum [m3]
Output:Variable,*,Water Use Equipment Total Volume,hourly; !- HVAC Sum [m3]

joseph-robertson:
## Pull Request Description

Closes #1785.

## Checklist

Not all may apply:


shorowit:
I'm on the fence here. On one side, I can see it being a useful output. On the other hand, it has the potential to be misleading/misinterpreted because it's really the total water use *associated only with mixed temperature water draws*, as we do not model cold water draws/use (though there's a [feature request](https://github.com/NREL/OpenStudio-HPXML/issues/1421) for it).

shorowit:
We discussed this during our call this morning and are leaning towards not proceeding with this change. At the point at which we model all (i.e., cold water uses) as described in https://github.com/NREL/OpenStudio-HPXML/issues/1421, then it would make sense to add it.

I'll leave this open for another week or two in case anyone else has further comments.

joseph-robertson:
> We discussed this during our call this morning and are leaning towards not proceeding with this change. At the point at which we model all (i.e., cold water uses) as described in #1421, then it would make sense to add it.
> 
> I'll leave this open for another week or two in case anyone else has further comments.

So in the future, if we modeled all cold water uses, we'd combine cold water uses + cold water associated with mixed temperature draws, into a single output?

shorowit:
Right, we would report total water use (hot + mixed + cold) like you did here. It'd be almost exactly like this PR except there would be some additional cold-only end uses. So hopefully we could just borrow this code at the point at which we do that.

PR #1791, opened 2024-07-29
Fix scheduled battery w/unavailable period

joseph-robertson:
## Pull Request Description

`Error: Could not find row='battery_charging' in unavailable_periods.csv` is thrown for an HPXML file w/scheduled battery and unavailable period.

The fix is to add entries in unavailable_periods.csv for battery charging and discharging; these schedules are unaffected by vacancy but _affected_ by power outage (i.e., assumes you're not capable of islanded operation -- see below for more information about this).

## Checklist

Not all may apply:

- [ ] ~Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)~


shorowit (commented):
I wonder if we should downgrade the error to a warning. Something like this?

> Warning: Could not find row='battery_charging' in unavailable_periods.csv; it will not be affected by the unavailable period[1].

[1] Ideally we could more specifically say "power outage" or "vacancy".

HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
```diff
@@ -33,3 +33,5 @@ house_fan,0,1
 kitchen_fan,1,1
 bath_fan,1,1
 whole_house_fan,0,1
+battery_charging,0,1
+battery_discharging,0,0
```
shorowit, line 37: Need to think about charging/discharging with respect to both PV and grid. The simplest solution for now would be to disallow charging/discharging during a power outage even if that's not the common scenario, but it would take effort to support something different.

shorowit:
Comment from @jmaguire1: Do we need to look at other schedules (like water heater GEB schedules) that might similarly be impacted?

How can we make sure we don't miss these things going forward?

joseph-robertson:
> Comment from @jmaguire1: Do we need to look at other schedules (like water heater GEB schedules) that might similarly be impacted?
> 
> How can we make sure we don't miss these things going forward?

The sample file `base-schedules-detailed-mixed-timesteps-power-outage.xml` includes a defined power outage period as well as references to:
* detailed HVAC setpoint schedules
* detailed water heater setpoint schedule
* stochastic occupancy schedules

jmaguire1:
OK cool, glad we already have a test file with all the schedules together! I thought we might but wasn't sure.

HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
jmaguire1: I said this yesterday as well, but: I'd vote in favor of that, at least for the time being. It's basically assuming you're not capable of islanded operation [https://www.energysage.com/energy-storage/islanding-and-batteries-what-you-need-to-know/](url).

I'd certainly like to support islanded operation at some point, but that's going to require an EMS program to control charging to only happen from PV, and not the grid (so that part isn't super complicated, but is an extra step). The other issue is if you truly are in a home with solar + storage that can island and charge off PV, you're probably using some loads, at least as much as the battery can support them, which is not something we model right now. A lot of homes like this have what's called a critical loads panel [,](https://www.energysage.com/energy-storage/what-are-critical-load-panels/) and that's the only thing that gets powered.

I'd love to find some funds to be able to explicitly model trying to run off solar plus storage (or a backup generator?) during a power outage, but I think that's a big enough list it'd require some separate $$.

HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
```diff
@@ -33,3 +33,5 @@ house_fan,0,1
 kitchen_fan,1,1
 bath_fan,1,1
 whole_house_fan,0,1
+battery_charging,0,1
+battery_discharging,0,1
```
joseph-robertson, line 37: This is the fix.

PR #1825, opened 2024-09-06
Fix negative defrost power when applying advanced defrost with small cooling capacity

yzhou601:
## Pull Request Description

Refer to [this line](https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/hvac.rb#L4822),
q_dot/cop is based on cooling capacity, which can be smaller than the odu fan power, causing negative defrost power in the idf.

cc @joseph-robertson  @jonwinkler 

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4815,24 +4815,28 @@ def self.calculate_heat_pump_defrost_load_power_watts(heat_pump, unit_multiplier
     power_design = fan_watts_per_cfm * design_airflow_1x
     p_dot_blower = power_design * defrost_power_fraction
     # Based on manufacturer data for ~70 systems ranging from 1.5 to 5 tons with varying efficiency levels
-    p_dot_odu_fan = 44.348 * UnitConversions.convert(nominal_cooling_capacity_1x, 'Btu/hr', 'ton') + 62.452
-    rated_clg_cop = heat_pump.additional_properties.cool_rated_cops[-1]
+    if nominal_cooling_capacity_1x > 18000.0 # > 1.5 tons
+      p_dot_odu_fan = 44.348 * UnitConversions.convert(nominal_cooling_capacity_1x, 'Btu/hr', 'ton') + 62.452
+    else # < 1.5 tons, scale fan power to avoid negative p_dot_defrost
+      p_dot_odu_fan = 128.974 * (nominal_cooling_capacity_1x / 18000.0)
```
yzhou601, line 4821: @jonwinkler The actual change is here, I scaled down the fan power when cooling capacity is smaller than 1.5 tons, which is the smallest data point for regression.

jonwinkler: Thanks, @yzhou601. I think this looks reasonable. It took me a second to figure out how the constant `128.974` was determined, but I see now. Maybe add a comment to the code mentioning how that was calculated.

yzhou601: Comment added, this PR will be ready for review after ci runs green

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5085,14 +5085,14 @@ def self.apply_installation_quality(model, heating_system, cooling_system, unita
...
+  # @param heat_pump [HPXML::HeatPump] HPXML Heat Pump object
   # @param unit_multiplier [Integer] Number of similar dwelling units
-  # @param design_airflow [TODO] TODO
-  # @param max_heating_airflow [TODO] TODO
-  # @param fan_watts_per_cfm [TODO] TODO
-  # @return [TODO] TODO
+  # @param design_airflow [Float] Heat pump design air flow rate
+  # @param max_heating_airflow [Float] Maximum heat pump heating air flow rate
+  # @param fan_watts_per_cfm [Float] Heat pump watts per cfm
+  # @return [Array<Float, Float>] Calculated delivered cooling q_dot and compressor power p_dot
```
shorowit, line 5095: Thanks for filling this out. Please add units (power in W? cooling load in ?? airflow rates in cfm?)

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5085,14 +5085,14 @@ def self.apply_installation_quality(model, heating_system, cooling_system, unita
     program_calling_manager.addProgram(fault_program)
   end
 
-  # TODO
+  # Calculate delivered cooling q_dot and compressor power p_dot during defrost
```
shorowit, line 5088:
```suggestion
  # Calculate delivered cooling load and compressor power during defrost when using
  # the advanced defrost model.
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5117,24 +5117,29 @@ def self.calculate_heat_pump_defrost_load_power_watts(heat_pump, unit_multiplier
...
     q_dot_defrost = UnitConversions.convert(nominal_cooling_capacity_1x, 'Btu/hr', 'W') * capacity_defrost_multiplier
-    cop_defrost = rated_clg_cop * cop_defrost_multiplier
+    cop_defrost = heat_pump.additional_properties.cool_rated_cops[-1] * cop_defrost_multiplier
     p_dot_defrost = (q_dot_defrost / cop_defrost - p_dot_odu_fan + p_dot_blower) * unit_multiplier # p_dot_defrost is used in coil object, which needs to be scaled up for unit multiplier
+
     return q_dot_defrost, p_dot_defrost
   end
 
-  # TODO
+  # Create EMS program and Other equipment objects to account for delivered cooling and supplemental heating energy during defrost
```
shorowit, line 5133:
```suggestion
  # Create EMS program and Other equipment objects to account for delivered cooling load and supplemental heating energy during defrost
  # when using the advanced defrost model.
```

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5117,24 +5117,29 @@ def self.calculate_heat_pump_defrost_load_power_watts(heat_pump, unit_multiplier
...
   # @param air_loop_unitary [OpenStudio::Model::AirLoopHVACUnitarySystem] Air loop for the HVAC system
-  # @param conditioned_space [TODO] TODO
-  # @param htg_supp_coil [TODO] TODO
-  # @param heat_pump [TODO] TODO
-  # @param q_dot_defrost [TODO] TODO
-  # @return [TODO] TODO
+  # @param conditioned_space [OpenStudio::Model::Space]  OpenStudio Space object for conditioned zone
+  # @param htg_supp_coil [OpenStudio::Model::CoilHeatingElectric or CoilHeatingElectricMultiStage] OpenStudio Supplemental Heating Coil object
+  # @param heat_pump [HPXML::HeatPump] HPXML Heat Pump object
+  # @param q_dot_defrost [Float] Calculated delivered cooling q_dot
```
shorowit, line 5141: Add units

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5117,24 +5117,29 @@ def self.calculate_heat_pump_defrost_load_power_watts(heat_pump, unit_multiplier
     power_design = fan_watts_per_cfm * design_airflow_1x
```
shorowit, line 5117: All these `foo_1x` variables are pretty confusing. It'd be simpler if you did the calculations normally (without adjustments for the unit multiplier) and then just apply the unit multiplier as needed at the point at which you are assigning values to the EnergyPlus objects. For example, you might apply the unit multiplier to an object field [like this](https://github.com/NREL/OpenStudio-HPXML/blob/88e256c9f0be00a35eca1d521513d644af8631fe/HPXMLtoOpenStudio/resources/waterheater.rb#L953) or apply the unit multiplier to an EMS actuator [like this](https://github.com/NREL/OpenStudio-HPXML/blob/88e256c9f0be00a35eca1d521513d644af8631fe/HPXMLtoOpenStudio/resources/waterheater.rb#L1782-L1784).

shorowit (commented):
The bugfix seems reasonable, but a few comments...

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: I think the challenge here is that the airflow and capacities are already with unit multiplier at calling, but the regression is based on capacity per system, so I have to revert first then apply unit multiplier at the end. I've tried my best to simplify a bit and added a comment here.

shorowit: Ah, fair enough, that's a good reason to keep it how it is. Thanks for doing what you could do to simplify it.

shorowit (approved):
Looks good to me. Will merge assuming CI is happy.

Issue #1759, opened 2024-06-25
Ruby documentation: remaining TODOs


joseph-robertson:
- BuildResidentialHPXML
  - [x] @joseph-robertson measure.rb
  - [x] @joseph-robertson resources/geometry.rb
- BuildResidentialScheduleFile
  - [x] @joseph-robertson measure.rb
  - [x] @rajeee @afontani resources/schedules.rb https://github.com/NREL/OpenStudio-HPXML/pull/1757
  - [ ] resources/README.md (https://github.com/NREL/OpenStudio-HPXML/pull/1784#discussion_r1700774489)
  - [ ] resources/schedules.csv (https://github.com/NREL/OpenStudio-HPXML/pull/1784#discussion_r1706194920)
- HPXMLtoOpenStudio
  - [x] measure.rb
    - [x] @joseph-robertson move methods to resource files to make measure.rb simpler
  - [x] @shorowit resources/airflow.rb
  - [x] @joseph-robertson resources/battery.rb
  - [x] @shorowit resources/constants.rb
  - [ ] @shorowit resources/constructions.rb
  - [x] @shorowit resources/energyplus.rb
  - [x] @joseph-robertson resources/generator.rb
  - [x] @joseph-robertson resources/geometry.rb
  - [x] @shorowit resources/hotwater_appliances.rb
  - [x] @shorowit resources/hpxml.rb
  - [x] @shorowit resources/hpxml_defaults.rb
    - [x] @shorowit Move defaulting methods from other resource files into here
  - [x] @shorowit resources/hvac.rb
  - [x] @shorowit resources/hvac_sizing.rb
  - [x] @shorowit resources/lighting.rb
  - [x] @joseph-robertson resources/location.rb
  - [x] @shorowit resources/materials.rb
  - [x] @joseph-robertson resources/meta_measure.rb
  - [x] @shorowit resources/misc_loads.rb
  - [x] @joseph-robertson resources/model.rb
    - [ ] Create general object creation methods in here
  - [x] @shorowit resources/output.rb
  - [x] @shorowit resources/psychrometrics.rb
  - [x] @joseph-robertson resources/pv.rb
  - [x] @joseph-robertson resources/schedules.rb
  - [x] @shorowit resources/simcontrols.rb
  - [x] @shorowit resources/unit_conversions.rb
  - [x] @joseph-robertson resources/util.rb
  - [x] @shorowit resources/version.rb
[...]

PR #1840, opened 2024-09-19
More ruby documentation; prototype general model.rb add methods

joseph-robertson:
## Pull Request Description

Addresses #1759.

Also starts prototyping generic `Model.add_foo` to model.rb to consolidate code that creates OSM objects. It makes the objects easier to use, and it allows us to introduce additional logic that applies to every object of the same type (like using EMS-friendly variable names or avoiding the need for a developer to remember to use `.to_s` on an object name to get a string).

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -4,21 +4,21 @@
...
-# @param print_prefix [TODO] TODO
-# @param ep_input_format [TODO] TODO
-# @param skip_simulation [TODO] TODO
-# @param suppress_print [TODO] TODO
-# @return [TODO] TODO
-def run_hpxml_workflow(rundir, measures, measures_dir, debug: false, output_vars: [],
-                       output_meters: [], run_measures_only: false, print_prefix: '',
-                       ep_input_format: 'idf', skip_simulation: false, suppress_print: false)
+# @param rundir [String] The run directory containing all simulation output files
+# @param measures [Hash] Map of OpenStudio-HPXML measure directory name => Array<Hash of measure arguments>
```
shorowit, line 8: Do you remember why it's `Array<Hash of measure arguments>` and not just `Hash of measure arguments`? I wonder if the use case for using an Array no longer applies and we could simplify this.

HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -4,21 +4,21 @@
...
+# @param run_measures_only [Boolean] True applies only OpenStudio Model measures, skipping forward transation and the simulation
+# @param ep_input_format [String] EnergyPlus input file format (idf, epjson)
+# @param suppress_print [Boolean] TODO
+# @return [Hash] Map of 'success' and 'runner' results
+def run_hpxml_workflow(rundir, measures, measures_dir, debug: false,
+                       run_measures_only: false, ep_input_format: 'idf', suppress_print: false)
+  output_vars = []
+  output_meters = []
+  print_prefix = ''
+  skip_simulation = false # True applies OpenStudio Model measures and forward translation, skipping the simulation
```
shorowit, line 20: I'd suggest deleting these variables and simplifying the code.

HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -312,29 +312,24 @@ def get_measure_instance(measure_rb_path)
...
-# @param lookup_file [TODO] TODO
-# @param measure_name [TODO] TODO
+# @param measure_args [OpenStudio::Measure::OSArgumentVector] Array of OpenStudio Measure Arguments
+# @param provided_args [Hash] Map of provided measure arguments to values
+# @param measure_name [String] OpenStudio-HPXML measure directory name
 # @param runner [OpenStudio::Measure::OSRunner] Object typically used to display warnings
-# @return [TODO] TODO
-def validate_measure_args(measure_args, provided_args, lookup_file, measure_name, runner = nil)
+# @return [Hash] Map of provided measure arguments to values
+def validate_measure_args(measure_args, provided_args, measure_name, runner = nil)
```
shorowit, line 320: I'm seeing that ResStock calls this method and passes `lookup_file`.

joseph-robertson: Hmm, from where?

joseph-robertson: Oh, nvm. I see it.

HPXMLtoOpenStudio/resources/meta_measure.rb
joseph-robertson: I do not. Yeah that crossed my mind as well.

shorowit: I suggest we change it, I'm not immediately seeing why it's needed. You could then simplify run_simulation.rb. In fact, I think it would mean that you could get rid of the use of `update_args_hash` in run_simulation.rb, at which point I think the method could be removed from meta_measure.rb entirely.

joseph-robertson: We've decided to remove the `update_args_hash` method, but keep the `measures = {"MeasureA": [{"arg1": "foo", "arg2": "bar", ...}, ...]}` structure. We need, for example, to be able to apply ReportSimulationOutput more than once in an `apply_measures` call.

joseph-robertson:
@shorowit Can you fill in remaining TODOs for `apply_foundation_wall` and `apply_foundation_slab` in HPXMLtoOpenStudio/resources/geometry.rb? I think they are related to kiva.

shorowit:
> @shorowit Can you fill in remaining TODOs for `apply_foundation_wall` and `apply_foundation_slab` in HPXMLtoOpenStudio/resources/geometry.rb? I think they are related to kiva.

Done.

PR #1850, opened 2024-10-04
Whole building Common Spaces

yzhou601:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/model.rb
```diff
@@ -779,21 +779,38 @@ def self.merge_unit_models(model, hpxml_osm_map)
       end
     end
 
-    hpxml_osm_map.values.each_with_index do |unit_model, unit_number|
+    unit_surface_to_obj_index_map = {} # map of unit model surface handle to whole building model object index
```
yzhou601, line 782: @shorowit Sadly that there's not a better solution to keep the surface-to-surface mapping information while knowing the handles after merging unit models (and I cannot do it in front, by assigning a surface in another model space as adjacent surface). The only solution I can only think of here is to use the indexes after calling model.addObjects. OS source codes specifically call out that the order will be kept, so it's fine for now, but we need to be careful if this is called more than once in the future.

HPXMLtoOpenStudio/resources/model.rb
```diff
@@ -804,6 +821,21 @@ def self.merge_unit_models(model, hpxml_osm_map)
...
+
+      unit_surface_handle = unit_model_objects[index].handle
+      surface = obj.to_Surface.get
+      adjacent_surface_index = unit_surface_to_obj_index_map[unit_model_map[unit_surface_handle]]
+      adjacent_surface = model_objects[adjacent_surface_index].to_Surface.get
+      next if surface.adjacentSurface.is_initialized
+
+      surface.setAdjacentSurface(adjacent_surface)
+      # Need to set the same construction to make OS working
+      adjacent_surface.setConstruction(surface.construction.get)
```
shorowit, line 837: I forget, did we discuss a plan for ensuring constructions are identical for the two walls that point to each other?

For shared boilers, I was proposing that one would describe the shared boiler once (in the first dwelling unit attached to it, e.g. [here](https://github.com/NREL/OpenStudio-HPXML/blob/c074c73f89f88dab8ecb52230098dd947648c69e/workflow/sample_files/base-bldgtype-mf-whole-building-shared-boilers-sequenced-operation.xml#L349-L362)), and then every other dwelling unit would have a boiler with the `sameas` attribute where no additional details would be provided (e.g., [here](https://github.com/NREL/OpenStudio-HPXML/blob/c074c73f89f88dab8ecb52230098dd947648c69e/workflow/sample_files/base-bldgtype-mf-whole-building-shared-boilers-sequenced-operation.xml#L878-L881)) so as to prevent inconsistency. If we went that path, ideally we could enforce it via the Schematron validation. Is that the path you're going to go? Or did you have another idea in mind?

shorowit (commented):
Great to see some progress! A couple initial questions/thoughts.

HPXMLtoOpenStudio/resources/model.rb
yzhou601: Yes that's what I think, add more error-checking in EPvalidator to make sure there's no inconsistent construction, did you disallow the boiler inputs when using sameas? I can do the same restriction or I can check the constructions are exactly the same.

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -1718,9 +1723,18 @@ def self.set_surface_interior(model, spaces, surface, hpxml_surface, hpxml_bldg)
...
-  def self.set_surface_exterior(model, spaces, surface, hpxml_surface, hpxml_bldg)
+  def self.set_surface_exterior(model, spaces, surface, hpxml_surface, hpxml_bldg, common_surface_id_map)
     exterior_adjacent_to = hpxml_surface.exterior_adjacent_to
     is_adiabatic = hpxml_surface.is_adiabatic
+    if not hpxml_surface.sameas_id.nil?
+      if common_surface_id_map.nil?
+        fail "unexpected surface sameas attribute assigned: #{surface.sameas_id}."
+      else
+        # Store surface id and OS surface object mapping, process later
+        common_surface_id_map[hpxml_surface.id] = surface.handle
```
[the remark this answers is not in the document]
yzhou601: I tried to use the AdditionalProperty to store the handles, but the handles changed after merging, while I think AdditionalProperty has them as strings (only allow values of string, double, integer and bool) so it won't be mapped to the new handles. I can store the hpxml ids as AdditionalProperty, which probably can remove one map from current implementation.

HPXMLtoOpenStudio/resources/model.rb
shorowit: I don’t remember if I got that far with the shared boiler implementation, but I would favor disallowing the inputs.

HPXMLtoOpenStudio/resources/geometry.rb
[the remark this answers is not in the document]
shorowit: Oh, right. Yes, seems like storing hpxml ids should work!

HPXMLtoOpenStudio/resources/model.rb
yzhou601: Checked the links, it seems that you disallowed additional inputs, I'm going to do the same.

HPXMLtoOpenStudio/resources/geometry.rb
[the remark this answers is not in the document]
yzhou601: Thanks Scott! This is really helpful, it turns out that if I store both hpxml id and sameas id as additional properties, I can directly eliminate all the maps (see my latest commit), sorry that I was in a rabbit hole of OS handles and it became unnecessarily complicated.

yzhou601:
The approach seems working! 🎉 Going to make this PR more concrete next.

HPXMLtoOpenStudio/resources/geometry.rb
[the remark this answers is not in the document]
shorowit: Awesome!

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -693,9 +693,19 @@
     </sch:rule>
   </sch:pattern>
 
+  <sch:pattern>
+    <sch:title>[AdjacentRimJoist]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Enclosure/h:RimJoists/h:RimJoist[h:SystemIdentifier/@sameas]'>
+      <sch:assert role='ERROR' test='count(h:ExteriorAdjacentTo) = 0'>Expected 0 element(s) for xpath: ExteriorAdjacentTo</sch:assert>
+      <sch:assert role='ERROR' test='count(h:InteriorAdjacentTo) = 0'>Expected 0 element(s) for xpath: InteriorAdjacentTo</sch:assert>
+      <sch:assert role='ERROR' test='count(h:Area) = 0'>Expected 0 element(s) for xpath: Area</sch:assert>
+      <sch:assert role='ERROR' test='count(h:Insulation/h:AssemblyEffectiveRValue) = 0'>Expected 0 element(s) for xpath: Insulation/AssemblyEffectiveRValue</sch:assert>
```
shorowit, line 702: Can we simply use `count(*)` or similar to prevent any detailed inputs? We'll also want a test_validation.rb test that shows these Schematron errors getting hit.

HPXMLtoOpenStudio/resources/model.rb
```diff
@@ -941,6 +940,21 @@ def self.merge_unit_models(model, hpxml_osm_map)
         end
       end
     end
+
+    model_objects.each do |obj|
```
shorowit, line 944: Currently it feels a bit more complicated than I would have expected just because you're operating on the objects of the final merged model, so you have to know which surfaces are linked and what space to attach them to. I wonder if this can be simplified.

yzhou601:
One more remaining question related to calling individual dwelling unit (wrote down in case I forgot): Should we enforce the surfaces with full descriptions to be specified in conditioned dwelling units so that they can be called alone? Technically for surfaces connecting two Building elements, users can specify the surface full description in either Building element, which will make the call of running individual dwelling unit error-prone. We can either: 
1. Only allow whole building simulation when sameas is provided, then the surface inputs can be specified in either Building element, or:
2. Only allow surface  full descriptions all in conditioned units, and the user can call building ids of conditioned units to simulate individual dwelling units
Honestly I was implementing assuming the former in the first place, considering how MulTEA uses this feature, but I'm fine with the second for more flexibility.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -2871,9 +2878,9 @@
...
+      <sch:assert role='ERROR' test='count(h:Roofs/h:Roof[h:InteriorAdjacentTo="conditioned space"]) + count(h:Floors/h:Floor[h:InteriorAdjacentTo="conditioned space" and (h:ExteriorAdjacentTo="attic - vented" or h:ExteriorAdjacentTo="attic - unvented" or ((h:ExteriorAdjacentTo="other housing unit" or h:ExteriorAdjacentTo="other heated space" or h:ExteriorAdjacentTo="other multifamily buffer space" or h:ExteriorAdjacentTo="other non-freezing space") and h:FloorOrCeiling="ceiling"))]) + count(h:Floors/h:Floor[h:SystemIdentifier/@sameas]) &gt;= 1'>There must be at least one ceiling or roof adjacent to conditioned space.</sch:assert>
```
yzhou601, line 2881: I relaxed this rule to count "sameas" floor also as a ceiling/floor(two lines below), but we have no better knowledge of whether it's a floor or ceiling, or if user only specifies one of them. Let me know if any better thought.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -2871,9 +2878,9 @@
...
+      <sch:assert role='ERROR' test='count(h:Walls/h:Wall[h:InteriorAdjacentTo="conditioned space" and h:ExteriorAdjacentTo="outside"]) + count(h:FoundationWalls/h:FoundationWall[h:InteriorAdjacentTo="conditioned space" and h:ExteriorAdjacentTo="ground"]) + count(h:Walls/h:Wall[h:SystemIdentifier/@sameas]) + count(h:FoundationWalls/h:FoundationWall[h:SystemIdentifier/@sameas]) &gt;= 1'>There must be at least one exterior wall or foundation wall adjacent to conditioned space.</sch:assert> <!-- FIXME: Should we relax this restriction for internal zones? -->
```
yzhou601, line 2882: A basement can be a separate Building element, So I added foundation wall here too.

workflow/sample_files/base-bldgtype-mf-whole-building-common-spaces.xml
shorowit, line 1: @yzhou601 thinks that the `sameas` attribute can be on either surface; let's add a second test file that is identical to this but with the sameas surfaces reversed. We should get identical results.

shorowit:
- [ ] Add `referenced_by_sameas=ID` property in `HPXML::get_sameas_obj` method
  - [ ] Add error-checking if multiple references
  - [ ] Add error-checking if sameas reference in same building?
- [ ] Treat any surfaces with sameas (in both directions) as adiabatic for design load calculations
- [ ] Improve code logic and/or variable names and/or add comments in e.g. `apply_walls` and similar?

PR #1853, opened 2024-10-10
Electric panel

joseph-robertson:
## Pull Request Description

Support electric panel.

Related to: https://github.com/hpxmlwg/hpxml/pull/420.

## Checklist

Not all may apply:

### DIFF
--- a/BuildResidentialHPXML/README.md
+++ b/BuildResidentialHPXML/README.md
@@ -4408,6 +4408,76 @@ Maximum power output of the second PV system. For a shared system, this is the t
 
 <br/>
 
+**Electric Panel: Service Voltage**
+
+TODO
+
+- **Name:** ``electric_panel_service_voltage``
+- **Type:** ``Choice``
+
+- **Required:** ``false``
+
+- **Choices:** `120`, `240`
+
+<br/>
+
+**Electric Panel: Service Rating**
+
+TODO
+
+- **Name:** ``electric_panel_service_rating``
+- **Type:** ``Double``
+
+- **Units:** ``A``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Load Types**
+
+TODO
+
+- **Name:** ``electric_panel_load_types``
+- **Type:** ``String``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Load Watts**
+
+TODO
+
+- **Name:** ``electric_panel_load_watts``
+- **Type:** ``String``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Load Voltages**
+
+TODO
+
+- **Name:** ``electric_panel_load_voltages``
+- **Type:** ``String``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Load Additions**
+
+TODO
+
+- **Name:** ``electric_panel_load_additions``
+- **Type:** ``String``
+
+- **Required:** ``false``
+
+<br/>
+
 **Battery: Present**
 
 Whether there is a lithium ion battery present.
--- a/BuildResidentialHPXML/measure.rb
+++ b/BuildResidentialHPXML/measure.rb
@@ -2621,6 +2621,42 @@ class BuildResidentialHPXML < OpenStudio::Measure::ModelMeasure
     arg.setDefaultValue(4000)
     args << arg
 
+    electric_panel_voltage_choices = OpenStudio::StringVector.new
+    electric_panel_voltage_choices << HPXML::ElectricPanelVoltage120
+    electric_panel_voltage_choices << HPXML::ElectricPanelVoltage240
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('electric_panel_service_voltage', electric_panel_voltage_choices, false)
+    arg.setDisplayName('Electric Panel: Service Voltage')
+    arg.setDescription('TODO')
+    arg.setUnits('V')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('electric_panel_service_rating', false)
+    arg.setDisplayName('Electric Panel: Service Rating')
+    arg.setDescription('TODO')
+    arg.setUnits('A')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeStringArgument('electric_panel_load_types', false)
+    arg.setDisplayName('Electric Panel: Load Types')
+    arg.setDescription('TODO')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeStringArgument('electric_panel_load_watts', false)
+    arg.setDisplayName('Electric Panel: Load Watts')
+    arg.setDescription('TODO')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeStringArgument('electric_panel_load_voltages', false)
+    arg.setDisplayName('Electric Panel: Load Voltages')
+    arg.setDescription('TODO')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument.makeStringArgument('electric_panel_load_additions', false)
+    arg.setDisplayName('Electric Panel: Load Additions')
+    arg.setDescription('TODO')
+    args << arg
+
     battery_location_choices = OpenStudio::StringVector.new
     battery_location_choices << HPXML::LocationConditionedSpace
     battery_location_choices << HPXML::LocationBasementConditioned
@@ -3934,6 +3970,7 @@ module HPXMLFile
     set_water_fixtures(hpxml_bldg, args)
     set_solar_thermal(hpxml_bldg, args, weather)
     set_pv_systems(hpxml_bldg, args, weather)
+    set_electric_panel(hpxml_bldg, args)
     set_battery(hpxml_bldg, args)
     set_lighting(hpxml_bldg, args)
     set_dehumidifier(hpxml_bldg, args)
@@ -6797,6 +6834,44 @@ module HPXMLFile
                              inverter_efficiency: args[:pv_system_inverter_efficiency])
     hpxml_bldg.pv_systems.each do |pv_system|
       pv_system.inverter_idref = hpxml_bldg.inverters[-1].id
+    end
+  end
+
+  # Set the electric panel properties, including:
+  # - service voltage
+  # - max current service rating
+  # - individual panel loads
+  #
+  # @param hpxml_bldg [HPXML::Building] HPXML Building object representing an individual dwelling unit
+  # @param args [Hash] Map of :argument_name => value
+  # @return [nil]
+  def self.set_electric_panel(hpxml_bldg, args)
+    return if args[:electric_panel_service_voltage].nil? && args[:electric_panel_service_rating].nil?
+
+    hpxml_bldg.electric_panels.add(id: "ElectricPanel#{hpxml_bldg.electric_panels.size + 1}",
+                                   voltage: args[:electric_panel_service_voltage],
+                                   max_current_rating: args[:electric_panel_service_rating])
+
+    if not args[:electric_panel_load_types].nil?
+      panel_loads = hpxml_bldg.electric_panels[0].panel_loads
+
+      electric_panel_load_types = args[:electric_panel_load_types].split(',').map(&:strip)
+      electric_panel_load_watts = args[:electric_panel_load_watts].split(',').map(&:strip)
+      electric_panel_load_voltages = args[:electric_panel_load_voltages].split(',').map(&:strip)
+      electric_panel_load_additions = args[:electric_panel_load_additions].split(',').map(&:strip)
+
+      electric_panel_loads = electric_panel_load_types.zip(electric_panel_load_watts,
+                                                           electric_panel_load_voltages,
+                                                           electric_panel_load_additions)
+
+      electric_panel_loads.each do |electric_panel_load|
+        type, watts, voltage, addition = electric_panel_load
+
+        panel_loads.add(type: type,
+                        watts: watts,
+                        voltage: voltage,
+                        addition: addition)
+      end
     end
   end
 
--- a/HPXMLtoOpenStudio/resources/hpxml.rb
+++ b/HPXMLtoOpenStudio/resources/hpxml.rb
@@ -113,6 +113,12 @@ class HPXML < Object
   DuctTypeSupply = 'supply'
   DWHRFacilitiesConnectedAll = 'all'
   DWHRFacilitiesConnectedOne = 'one'
+  ElectricPanelLoadTypeHeating = 'Heating'
+  ElectricPanelLoadTypeCooling = 'Heating'
+  ElectricPanelLoadTypeWaterHeater = 'Water Heater'
+  ElectricPanelLoadTypeClothesDryer = 'Clothes Dryer'
+  ElectricPanelVoltage120 = '120'
+  ElectricPanelVoltage240 = '240'
   ElectricResistanceDistributionRadiantCeiling = 'radiant ceiling'
   ElectricResistanceDistributionRadiantFloor = 'radiant floor'
   ElectricResistanceDistributionBaseboard = 'baseboard'
@@ -1413,6 +1419,7 @@ class HPXML < Object
                    :solar_thermal_systems,         # [HPXML::SolarThermalSystems]
                    :pv_systems,                    # [HPXML::PVSystems]
                    :inverters,                     # [HPXML::Inverters]
+                   :electric_panels,               # [HPXML::ElectricPanels]
                    :batteries,                     # [HPXML::Batteries]
                    :generators,                    # [HPXML::Generators]
                    :clothes_washers,               # [HPXML::ClothesWashers]
@@ -1552,6 +1559,7 @@ class HPXML < Object
       @solar_thermal_systems.to_doc(building)
       @pv_systems.to_doc(building)
       @inverters.to_doc(building)
+      @electric_panels.to_doc(building)
       @batteries.to_doc(building)
       @generators.to_doc(building)
       @clothes_washers.to_doc(building)
@@ -1638,6 +1646,7 @@ class HPXML < Object
       @solar_thermal_systems = SolarThermalSystems.new(self, building)
       @pv_systems = PVSystems.new(self, building)
       @inverters = Inverters.new(self, building)
+      @electric_panels = ElectricPanels.new(self, building)
       @batteries = Batteries.new(self, building)
       @generators = Generators.new(self, building)
       @clothes_washers = ClothesWashers.new(self, building)
@@ -9114,6 +9123,162 @@ class HPXML < Object
 
       @id = HPXML::get_id(inverter)
       @inverter_efficiency = XMLHelper.get_value(inverter, 'InverterEfficiency', :float)
+    end
+  end
+
+  # Array of HPXML::ElectricPanel objects.
+  class ElectricPanels < BaseArrayElement
+    # Adds a new object, with the specified keyword arguments, to the array.
+    #
+    # @return [nil]
+    def add(**kwargs)
+      self << ElectricPanel.new(@parent_object, **kwargs)
+    end
+
+    # Populates the HPXML object(s) from the XML document.
+    #
+    # @param building [Oga::XML::Element] The current Building XML element
+    # @return [nil]
+    def from_doc(building)
+      return if building.nil?
+
+      XMLHelper.get_elements(building, 'BuildingDetails/Systems/ElectricPanels/ElectricPanel').each do |electric_panel|
+        self << ElectricPanel.new(@parent_object, electric_panel)
+      end
+    end
+  end
+
+  # Object for /HPXML/Building/BuildingDetails/Systems/ElectricPanels/ElectricPanel.
+  class ElectricPanel < BaseElement
+    def initialize(hpxml_element, *args, **kwargs)
+      @panel_loads = PanelLoads.new(hpxml_element)
+      super(hpxml_element, *args, **kwargs)
+    end
+    CLASS_ATTRS = [:panel_loads]
+    ATTRS = [:id, # [String] SystemIdentifier/@id
+             :voltage,
+             :max_current_rating]
+    attr_reader(*CLASS_ATTRS)
+    attr_accessor(*ATTRS)
+
+    # Deletes the current object from the array.
+    #
+    # @return [nil]
+    def delete
+      @parent_object.electric_panels.delete(self)
+    end
+
+    # Additional error-checking beyond what's checked in Schema/Schematron validators.
+    #
+    # @return [Array<String>] List of error messages
+    def check_for_errors
+      errors = []
+      return errors
+    end
+
+    # Adds this object to the provided Oga XML element.
+    #
+    # @param building [Oga::XML::Element] The current Building XML element
+    # @return [nil]
+    def to_doc(building)
+      return if nil?
+
+      electric_panels = XMLHelper.create_elements_as_needed(building, ['BuildingDetails', 'Systems', 'ElectricPanels'])
+      electric_panel = XMLHelper.add_element(electric_panels, 'ElectricPanel')
+      sys_id = XMLHelper.add_element(electric_panel, 'SystemIdentifier')
+      XMLHelper.add_attribute(sys_id, 'id', @id)
+      XMLHelper.add_element(electric_panel, 'Voltage', @voltage, :string, @is_voltage_defaulted) unless @voltage.nil?
+      XMLHelper.add_element(electric_panel, 'MaxCurrentRating', @max_current_rating, :float, @max_current_rating_defaulted) unless @max_current_rating.nil?
+      @panel_loads.to_doc(electric_panel)
+    end
+
+    # Populates the HPXML object(s) from the XML document.
+    #
+    # @param electric_panel [Oga::XML::Element] The current ElectricPanel XML element
+    # @return [nil]
+    def from_doc(electric_panel)
+      return if electric_panel.nil?
+
+      @id = HPXML::get_id(electric_panel)
+      @voltage = XMLHelper.get_value(electric_panel, 'Voltage', :string)
+      @max_current_rating = XMLHelper.get_value(electric_panel, 'MaxCurrentRating', :float)
+      @panel_loads.from_doc(electric_panel)
+    end
+  end
+
+  # Array of HPXML::PanelLoad objects.
+  class PanelLoads < BaseArrayElement
+    # Adds a new object, with the specified keyword arguments, to the array.
+    #
+    # @return [nil]
+    def add(**kwargs)
+      self << PanelLoad.new(@parent_object, **kwargs)
+    end
+
+    # Populates the HPXML object(s) from the XML document.
+    #
+    # @param electric_panel [Oga::XML::Element] The current Electric panel XML element
+    # @return [nil]
+    def from_doc(electric_panel)
+      return if electric_panel.nil?
+
+      XMLHelper.get_elements(electric_panel, 'extension/PanelLoads/PanelLoad').each do |panel_load|
+        self << PanelLoad.new(@parent_object, panel_load)
+      end
+    end
+  end
+
+  # Object for /HPXML/Building/BuildingDetails/Systems/ElectricPanels/ElectricPanel/extension/PanelLoads/PanelLoad.
+  class PanelLoad < BaseElement
+    ATTRS = [:type,
+             :watts,
+             :voltage,
+             :addition]
+    attr_accessor(*ATTRS)
+
+    # Deletes the current object from the array.
+    #
+    # @return [nil]
+    def delete
+      @parent_object.electric_panels.each do |electric_panel|
+        electric_panel.panel_loads.delete(self)
+      end
+    end
+
+    # Additional error-checking beyond what's checked in Schema/Schematron validators.
+    #
+    # @return [Array<String>] List of error messages
+    def check_for_errors
+      errors = []
+      return errors
+    end
+
+    # Adds this object to the provided Oga XML element.
+    #
+    # @param building [Oga::XML::Element] The current Building XML element
+    # @return [nil]
+    def to_doc(electric_panel)
+      return if nil?
+
+      panel_loads = XMLHelper.create_elements_as_needed(electric_panel, ['extension', 'PanelLoads'])
+      panel_load = XMLHelper.add_element(panel_loads, 'PanelLoad')
+      XMLHelper.add_element(panel_load, 'Type', @type, :string, @is_type_defaulted) unless @type.nil?
+      XMLHelper.add_element(panel_load, 'Watts', @watts, :float, @is_watts_defaulted) unless @watts.nil?
+      XMLHelper.add_element(panel_load, 'Voltage', @voltage, :string, @is_voltage_defaulted) unless @voltage.nil?
+      XMLHelper.add_element(panel_load, 'Addition', @addition, :boolean, @is_addition_defaulted) unless @addition.nil?
+    end
+
+    # Populates the HPXML object(s) from the XML document.
+    #
+    # @param panel_load [Oga::XML::Element] The current PanelLoad XML element
+    # @return [nil]
+    def from_doc(panel_load)
+      return if panel_load.nil?
+
+      @type = XMLHelper.get_value(panel_load, 'Type', :string)
+      @watts = XMLHelper.get_value(panel_load, 'Watts', :float)
+      @voltage = XMLHelper.get_value(panel_load, 'Voltage', :string)
+      @addition = XMLHelper.get_value(panel_load, 'Addition', :boolean)
     end
   end
 
--- a/docs/source/workflow_inputs.rst
+++ b/docs/source/workflow_inputs.rst
@@ -4544,6 +4544,44 @@ In addition, the PVSystem must be connected to an inverter that is entered as a 
 
   .. [#] For homes with multiple inverters, all InverterEfficiency elements must have the same value.
 
+.. _hpxml_electric_panels:
+
+HPXML Electric Panels
+*********************
+
+A single electric panel can be entered as a ``/HPXML/Building/BuildingDetails/Systems/ElectricPanels/ElectricPanel``.
+
+  ====================================================  =======  =========  =======================  ========  ========  ============================================
+  Element                                               Type     Units      Constraints              Required  Default   Notes
+  ====================================================  =======  =========  =======================  ========  ========  ============================================
+  ``SystemIdentifier``                                  id                                           Yes                 Unique identifier
+  ``Voltage``                                           string              See [#]_                 No        240?
+  ``MaxCurrentRating``                                  double                                       No        ?
+  ``PanelLoads``                                        element                                      No        <none>    Individual electric panel loads [#]_
+  ====================================================  =======  =========  =======================  ========  ========  ============================================
+
+  .. [#] Voltage choices are "120" or "240".
+  .. [#] If PanelLoads is provided, see :ref:`panel_loads`.
+
+.. _panel_loads:
+
+Panel Loads
+~~~~~~~~~~~
+
+TODO entered in ``extension/PanelLoads/PanelLoad``.
+
+  ==============================================  ========  ==============  ===========  ========  =========  ==========================================
+  Element                                         Type      Units           Constraints  Required  Default    Notes
+  ==============================================  ========  ==============  ===========  ========  =========  ==========================================
+  ``Type``                                        string                    See [#]_     Yes
+  ``Watts``                                       double    W                            Yes
+  ``Voltage``                                     string    V               See [#]_     No        120?
+  ``Addition``                                    boolean                                No        false
+  ==============================================  ========  ==============  ===========  ========  =========  ==========================================
+
+  .. [#] Type choices are "Heating", "Cooling", "Hot Water", or "Clothes Dryer".
+  .. [#] Voltage choices are "120" or "240".
+
 .. _hpxml_batteries:
 
 HPXML Batteries
--- a/BuildResidentialHPXML/measure.xml
+++ b/BuildResidentialHPXML/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.1</schema_version>
   <name>build_residential_hpxml</name>
   <uid>a13a8983-2b01-4930-8af2-42030b6e4233</uid>
-  <version_id>32809635-5673-4b10-bf0a-de06a038b619</version_id>
-  <version_modified>2024-10-01T18:49:48Z</version_modified>
+  <version_id>1556d435-048f-43a2-a4ac-c1be820c07d0</version_id>
+  <version_modified>2024-10-10T22:43:09Z</version_modified>
   <xml_checksum>2C38F48B</xml_checksum>
   <class_name>BuildResidentialHPXML</class_name>
   <display_name>HPXML Builder</display_name>
@@ -5389,6 +5389,66 @@
       <default_value>4000</default_value>
     </argument>
     <argument>
+      <name>electric_panel_service_voltage</name>
+      <display_name>Electric Panel: Service Voltage</display_name>
+      <description>TODO</description>
+      <type>Choice</type>
+      <units>V</units>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+      <choices>
+        <choice>
+          <value>120</value>
+          <display_name>120</display_name>
+        </choice>
+        <choice>
+          <value>240</value>
+          <display_name>240</display_name>
+        </choice>
+      </choices>
+    </argument>
+    <argument>
+      <name>electric_panel_service_rating</name>
+      <display_name>Electric Panel: Service Rating</display_name>
+      <description>TODO</description>
+      <type>Double</type>
+      <units>A</units>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+    </argument>
+    <argument>
+      <name>electric_panel_load_types</name>
+      <display_name>Electric Panel: Load Types</display_name>
+      <description>TODO</description>
+      <type>String</type>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+    </argument>
+    <argument>
+      <name>electric_panel_load_watts</name>
+      <display_name>Electric Panel: Load Watts</display_name>
+      <description>TODO</description>
+      <type>String</type>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+    </argument>
+    <argument>
+      <name>electric_panel_load_voltages</name>
+      <display_name>Electric Panel: Load Voltages</display_name>
+      <description>TODO</description>
+      <type>String</type>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+    </argument>
+    <argument>
+      <name>electric_panel_load_additions</name>
+      <display_name>Electric Panel: Load Additions</display_name>
+      <description>TODO</description>
+      <type>String</type>
+      <required>false</required>
+      <model_dependent>false</model_dependent>
+    </argument>
+    <argument>
       <name>battery_present</name>
       <display_name>Battery: Present</display_name>
       <description>Whether there is a lithium ion battery present.</description>
@@ -7518,7 +7578,7 @@
       <filename>README.md</filename>
       <filetype>md</filetype>
       <usage_type>readme</usage_type>
-      <checksum>9F5858F9</checksum>
+      <checksum>7E62D681</checksum>
     </file>
     <file>
       <filename>README.md.erb</filename>
@@ -7535,7 +7595,7 @@
       <filename>measure.rb</filename>
       <filetype>rb</filetype>
       <usage_type>script</usage_type>
-      <checksum>9F12F6CA</checksum>
+      <checksum>BE34498D</checksum>
     </file>
     <file>
       <filename>constants.rb</filename>
--- a/HPXMLtoOpenStudio/measure.xml
+++ b/HPXMLtoOpenStudio/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.1</schema_version>
   <name>hpxm_lto_openstudio</name>
   <uid>b1543b30-9465-45ff-ba04-1d1f85e763bc</uid>
-  <version_id>1424b04d-aae3-4360-9d50-3324d0babb33</version_id>
-  <version_modified>2024-10-05T18:13:33Z</version_modified>
+  <version_id>b81f661f-bed0-4e7e-af3d-7fe1ac2b4cd4</version_id>
+  <version_modified>2024-10-10T22:43:11Z</version_modified>
   <xml_checksum>D8922A73</xml_checksum>
   <class_name>HPXMLtoOpenStudio</class_name>
   <display_name>HPXML to OpenStudio Translator</display_name>
@@ -357,7 +357,7 @@
       <filename>hpxml.rb</filename>
       <filetype>rb</filetype>
       <usage_type>resource</usage_type>
-      <checksum>278B0E8F</checksum>
+      <checksum>076234E4</checksum>
     </file>
     <file>
       <filename>hpxml_schema/HPXML.xsd</filename>
@@ -375,7 +375,7 @@
       <filename>hpxml_schematron/EPvalidator.xml</filename>
       <filetype>xml</filetype>
       <usage_type>resource</usage_type>
-      <checksum>9A5ABED7</checksum>
+      <checksum>223546C9</checksum>
     </file>
     <file>
       <filename>hpxml_schematron/iso-schematron.xsd</filename>
@@ -646,12 +646,6 @@
       <filetype>rb</filetype>
       <usage_type>resource</usage_type>
       <checksum>93120E27</checksum>
-    </file>
-    <file>
-      <filename>in.schedules.csv</filename>
-      <filetype>csv</filetype>
-      <usage_type>test</usage_type>
-      <checksum>C8EBFF38</checksum>
     </file>
     <file>
       <filename>test_airflow.rb</filename>
--- a/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
+++ b/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
@@ -264,6 +264,7 @@
       <sch:assert role='ERROR' test='count(h:Systems/h:WaterHeating/h:WaterFixture) &gt;= 0'>Expected 0 or more element(s) for xpath: Systems/WaterHeating/WaterFixture</sch:assert> <!-- See [WaterFixture] -->
       <sch:assert role='ERROR' test='count(h:Systems/h:SolarThermal/h:SolarThermalSystem) &lt;= 1'>Expected 0 or 1 element(s) for xpath: Systems/SolarThermal/SolarThermalSystem</sch:assert> <!-- See [SolarThermalSystem] -->
       <sch:assert role='ERROR' test='count(h:Systems/h:Photovoltaics/h:PVSystem) &gt;= 0'>Expected 0 or more element(s) for xpath: Systems/Photovoltaics/PVSystem</sch:assert> <!-- See [PVSystem] -->
+      <sch:assert role='ERROR' test='count(h:Systems/h:ElectricPanels/h:ElectricPanel) &gt;= 0'>Expected 0 or more element(s) for xpath: Systems/ElectricPanels/ElectricPanel</sch:assert> <!-- See [ElectricPanel] -->
       <sch:assert role='ERROR' test='count(h:Systems/h:Batteries/h:Battery) &lt;= 1'>Expected 0 or 1 element(s) for xpath: Systems/Batteries/Battery</sch:assert> <!-- See [Battery] -->
       <sch:assert role='ERROR' test='count(h:Systems/h:extension/h:Generators/h:Generator) &gt;= 0'>Expected 0 or more element(s) for xpath: Systems/extension/Generators/Generator</sch:assert> <!-- See [Generator] -->
       <sch:assert role='ERROR' test='count(h:Appliances/h:ClothesWasher) &lt;= 1'>Expected 0 or 1 element(s) for xpath: Appliances/ClothesWasher</sch:assert> <!-- See [ClothesWasher] -->
@@ -2398,6 +2399,14 @@
     <sch:title>[Inverter]</sch:title>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:Photovoltaics/h:Inverter'>
       <sch:assert role='ERROR' test='count(h:InverterEfficiency) &lt;= 1'>Expected 0 or 1 element(s) for xpath: InverterEfficiency</sch:assert>
+    </sch:rule>
+  </sch:pattern>
+
+  <sch:pattern>
+    <sch:title>[ElectricPanel]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:ElectricPanels/h:ElectricPanel'>
+      <sch:assert role='ERROR' test='count(h:Voltage) = 1'>Expected 1 element(s) for xpath: Voltage</sch:assert>
+      <sch:assert role='ERROR' test='count(h:MaxCurrentRating) = 1'>Expected 1 element(s) for xpath: MaxCurrentRating</sch:assert>
     </sch:rule>
   </sch:pattern>
 
--- a/ReportUtilityBills/measure.xml
+++ b/ReportUtilityBills/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.1</schema_version>
   <name>report_utility_bills</name>
   <uid>ca88a425-e59a-4bc4-af51-c7e7d1e960fe</uid>
-  <version_id>105d1f2f-ea3f-4225-b340-4ea362671b8a</version_id>
-  <version_modified>2024-10-01T18:47:20Z</version_modified>
+  <version_id>31f263cc-5708-4cd6-a18c-58c432d19f87</version_id>
+  <version_modified>2024-10-10T22:43:14Z</version_modified>
   <xml_checksum>15BF4E57</xml_checksum>
   <class_name>ReportUtilityBills</class_name>
   <display_name>Utility Bills Report</display_name>
@@ -181,12 +181,6 @@
       <filetype>rb</filetype>
       <usage_type>script</usage_type>
       <checksum>CC656203</checksum>
-    </file>
-    <file>
-      <filename>detailed_rates/Adams Electric Cooperative Inc - Rate Schedule T1 TOD (Effective 2013-02-01).json</filename>
-      <filetype>json</filetype>
-      <usage_type>resource</usage_type>
-      <checksum>8E644347</checksum>
     </file>
     <file>
       <filename>detailed_rates/README.md</filename>
--- a/docs/source/workflow_outputs.rst
+++ b/docs/source/workflow_outputs.rst
@@ -424,6 +424,38 @@ Resilience outputs are listed below.
          The entire electric load is treated as a "critical load" that would be supported during an outage.
          Resilience hours are set to 0 for any timestep where the battery is not charged, even if there is sufficient PV to power the building.
 
+Panel Capacities
+~~~~~~~~~~~~~~~~
+
+Panel outputs are listed below.
+
+  ====================================================  ====================
+  Type                                                  Notes
+  ====================================================  ====================
+  Electric Panel: Load: Heating (W)
+  Electric Panel: Load: Cooling (W)
+  Electric Panel: Load: Hot Water (W)
+  Electric Panel: Load: Clothes Dryer (W)
+  Electric Panel: Load: Dishwasher (W)
+  Electric Panel: Load: Range/Oven (W)
+  Electric Panel: Load: Permanent Spa Heater (W)
+  Electric Panel: Load: Permanent Spa Pump (W)
+  Electric Panel: Load: Pool Heater (W)
+  Electric Panel: Load: Pool Pump (W)
+  Electric Panel: Load: Well Pump (W)
+  Electric Panel: Load: Electric Vehicle Charging (W)
+  Electric Panel: Load: Lighting (W)
+  Electric Panel: Load: Other (W)
+  Electric Panel: Capacity: Load-Based (W)              220.83
+  Electric Panel: Capacity: Load-Based (A)              220.83
+  Electric Panel: Constraint: Load-Based (A)            220.83
+  Electric Panel: Capacity: Meter-Based (W)             220.87
+  Electric Panel: Capacity: Meter-Based (A)             220.87
+  Electric Panel: Constraint: Meter-Based (A)           220.87
+  Electric Panel: Breaker Space: HVAC (#)
+  Electric Panel: Breaker Space: Whole Panel (#)
+  ====================================================  ====================
+
 HVAC Capacities
 ~~~~~~~~~~~~~~~
 
--- a/workflow/hpxml_inputs.json
+++ b/workflow/hpxml_inputs.json
@@ -1601,6 +1601,15 @@
     "water_heater_setpoint_temperature": null,
     "schedules_filepaths": "../../HPXMLtoOpenStudio/resources/schedule_files/water-heater-setpoints.csv"
   },
+  "sample_files/base-electric-panel.xml": {
+    "parent_hpxml": "sample_files/base.xml",
+    "electric_panel_service_voltage": "240",
+    "electric_panel_service_rating": 100,
+    "electric_panel_load_types": "Heating, Cooling, Hot Water, Clothes Dryer",
+    "electric_panel_load_watts": "17943, 16000, 4500, 5760",
+    "electric_panel_load_voltages": "240, 240, 120, 240",
+    "electric_panel_load_additions": "true, true, false, false"
+  },
   "sample_files/base-enclosure-2stories.xml": {
     "parent_hpxml": "sample_files/base.xml",
     "geometry_unit_num_floors_above_grade": 2,
--- a/workflow/sample_files/base-electric-panel.xml
+++ b/workflow/sample_files/base-electric-panel.xml
@@ -1,0 +1,586 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<HPXML xmlns='http://hpxmlonline.com/2023/09' schemaVersion='4.0'>
+  <XMLTransactionHeaderInformation>
+    <XMLType>HPXML</XMLType>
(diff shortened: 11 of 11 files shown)


docs/source/workflow_inputs.rst
```diff
@@ -4542,6 +4542,79 @@ In addition, the PVSystem must be connected to an inverter that is entered as a
...
+
+A single electric panel can be entered as a ``/HPXML/Building/BuildingDetails/Systems/ElectricPanels/ElectricPanel``.
+
+  ====================================================  =======  =========  =======================  ========  ========  ============================================
+  Element                                               Type     Units      Constraints              Required  Default   Notes
+  ====================================================  =======  =========  =======================  ========  ========  ============================================
+  ``SystemIdentifier``                                  id                                           Yes                 Unique identifier
+  ``Voltage``                                           string              See [#]_                 No        240
+  ``MaxCurrentRating``                                  double                                       No        150
+  ``PanelLoads``                                        element                                      No        See [#]_  Individual electric panel loads [#]_
```
shorowit, line 4558: Should be `extension/PanelLoads`

BuildResidentialHPXML/measure.rb
```diff
@@ -2627,6 +2627,42 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
+    electric_panel_voltage_choices << HPXML::ElectricPanelVoltage120
+    electric_panel_voltage_choices << HPXML::ElectricPanelVoltage240
+
+    arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('electric_panel_service_voltage', electric_panel_voltage_choices, false)
+    arg.setDisplayName('Electric Panel: Service Voltage')
+    arg.setDescription("The service voltage of the electric panel. If not provided, the OS-HPXML default (see <a href='#{docs_base_url}#hpxml-electric-panels'>HPXML Electric Panels</a>) is used.")
+    arg.setUnits('V')
+    args << arg
+
+    arg = OpenStudio::Measure::OSArgument::makeDoubleArgument('electric_panel_service_rating', false)
```
joseph-robertson, line 2640: We'd likely have an additional argument, something like `electric_panel_service_breaker_spaces_remaining`.

jmaguire1:
@joseph-robertson: For HPWHs: yes, you can assume that there's not simultaneous operation of the elements and the HP, and just use the max value of the HP power and the backup power (which since it's an ER element is the same as the backup capacity).

120V HPWHs CAN have backup, but they also have to comply with the power limits of being 120V, so they're pretty small (~900W). AO Smith has backup elements; Rheem does not. The real advantage of having backup elements isn't that you can reheat faster, which is why 240V HPWHs have elements, but that it can run when the HP can't. So for example if you install in a garage in a warm climate, but a winter storm moves in and your garage is below the minimum temperature where the HP can run (~40 F), you can still get hot water. It's really mostly about making sure these units can still operate during rare cold events when installed in unconditioned spaces in warm climates.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5570,6 +5800,346 @@ def self.get_ceiling_fan_months(weather)
...
+
+        watts += HVAC.get_dx_cooling_coil_power_watts_from_capacity(UnitConversions.convert(heat_pump.cooling_capacity, 'btu/hr', 'kbtu/hr'))
+        watts += HVAC.get_blower_fan_power_watts(heat_pump.fan_watts_per_cfm, heat_pump.cooling_airflow_cfm)
+
+        if heat_pump.fraction_heat_load_served == 0
+          breaker_spaces += 3 # ODU; the 3 we missed adding to heating
+        end
+      end
+
+    elsif type == HPXML::ElectricPanelLoadTypeWaterHeater
```
joseph-robertson, line 5980: Can we move these assignments to a table?

JLReyna: Which of the following were you thinking of for a table? All of these?
- ElectricPanelLoadTypeWaterHeater
- ElectricPanelLoadTypeClothesDryer
- ElectricPanelLoadTypeDishwasher
- ElectricPanelLoadTypeRangeOven
- ElectricPanelLoadTypeMechVent
- ElectricPanelLoadTypePermanentSpaPump
- ElectricPanelLoadTypePermanentSpaHeater
- ElectricPanelLoadTypePoolHeater
- ElectricPanelLoadTypePoolPump
- ElectricPanelLoadTypeWellPump
- ElectricPanelLoadTypeElectricVehicleCharging
- ElectricPanelLoadTypeLighting (i.e., the floor area multiplier?)
- ElectricPanelLoadTypeKitchen
- ElectricPanelLoadTypeLaundry
- ElectricPanelLoadTypeOther

What would be the easiest format to read in, considering some of the apply logic needed?

joseph-robertson: Are you using "apply logic" to mean that basically each of the above categories is somewhat unique in that power ratings are assigned based on different properties? For example, water heater uses number of bathrooms but well pump uses number of bedrooms?

So a flat table would be pretty sparse. Yeah, I'm not really sure what the best format is. Maybe table (or similar) isn't the best approach -- I was just trying to think of a way these power rating values could be easily updated/maintained without having to read through a bunch of lines of code.

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -993,6 +993,94 @@ def self.get_total_hvac_capacities(hpxml_bldg)
     return htg_cap, clg_cap, hp_backup_cap
   end
 
+  # Calculates total panel loads (across all panel loads for a given panel load type) for a given HPXML Building.
+  #
+  # @param hpxml_bldg [HPXML::Building] HPXML Building object representing an individual dwelling unit
+  # @return [Array<Double * 15>] Total panel load for each panel load type (W)
+  def self.get_total_panel_loads(hpxml_bldg)
```
lixiliu, line 1000: We should revisit that. Service entrance for building has a different set of demand factors that needs to be applied for load coincidence, in addition to unit_multiplier. Let's discuss.

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -993,6 +993,94 @@ def self.get_total_hvac_capacities(hpxml_bldg)
...
+        elsif panel_load.type == HPXML::ElectricPanelLoadTypeOther
+          oth += panel_load.power * unit_multiplier
+        end
+      end
+    end
+    return htg, clg, hw, cd, dw, ov, vf, sh, sp, ph, pp, wp, ev, ltg, oth
+  end
+
+  # TODO
+  def self.get_total_breaker_spaces(hpxml_bldg)
```
lixiliu, line 1041: This is not meaningful, because each apartment will have their own unit panel that we must evaluate on independently. Unit panels are wired to the service main panel of the building along with other shared systems but I am not sure about the exact details.

Let's discuss.

HPXMLtoOpenStudio/resources/electric_panel.rb
```diff
@@ -0,0 +1,237 @@
...
+  # @return [nil]
+  def self.calculate_load_based(hpxml_bldg, electric_panel, panel_loads, panel_calculation_type)
+    if panel_calculation_type == HPXML::ElectricPanelLoadCalculationType2023LoadBased
+      htg_existing = get_panel_load_heating(hpxml_bldg, electric_panel, addition: false)
+      htg_new = get_panel_load_heating(hpxml_bldg, electric_panel, addition: true)
+      clg_existing = electric_panel.panel_loads.select { |panel_load| panel_load.type == HPXML::ElectricPanelLoadTypeCooling && !panel_load.addition }.map { |pl| pl.power }.sum(0.0)
+      clg_new = electric_panel.panel_loads.select { |panel_load| panel_load.type == HPXML::ElectricPanelLoadTypeCooling && panel_load.addition }.map { |pl| pl.power }.sum(0.0)
+
+      # Part A
+      other_load = [htg_existing, clg_existing].max
```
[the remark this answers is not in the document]
lixiliu: Additionally, we need to special handle 120V dryer.

Currently we assign one 1.5kVA (20A) circuit for laundry, which will support the washer (requires 12A on avg). We should check that if the laundry circuit can accommodate the 120V dryer, we remove that load from panel_loads. I think this will almost always going to be no, but we should add that adjustment.

Would we do the adjustment here?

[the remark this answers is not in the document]
lixiliu: BTW, where is the kitchen and laundry circuit VA specified in the code?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+          watts += 1119
+        end
+        breaker_spaces += 2
+      end
+    elsif type == HPXML::ElectricPanelLoadTypeElectricVehicleCharging
+      hpxml_bldg.plug_loads.each do |plug_load|
+        next if plug_load.plug_load_type != HPXML::PlugLoadTypeElectricVehicleCharging
+        next if !system_ids.include?(plug_load.id)
+
+        # FIXME: next if MF?
```
JLReyna, line 6104: @lixiliu  - I think we can leave this as-is (i.e., include MF EVs in the panel calculation if there is an EV charging plug load). Do you agree?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+          breaker_spaces += 1
+        end
+        breaker_spaces += 1
+      end
+    elsif type == HPXML::ElectricPanelLoadTypeLighting
+      watts += 3 * hpxml_bldg.building_construction.conditioned_floor_area
+      breaker_spaces += 0
+    elsif type == HPXML::ElectricPanelLoadTypeKitchen
+      watts += 3000
+      breaker_spaces += 0
```
JLReyna, line 6119: shouldn't breaker spaces be 2 for the two kitchen 1.5 kVA small appliance circuits?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+        next if ![HPXML::HeaterTypeElectricResistance, HPXML::HeaterTypeHeatPump].include?(pool.heater_type) # FIXME: probably want to separate out HeatPump here
+
+        watts += 27000
+        breaker_spaces += 2
+      end
+    elsif type == HPXML::ElectricPanelLoadTypePoolPump
+      hpxml_bldg.pools.each do |pool|
+        next if !system_ids.include?(pool.pump_id)
+
+        watts += 1491
```
JLReyna, line 6084: @lixiliu - this has all pool pumps receiving the 2 HP pool pump capacity. Open question if we want to retag / update the actual pool pumps in ResStock to get the labeling and energy consumption to match.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+        next if !system_ids.include?(permanent_spa.pump_id)
+
+        watts += 1491
+        breaker_spaces += 1
+      end
+    elsif type == HPXML::ElectricPanelLoadTypePoolHeater
+      hpxml_bldg.pools.each do |pool|
+        next if !system_ids.include?(pool.heater_id)
+        next if ![HPXML::HeaterTypeElectricResistance, HPXML::HeaterTypeHeatPump].include?(pool.heater_type) # FIXME: probably want to separate out HeatPump here
+
```
JLReyna, line 6076: Yes, let's break out heat pump pool heaters since they're a lot lower. Recommended default value 5100 W, still 240V

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+          watts += 3000 # FIXME: base-mechvent-cfis-no-additional-runtime.xml, e.g., has no FanPower defaulted
+        end
+      end
+      breaker_spaces += 1 # intentionally outside the ventilation_fans loop
+    elsif type == HPXML::ElectricPanelLoadTypePermanentSpaHeater
+      hpxml_bldg.permanent_spas.each do |permanent_spa|
+        next if !system_ids.include?(permanent_spa.heater_id)
+        next if ![HPXML::HeaterTypeElectricResistance, HPXML::HeaterTypeHeatPump].include?(permanent_spa.heater_type)
+
+        watts += 1000 # FIXME
```
JLReyna, line 6062: 1000 W is what was in the [spreadsheet](https://github.com/NREL/resstock/blob/electrical_panels/Electrical_Panels_EUSS/panels/nameplate_rating_new_load.csv), but I think it's low (maybe why there is a FIXME?). I think 4000 W is a more reasonable value (e.g., [this example](https://www.homedepot.com/p/Lifesmart-LS700DX-7-Person-90-Jet-230V-Standard-Spa-with-Waterfall-401431510600/312535665). @lixiliu  do you agree?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+        watts += 1200
+        breaker_spaces += 1
+      end
+    elsif type == HPXML::ElectricPanelLoadTypeRangeOven
+      hpxml_bldg.cooking_ranges.each do |cooking_range|
+        next if !system_ids.include?(cooking_range.id)
+        next if cooking_range.fuel_type != HPXML::FuelTypeElectricity
+
+        if voltage == 120
+          watts += 1800
```
JLReyna, line 6036: Is 120V being assigned to ranges based on the `IsInduction` boolean?

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+
+<br/>
+
+**Electric Panel: Service Voltage**
+
+The service voltage of the electric panel. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#hpxml-electric-panels'>HPXML Electric Panels</a>) is used.
+
+- **Name:** ``electric_panel_service_voltage``
+- **Type:** ``Choice``
+
```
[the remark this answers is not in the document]
joseph-robertson: Hm, I've set units in the measure.rb. I'm not sure why they don't render in the README.md. Looks like perhaps units don't render for choice arguments. Other examples include `foundation_wall_insulation_location`, `hvac_perf_data_capacity_type`, and `geothermal_loop_pipe_diameter`.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
 
+**Electric Panel: Load Calculation Types**
+
+Types of electric panel load calculations. If multiple types, use a comma-separated list. If not provided, no electric panel loads are calculated.
+
+- **Name:** ``electric_panel_load_calculation_types``
+- **Type:** ``String``
+
+- **Required:** ``false``
+
```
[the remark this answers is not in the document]
joseph-robertson: Added "Possible types are: 2023 Load-Based, 2023 Meter-Based." to argument description for `electric_panel_load_calculation_types`.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+- **Units:** ``W``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Heating System Addition**
+
+Specifies whether the panel load heating sysem is an addition. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#panel-loads'>Panel Loads</a>) is used.
```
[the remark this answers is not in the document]
joseph-robertson: I went with "Whether the heating system is a new panel load addition to an existing service panel." This follows existing syntax (e.g., "heating system", "cooling system", "heat pump") for arguments.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+
+- **Units:** ``#``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Heating System Power**
+
+Specifies the panel load heating system power. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#panel-loads'>Panel Loads</a>) is used.
```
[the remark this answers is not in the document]
joseph-robertson: See above.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+Specifies whether the panel load well pump is an addition. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#panel-loads'>Panel Loads</a>) is used.
+
+- **Name:** ``electric_panel_load_misc_plug_loads_well_pump_addition``
+- **Type:** ``Boolean``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Misc Plug Loads Vehicle Power**
```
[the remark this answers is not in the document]
joseph-robertson: I had basically the same question. I've currently got `electric_panel_load_xxx_voltage` arguments for water heater, clothes dryer, cooking range, and electric vehicle. Should those arguments be defined alongside the "electric panel" arguments, or with the existing corresponding category?

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+- **Name:** ``electric_panel_load_heat_pump_cooling_power``
+- **Type:** ``Double``
+
+- **Units:** ``W``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Heat Pump Addition**
```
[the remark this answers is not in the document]
joseph-robertson: For HVAC? Really?

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+**Electric Panel: Breaker Spaces Type**
+
+The breaker spaces specification type of the electric panel. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#hpxml-electric-panels'>HPXML Electric Panels</a>) is used.
+
+- **Name:** ``electric_panel_breaker_spaces_type``
+- **Type:** ``Choice``
+
+- **Required:** ``false``
+
+- **Choices:** `total`, `headroom`
```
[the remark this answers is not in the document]
joseph-robertson: I'm confused here. For ResStock, don't we use "headroom" in BuildExistingModel and "total" in ApplyUpgrade?

BuildResidentialHPXML/measure.rb
```diff
@@ -6844,6 +7136,185 @@ def self.set_pv_systems(hpxml_bldg, args, weather)
...
+                                   voltage: args[:electric_panel_service_voltage],
+                                   max_current_rating: args[:electric_panel_service_rating],
+                                   headroom_breaker_spaces: headroom_breaker_spaces,
+                                   total_breaker_spaces: total_breaker_spaces)
+
+    panel_loads = hpxml_bldg.electric_panels[0].panel_loads
+
+    hpxml_bldg.heating_systems.each do |heating_system|
+      if heating_system.primary_system
+        panel_loads.add(type: HPXML::ElectricPanelLoadTypeHeating,
```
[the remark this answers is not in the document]
joseph-robertson: Yes. If power isn't assigned using BuildResidentialHPXML, then downstream defaulting would assign a power rating corresponding to blower fan or pump.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+- **Name:** ``electric_panel_load_bathroom_fans_addition``
+- **Type:** ``Boolean``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Water Heater Power**
+
+Specifies the panel load water heater power. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#panel-loads'>Panel Loads</a>) is used.
```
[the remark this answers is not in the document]
joseph-robertson: I thought this was sort of implied given the "Electric Panel" category. For example, "Misc Plug Loads" implies electric for well pump.

BuildResidentialHPXML/README.md
```diff
@@ -4421,6 +4421,614 @@ Maximum power output of the second PV system. For a shared system, this is the t
...
+- **Name:** ``electric_panel_load_permanent_spa_heater_addition``
+- **Type:** ``Boolean``
+
+- **Required:** ``false``
+
+<br/>
+
+**Electric Panel: Other Power**
+
+Specifies the panel load other power. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#panel-loads'>Panel Loads</a>) is used.
```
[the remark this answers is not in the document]
joseph-robertson: Added this to the argument description.

HPXMLtoOpenStudio/resources/defaults.rb
joseph-robertson: For both lighting and kitchen, I didn't see these as nonzero in the work plan document.

HPXMLtoOpenStudio/resources/defaults.rb
joseph-robertson: I think we'd need to make some changes/updates on OS-HPXML's end since it appears the BuildResidentialHPXML measure _always_ writes an "unknown" pool pump type: https://github.com/NREL/OpenStudio-HPXML/blob/master/BuildResidentialHPXML/measure.rb#L7345.

HPXMLtoOpenStudio/resources/defaults.rb
joseph-robertson: I wrote the FIXME as a reminder to distinguish between electric resistance and heat pump spa heaters. 4000 W should apply to both types?

HPXMLtoOpenStudio/resources/defaults.rb
joseph-robertson: Same for spa pumps.

HPXMLtoOpenStudio/resources/defaults.rb
joseph-robertson: No. I'm defaulting all electric ranges to 240V. If instead specified as 120V, 1800 W gets applied (induction or not). This seems to be what https://github.com/NREL/resstock/blob/electrical_panels/Electrical_Panels_EUSS/panels/nameplate_rating_new_load.csv implies, no?

BuildResidentialHPXML/README.md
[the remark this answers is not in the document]
JLReyna: It might seem obvious to us, but I think we should specify electric-only to the description for all end uses that can use multiple fuels.

HPXMLtoOpenStudio/resources/electric_panel.rb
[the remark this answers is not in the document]
JLReyna: @lixiliu  - kitchen and laundry VA are specified in the [`defaults.rb`](https://github.com/NREL/OpenStudio-HPXML/blob/3cf18f50162ae6e5be6a575371e9321613433a4c/HPXMLtoOpenStudio/resources/defaults.rb#L3350-L3358) file

HPXMLtoOpenStudio/resources/electric_panel.rb
```diff
@@ -0,0 +1,237 @@
...
+  # @param hpxml_bldg [HPXML::Building] HPXML Building object representing an individual dwelling unit
+  # @param electric_panel [HPXML::ElectricPanel] Object that defines a single electric panel
+  # @param peak_fuels [Hash] Map of peak building electricity outputs
+  # @return [Array<Double, Double, Double>] The capacity (W), the capacity (A), and headroom (A)
+  def self.calculate_meter_based(hpxml_bldg, electric_panel, peak_fuels, panel_calculation_type)
+    if panel_calculation_type == HPXML::ElectricPanelLoadCalculationType2023MeterBased
+      htg_new = get_panel_load_heating(hpxml_bldg, electric_panel, addition: true)
+      clg_new = electric_panel.panel_loads.select { |panel_load| panel_load.type == HPXML::ElectricPanelLoadTypeCooling && panel_load.addition }.map { |pl| pl.power }.sum(0.0)
+
+      new_loads = [htg_new, clg_new].max
```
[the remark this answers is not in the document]
JLReyna: @lixiliu - I think the latter (i.e., max(CAC, ER furnace)) makes the most sense to me, but you understand the code more deeply than me. Would most interpreters literally only include the added CAC, or would all of "HVAC" be considered new, so you'd take the max between heating and cooling?

HPXMLtoOpenStudio/resources/defaults.rb
JLReyna: @joseph-robertson - I interpreted each of the two kitchen branches as needing a breaker space (looking at the line in Table 1 of the Workplan). 

@lixiliu - can you confirm?

HPXMLtoOpenStudio/resources/defaults.rb
JLReyna: Oh that's interesting that unknown type is always assigned.. I think there's two intertwined concepts here of power rating (and the associated panel capacity implications) and energy consumption. In `options_lookup.tsv` ResStock has 3 different pool pumps of different power ratings (0.75 HP, 1 HP, and National Average). For all of these `pool_pump_annual_kwh` is set to `auto`, but the `pool_pump_usage_multiplier` varies (0.75, 1.0, and 0.075 respectively).

The issue with this is that Lixi has identified that all of our pool pumps in ResStock are undersized. For the purposes of the panels project, she has suggested the capacity default be based on a 2 HP pool pump, which is likely more common in practice. I think this makes sense for this project, and I think it's fine to assign as a default in OS-HPXML. I don't think this is significant to call out as an argument in `options_lookup.tsv`. 

The related issue I'm highlighting is that ResStock seems to have undersized pool pumps, which leads to both an energy issue as well as a labeling discrepancy (i.e., ResStock says it's a 1 HP pool pump, but OS-HPXML assumes the panel capacity is equivalent to a 2 HP pool pump).

@lixiliu - I'd propose we leave everything as-is and open a `resstock-estimation` issue on improving pool pump distributions, highlighting the capacity discrepancy. Wanted your signoff on this though

HPXMLtoOpenStudio/resources/defaults.rb
JLReyna: Yes, I think you're right. Just clarifying that 120V would have to be a direct argument, not automatically assigned for induction since I couldn't find it in the code

HPXMLtoOpenStudio/resources/data/default_panels.csv
```diff
@@ -0,0 +1,29 @@
+Load Name,Voltage,Power Rating,Breaker Spaces,Units,Load Type,Notes
```
JLReyna, line 1: Thanks for expanding this to include load type and breaker spaces. I was on the fence, but I think it makes sense to call them out both for clarity and so that they can be used directly

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5606,6 +5835,486 @@ def self.get_ceiling_fan_months(weather)
...
+      hpxml_bldg.dishwashers.each do |dishwasher|
+        next if !system_ids.include?(dishwasher.id)
+
+        if voltage == HPXML::ElectricPanelVoltage240
+          runner.registerWarning("PanelLoad '#{type}' has PanelLoad/Voltage (#{voltage}) not specified in default_panels.csv; PanelLoad/PowerRating will be defaulted.")
+          default_panels_csv_data['dishwasher'][voltage] = {}
+          default_panels_csv_data['dishwasher'][voltage]['PowerRating'] = default_panels_csv_data['dishwasher'][HPXML::ElectricPanelVoltage120]['PowerRating']
+        end
+        watts += default_panels_csv_data['dishwasher'][voltage]['PowerRating']
+      end
```
joseph-robertson, line 6042: Still a little confused right in here. We enter this block when:
* there is a dishwasher
* the power rating was not specified

The defaults table specifies power rating and breaker spaces for a 120V dishwasher. What if a user specified a 240V dishwasher? We should assume the power rating of a 120V dishwasher, and throw a warning?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5606,6 +5835,486 @@ def self.get_ceiling_fan_months(weather)
...
+      hpxml_bldg.dishwashers.each do |dishwasher|
+        next if !system_ids.include?(dishwasher.id)
+
+        if voltage == HPXML::ElectricPanelVoltage240
+          runner.registerWarning("PanelLoad '#{type}' has PanelLoad/Voltage (#{voltage}) not specified in default_panels.csv; PanelLoad/BreakerSpaces will be recalculated.")
+          default_panels_csv_data['dishwasher'][voltage] = {}
+          default_panels_csv_data['dishwasher'][voltage]['BreakerSpaces'] = get_breaker_spaces_from_power_watts_and_voltage(watts, voltage)
+        end
+        breaker_spaces += default_panels_csv_data['dishwasher'][voltage]['BreakerSpaces']
+      end
```
joseph-robertson, line 6241: Similar question down here. We've now assumed the power rating of a 120V dishwasher. Breaker spaces was not specified, so we enter this block. We should now calculate breaker spaces using our method with the 120V power rating and 240V specification, and throw a warning?

HPXMLtoOpenStudio/resources/data/default_panels.csv
```diff
@@ -0,0 +1,30 @@
...
+wh_tankless1,240,18000,W,4,hot water,Tankless water heater - one bathroom.
+wh_tankless2,240,24000,W,4,hot water,Tankless water heater - two bathrooms.
+wh_tankless3,240,36000,W,6,hot water,Tankless water heater - three or more bathrooms.
+mechvent,120,3000,W,1,mech vent,Using autosized values for ventilation capacity unless unavailable.
+other,120,373,W,1,other,Only assumes 1 garage door opener if a garage is present and does not account for other permanent appliances (e.g. garbage disposal).
+wh_tank,240,auto,W,auto,hot water,Storage water heater.
+wh_hp,120,auto,W,auto,hot water,120V heat pump water heater. Assumes dedicated circuit.
+wh_hp,240,auto,W,auto,hot water,240V heat pump water heater.
+heating,240,auto,W,auto,heating,Using autosize values for capacity of all heating.
+cooling,240,auto,W,auto,cooling,Using autosize values for capacity of all cooling.
```
joseph-robertson, line 30: We are currently reading directly from this table for breaker spaces. Seems like the Breaker Spaces column in this table should be programmatically/automatically updated? What if our calculation method gets modified? Seems like we should either:
* not read directly from this table but rather use our calculation method, or
* populate Breaker Space fields in this table using our calculation method

BuildResidentialHPXML/measure.rb
```diff
@@ -6844,6 +7136,185 @@ def self.set_pv_systems(hpxml_bldg, args, weather)
...
+      end
+    end
+
+    hpxml_bldg.permanent_spas.each do |permanent_spa|
+      panel_loads.add(type: HPXML::ElectricPanelLoadTypePermanentSpaPump,
+                      power: args[:permanent_spa_pump_panel_load_watts],
+                      addition: args[:permanent_spa_pump_panel_load_addition],
+                      system_idrefs: [permanent_spa.pump_id])
+
+      panel_loads.add(type: HPXML::ElectricPanelLoadTypePermanentSpaHeater,
```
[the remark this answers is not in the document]
joseph-robertson: Good catch.

HPXMLtoOpenStudio/resources/electric_panel.rb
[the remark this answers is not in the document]
lixiliu: I think the code is leaving a lot of room for interpretation. After some careful consideration I think we should stick with the code interpretation (i.e., the former). The argument for 2026 revision is the 2023 code is ambiguous and does not allow credits for loads taken out and it is fixing that.

In any case, it isn't max(CAC, ER furnace) because ER furnace is an existing load that makes up the metered peak in the example.

If we start allowing max(0, CAC-ER furnace) as the new load, I fear we are opening doors to allowing loads to be subtracted from the metered peak. Because in the case of a home that already has an ER furnace AND CAC and it is being converted to a HP, we are supposed to add the HP load directly as the new load in 220.87.

I'd vote not to change the logic here.

HPXMLtoOpenStudio/resources/electric_panel.rb
```diff
@@ -0,0 +1,245 @@
+# frozen_string_literal: true
+
+# Collection of methods related to electric panel load calculations.
+module ElectricPanel
```
lixiliu, line 4: Can we change the name from `electric panel` to `electrical panel` in the code? I think the latter is the more standard term.

shorowit: For better or worse it appears to be `ElectricPanel` in the HPXML schema, so we should probably have everything consistent with that.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
 
+  # FIXME
+  # Returns the number of breaker spaces based on TODO.
+  #
+  # @param TODO
+  # @return TODO
+  def self.get_breaker_spaces_from_power_watts(watts, voltage)
+    required_amperage = watts / voltage
+    num_branches = (required_amperage / 50).ceil
+    num_breakers = num_branches * 2
```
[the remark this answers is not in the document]
joseph-robertson: Hm you must be reviewing/looking at an older version of this method. Your proposed changed has already essentially been pulled in...

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -5591,6 +5821,317 @@ def self.get_ceiling_fan_months(weather)
...
+    return num_breakers
+  end
+
+  # Gets the default properties for electric panels.
+  #
+  # @return [Hash] Map of property type => value
+  def self.get_electric_panel_values()
+    return { panel_voltage: HPXML::ElectricPanelVoltage240,
+             max_current_rating: 150.0, # A
```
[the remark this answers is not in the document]
joseph-robertson: OK

BuildResidentialHPXML/README.md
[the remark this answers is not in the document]
joseph-robertson: OK. I added a `electric_panel_load_heat_pump_voltage` argument to BuildResidentialHPXML.

HPXMLtoOpenStudio/resources/data/default_panels.csv
joseph-robertson: I'll add to the docs the calculation of breaker spaces. It's TBD where/how population of the Breaker Spaces column in the defaults csv file should happen.

joseph-robertson: Decision is to document the breaker spaces calculation in the docs, and also include a note about keeping default_panels.csv up to date in case the breaker spaces calculation method is updated.

HPXMLtoOpenStudio/resources/output.rb
joseph-robertson: Perhaps load-based capacity shouldn't change based on the unit multiplier? Although for meter-based capacity, it would have to since peak electricity would already be scaled?

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -1155,6 +1246,103 @@ def self.append_sizing_results(hpxml_bldgs, results_out)
...
+    results_out << ['Electric Panel Load: Pool Pump (W)', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_panel_loads(hpxml_bldg)[10] }.sum(0.0).round(1)]
+    results_out << ['Electric Panel Load: Well Pump (W)', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_panel_loads(hpxml_bldg)[11] }.sum(0.0).round(1)]
+    results_out << ['Electric Panel Load: Electric Vehicle Charging (W)', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_panel_loads(hpxml_bldg)[12] }.sum(0.0).round(1)]
+    results_out << ['Electric Panel Load: Lighting (W)', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_panel_loads(hpxml_bldg)[13] }.sum(0.0).round(1)]
+    # FIXME: Kitchen? Laundry?
+    results_out << ['Electric Panel Load: Other (W)', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_panel_loads(hpxml_bldg)[14] }.sum(0.0).round(1)]
+
+    # Summary breaker spaces
+    # results_out << [line_break]
+    # results_out << ['Electric Panel Breaker Spaces: Heating Count', hpxml_bldgs.map { |hpxml_bldg| Outputs.get_total_breaker_spaces(hpxml_bldg)[0] }.sum(0)]
```
joseph-robertson, line 1278: BranchCircuits no longer have a "type" associated with them. So do we lose the ability to report these kinds of outputs?

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -6373,6 +6417,42 @@ def initialize(hpxml_element, *args, **kwargs)
     attr_reader(*CLASS_ATTRS)
     attr_accessor(*ATTRS)
 
+    # Returns the branch circuit that the component may be attached to.
+    #
+    # @return [HPXML::BranchCircuit] Branch circuit object
+    def branch_circuit
```
joseph-robertson, line 6423: Each component can be referenced by at most one branch circuit.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -6373,6 +6417,42 @@ def initialize(hpxml_element, *args, **kwargs)
...
+          return branch_circuit
+        end
+      end
+      return
+    end
+
+    # Returns any demand loads that the component may be attached to.
+    #
+    # @return [Array<HPXML::DemandLoad>] List of demand load objects
+    def demand_loads
```
joseph-robertson, line 6438: Components, e.g. heat pump, can be referenced by more than one demand load.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -9256,6 +9493,450 @@ def from_doc(inverter)
...
+    end
+  end
+
+  # Object for /HPXML/Building/BuildingDetails/Systems/ElectricPanels/ElectricPanel/DemandLoads/DemandLoad.
+  class DemandLoad < BaseElement
+    ATTRS = [:id, # [String] SystemIdentifier/@id
+             :type,           # [String] LoadType
+             :power,          # [Double] PowerRating
+             :is_new_load,       # [Boolean] IsNewLoad
+             :component_idrefs]  # [Array<String>] AttachedToComponent/@idref
```
joseph-robertson, line 9809: Does it make more sense for a demand load to reference branch circuit(s)?

HPXMLtoOpenStudio/resources/hpxml.rb
joseph-robertson: We've changed `demand_loads` to `service_feeders`.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 42 + 3114 + 17584, 6)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 42 + 3114, 2)
+
+    test_name = 'Ductless MSHP w/separate gas backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ductless-backup-furnace.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 42 + 3114 + 655, 3)
```
lixiliu, line 364: Same question about 42, 3 spaces is right

lixiliu (commented):
Thanks for the making all these cases! It's a lot to think through.

I should follow up with you about getting the capacity value for each of the HVAC configs so I can also check the load values.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+    args_hash = { 'hpxml_path' => File.absolute_path(@tmp_hpxml_path),
+                  'skip_validation' => true }
+
+    test_name = 'Room air conditioner only'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-room-ac-only.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 0, 0)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 2011, 1)
```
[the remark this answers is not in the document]
joseph-robertson: Why not assume dedicated circuit here like all other 120V system types?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 393 + 14355, 4)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 0, 0)
+
+    test_name = 'Gas furnace + central air conditioner'
+    hpxml, _hpxml_bldg = _create_hpxml('base.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 295, 1)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 300 + 4022, 2)
```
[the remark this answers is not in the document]
joseph-robertson: I think this applies to several of your comments below, but didn't we decide that blower fan power would be determined using fan W/cfm multiplied by airflow cfm?

From the current version of RTD:
image

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 82 + 10766, 2)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 0, 0)
+
+    test_name = 'Large electric boiler only'
+    hpxml, hpxml_bldg = _create_hpxml('base-hvac-boiler-elec-only.xml', test_name)
+    hpxml_bldg.heating_systems[0].heating_capacity = 48000
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 82 + 14355, 4)
```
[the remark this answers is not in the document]
joseph-robertson: Pump power is a function only of electric auxiliary energy (i.e., `electric_auxiliary_energy / 2.08`).

HPXMLtoOpenStudio/tests/test_electric_panel.rb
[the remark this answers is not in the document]
lixiliu: I am thinking more about their usage in terms of an upgrade. People installing new 120V room ACs will unlikely be installing a new dedicated branch circuit for it. Instead, they will plug into an existing receptacle circuit that do not have other loads plugged in, which also counts as a dedicated circuit. This is also the case for most level 2 EV chargers.

When we add the appropriate number of general lighting/receptacle branch circuits later, it may become a bit clearer that we are accounting for plug loads like room ACs as part of receptacles.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 561 + 5839 + 10551, 6)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 600 + 5839, 6)
+
+    test_name = 'ASHP w/integrated gas backup switchover'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-dual-fuel-air-to-air-heat-pump-1-speed.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 561 + 5839, 3)
```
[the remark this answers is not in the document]
joseph-robertson: For these cases, this error is thrown: `Switchover temperature should only be used for a heat pump with fossil fuel backup; use compressor lockout temperature instead.`

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 400 + 4022, 2)
+
+    test_name = 'Large central air conditioner only'
+    hpxml, hpxml_bldg = _create_hpxml('base-hvac-central-ac-only-1-speed.xml', test_name)
+    hpxml_bldg.cooling_systems[0].cooling_capacity = 48000
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 0, 0)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 800 + 7657, 2)
```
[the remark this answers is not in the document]
joseph-robertson: Why not 3?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,542 @@
...
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 202 + 5839, 2)
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 216 + 5839, 2)
+
+    test_name = 'Ducted MSHP w/integrated electric backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ducted.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power_and_breaker_spaces(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 202 + 5839 + 10551, 6)
```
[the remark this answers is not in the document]
joseph-robertson: Where'd the 5839 go?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating], 4)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeCooling], 3)
+
+    test_name = 'ASHP w/out backup'
+    hpxml, hpxml_bldg = _create_hpxml('base-hvac-air-to-air-heat-pump-1-speed.xml', test_name)
+    hpxml_bldg.heat_pumps[0].backup_type = nil
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 561 + 5839)
```
lixiliu, line 315: Shouldn't heating and cooling be exactly the same here?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    # test_name = 'ASHP w/separate electric backup switchover'
+
+    test_name = 'ASHP w/integrated gas backup switchover'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-dual-fuel-air-to-air-heat-pump-1-speed.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 561 + 5839)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 600 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 4)
```
lixiliu, line 339: This is gas furnace backup right? It should be 3 here

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+
+    test_name = 'ASHP w/integrated electric backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-air-to-air-heat-pump-1-speed.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 561 + 5839 + 10551)
```
lixiliu, line 324: This is for backup with overlapping lockout temperatures right?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 600 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 6)
+
+    # Switchover temperature should only be used for a heat pump with fossil fuel backup; use compressor lockout temperature instead.
+    # test_name = 'ASHP w/integrated electric backup switchover'
```
lixiliu, line 329: Can we stub out this case? Should also apply when the lockout temps are the same.

For this the heating load would omit the ODU load or the heat strip load, whichever is smaller.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+
+    test_name = 'Ducted MSHP w/out backup'
+    hpxml, hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ducted.xml', test_name)
+    hpxml_bldg.heat_pumps[0].backup_type = nil
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 202 + 5839)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 216 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 4)
```
lixiliu, line 385: 2 here. If without backup or with separate backup, we assume ODU and IDU are on the same circuit, so we'll have ((IDU+ODU [kw]) / 12 kW).ceil * 2.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+
+    test_name = 'Ducted MSHP w/integrated electric backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ducted.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 202 + 5839 + 10551)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 216 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 6)
```
lixiliu, line 394: 4 here. (202+10551) = 10753 < 12000, 2 slots for IDU + 2 slots for ODU = 4

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+
+    test_name = 'Ducted MSHP w/integrated gas backup switchover'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-dual-fuel-mini-split-heat-pump-ducted.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 202 + 5839)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 216 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 4)
```
lixiliu, line 403: 3

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+
+    test_name = 'Ductless MSHP w/out backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ductless.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 84 + 5839)
```
lixiliu, line 410: No AHU load right?

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 84 + 5839)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 2)
+
+    test_name = 'Ductless MSHP w/separate electric backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ductless-backup-baseboard.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 42 + 3114 + 17584)
```
lixiliu, line 419: No AHU load, MSHP + baseboard here

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 42 + 3114)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating, HPXML::ElectricPanelLoadTypeCooling], 6)
+
+    test_name = 'Ductless MSHP w/separate gas backup'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-mini-split-heat-pump-ductless-backup-furnace.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 42 + 3114 + 655)
```
lixiliu, line 428: What is the 655 here? The gas furnace?

lixiliu (commented):
Some more comments

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeCooling], 2)
+
+    test_name = 'Large electric furnace + central air conditioner'
+    hpxml, hpxml_bldg = _create_hpxml('base-hvac-furnace-elec-central-ac-1-speed.xml', test_name)
+    hpxml_bldg.heating_systems[0].heating_capacity = 48000
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 393 + 14067)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 300 + 4022)
```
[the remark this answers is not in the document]
joseph-robertson: Alternatively, we could:
* take the max of heating and cooling air handler loads (the proposed fix)
* always take the heating air handler load

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 0)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating], 1)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeCooling], 0)
+
+    test_name = 'Electric boiler only'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-boiler-elec-only.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 82 + 10766)
```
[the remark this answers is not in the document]
joseph-robertson: Is there an inconsistency in the docs around EAE for in-unit boilers? Elec gives 170 and gas gives 200...

HPXMLtoOpenStudio/tests/test_electric_panel.rb
```diff
@@ -0,0 +1,631 @@
...
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeHeating], 4)
+    _test_occupied_spaces(hpxml_bldg, [HPXML::ElectricPanelLoadTypeCooling], 2)
+
+    test_name = 'Central air conditioner only'
+    hpxml, _hpxml_bldg = _create_hpxml('base-hvac-central-ac-only-1-speed.xml', test_name)
+    XMLHelper.write_file(hpxml.to_doc, @tmp_hpxml_path)
+    _model, _hpxml, hpxml_bldg = _test_measure(args_hash)
+
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeHeating, 0)
+    _test_service_feeder_power(hpxml_bldg, HPXML::ElectricPanelLoadTypeCooling, 400 + 4022)
```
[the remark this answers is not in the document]
lixiliu: 0.5 W/cfm for CAC only and 0.375W/cfm for attachedToFurnace

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: Same reason as above -- heating and cooling airflow rates are different.

joseph-robertson: Without backup, taking the max here or across heating/cooling loads doesn't matter; but with backup it matters (slightly).

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: Yes, it should. Compared to a ASHP w/out backup (which should have 2 for the AHU), it's likely we are adding a HP to already existing gas furnace.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: Yes.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: This applies to MSHP w/out backup.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: This is like `ASHP w/integrated gas backup switchover`.

HPXMLtoOpenStudio/tests/test_electric_panel.rb
joseph-robertson: Yes.

joseph-robertson:
- [ ] Allow all of Total/Occupied/Headroom? And provide warnings?
- [ ] What happens if, e.g., there is gas dryer and something else on a certain breaker? What happens in the outputs? Do we need the category breakout for Breaker Spaces?
- [ ] Should defaulting of ServiceFeeders be all or none?
- [ ] How do we go from BranchCircuits to output categories? Need more error checking on, e.g., a component shared across multiple service feeders or branch circuits. Need more real-word example cases; preventing/warnings may be ok here.
- [ ] Are there entered values which may make the NEC calculation invalid? E.g., lighting.

A large question is how useful this is going to be for HPXML users. Will they get warnings? Will they get invalid outputs?

lixiliu:
> A large question is how useful this is going to be for actual HPXML users. Based on entered information, will they get warnings? Will they get invalid outputs?
> 
> * [ ]  BranchCircuits. Allow all of Total/Occupied/Headroom to be entered? If the numbers don't add up, provide warnings?
> * [ ]  What happens if, e.g., there is gas dryer and something else on a certain breaker? What happens in the outputs? Do we _need_ the category breakout for Breaker Spaces?
> * [ ]  Should defaulting of ServiceFeeders be all or none? Are these really inputs that should be exposed? How does someone performing an NEC calculation determine, e.g., power rating values?
> * [ ]  How do we go from BranchCircuits to output categories? Need more error checking on, e.g., a component shared across multiple service feeders or branch circuits. Need more real-word example cases; preventing/warnings may be acceptable here.
> * [ ]  Are there entered power rating values which may make the NEC calculation invalid? E.g., lighting.
> * [ ]  Perhaps add more "real world" test cases/examples to the `real_homes` folder.
> * [ ]  BranchCircuits will only be created when doing NEC calc(s) (i.e., when ServiceFeeders are present). Otherwise, we'll assume that BranchCircuits is "complete" if any BranchCircuits are present (output will only contain breaker space information; we'll need to think about how to categorize Breaker Spaces).
> * [ ]  When doing NEC calc(s) and BranchCircuits are present, throw a warning that BranchCircuits is not empty (this can happen when the Build measure writes HP, WH, CD, etc. voltage). Otherwise, branch circuits (i.e., air handler) may be double counted.
> * [ ]  Kitchen (or laundry) service feeder can be greater than 3000 (or 1500), in increments of 1500; throw a warning if greater than 3000 and/or not a multiple of 1500.
> * [ ]  Lighting service feeder must be equal to 3 times CFA; throw a warning if it isn't.

HPXMLtoOpenStudio/resources/electric_panel.rb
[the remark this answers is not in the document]
lixiliu: @joseph-robertson found a bug in my recommended change that for part A, we are not currently accounting for max of heating or cooling. We should be.

PR #1867, opened 2024-10-23
BuildResidentialHPXML: make conditional arguments optional

joseph-robertson:
## Pull Request Description

Change any required arguments that are ignored based on other arguments, to optional.

Remove `setDefaultValue` from all OpenStudio arguments. Required arguments must now pass a value.

Create a `make_argument` wrapper for passing in argument name, type, required, etc and determining/creating appropriate OpenStudio arguments automatically:
* if there is OS-HPXML default (must be optional argument):
  * choice: append "auto" choice
  * boolean: create a choice argument with choices "true", "false", "auto"
  * double/integer: create a string argument ("auto" can be passed in)
* arguments with OS-HPXML default gets description about OS-HPXML defaults and href appended

Create `convertArgumentValues` method called from beginning of `run` method:
* delete arguments that passed in "auto"
* "true" to true, "false" to false, "10" to 10, "10.5" to 10.5, etc

## Checklist

Not all may apply:


shorowit (approved):
LGTM. Will wait on ResStock testing before merging.

joseph-robertson:
Should all optional arguments that use the OS-HPXML default when not provided have a default argument value set? For example, `site_type`, `site_shielding_of_home`, `site_soil_and_moisture_type` use OS-HPXML defaults but don't set a default argument value. Arguments `site_iecc_zone`, `site_city` don't set a default argument value but don't use OS-HPXML defaults.

The implication of not doing this, as I'm noticing in ResStock, is that I'd need to check for presence of "OS-HPXML default" in the argument description vs. checking that a default argument value is set.

joseph-robertson:
Does `ducts_supply_buried_insulation_level` and `ducts_return_buried_insulation_level` need updated argument descriptions pointing to https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#air-distribution?

HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -624,3 +624,38 @@ def is_integer?
...
+    elsif type == 'String'.to_OSArgumentType
+      begin
+        value = Float(value)
+        if value % 1 == 0
+          args[name] = Integer(value)
+        else
+          args[name] = value
+        end
+      rescue
+      end
```
shorowit, line 657: This is a bad idea. You should only be converting to integer/float for arguments that we've declared as integer/float, not simply converting anytime it's possible. I assume this is the reason that you then had to change a bunch of code so that it converts back to string.

You'll need to refactor the code so that you have access to the original desired argument type and use it here.

BuildResidentialHPXML/README.md
```diff
@@ -101,7 +101,7 @@ Specifies the unavailable period date ranges. Enter a date range like "Dec 15 -
 
 **Schedules: Unavailable Period Window Natural Ventilation Availabilities**
 
-The availability of the natural ventilation schedule during unavailable periods. Valid choices are: regular schedule, always available, always unavailable. If multiple periods, use a comma-separated list. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.8.1/workflow_inputs.html#hpxml-unavailable-periods'>HPXML Unavailable Periods</a>) is used.
+The availability of the natural ventilation schedule during unavailable periods. Valid choices are: regular schedule, always available, always unavailable. If multiple periods, use a comma-separated list. If auto or not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.8.1/workflow_inputs.html#hpxml-unavailable-periods'>HPXML Unavailable Periods</a>) is used.
```
shorowit, line 104: Can you change it to 'auto' or "auto"?

Changelog.md
```diff
@@ -37,6 +37,7 @@ __New Features__
 - BuildResidentialHPXML measure:
   - **Breaking change**: Replaced `slab_under_width` and `slab_perimeter_depth` arguments with `slab_under_insulation_width` and `slab_perimeter_insulation_depth`
   - **Breaking change**: Replaced `schedules_vacancy_periods`, `schedules_power_outage_periods`, and `schedules_power_outage_periods_window_natvent_availability` arguments with `schedules_unavailable_period_types`, `schedules_unavailable_period_dates`, and `schedules_unavailable_period_window_natvent_availabilities`; this improves flexibility for handling more unavailable period types.
+  - **Breaking change**: Removed `window_<side>_wwr` argument; a value less than 1 can be provided to the `window_area_<side>` and will be treated as a window-to-wall ratio.
```
shorowit, line 40: Need to move this to the next version.

HPXMLtoOpenStudio/resources/meta_measure.rb
```diff
@@ -624,3 +624,38 @@ def is_integer?
     return true
   end
 end
+
+# TODO
```
shorowit, line 628: Do this now (and anywhere else you added TODOs).

HPXMLtoOpenStudio/resources/meta_measure.rb
shorowit: Also, should this really be in meta_measure.rb? Do you anticipate using this anywhere else than the BuildResHPXML measure?

workflow/sample_files/base-auto.xml
shorowit, line 1: Does it really make sense to create this new sample file? Seems like it should just be a BuildResHPXML unit test.

workflow/template-build-and-run-hpxml-with-stochastic-occupancy.osw
```diff
@@ -6,11 +6,25 @@
...
+        "cooking_range_oven_present": true,
+        "cooking_range_oven_fuel_type": "natural gas",
+        "ceiling_fan_present": true,
+        "misc_plug_loads_television_present": true,
         "air_leakage_house_pressure": 50,
         "air_leakage_units": "ACH",
         "air_leakage_value": 3,
         "ceiling_assembly_r": 39.3,
-        "cooling_system_cooling_capacity": 24000.0,
+        "cooling_system_cooling_capacity": "24000.0",
```
shorowit, line 27: Ugh, really? This might be a deal breaker. This is totally confusing and no one will be able to keep track of when they need a number vs a string. I didn't realize OpenStudio was going to require this.

BuildResidentialHPXML/measure.rb
```diff
@@ -3232,304 +4671,553 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
...
+    )
+
+    return args
+  end
 
+  # TODO
+  def defaultOptionalArgumentValues(args)
+    # these were previously required arguments with default values set (i.e., arg.setDefaultValue(xxx))
+    args[:floor_over_foundation_assembly_r] = 28.1 if args[:floor_over_foundation_assembly_r].nil?
+    args[:floor_over_garage_assembly_r] = 28.1 if args[:floor_over_garage_assembly_r].nil?
```
shorowit, line 5140: I was really hoping we wouldn't need these (totally arbitrary) defaults. Why?

But if for some reason we *do* need them, this is definitely a step back from what we had before. Previously, at least a user opening the measure in the OS App or PAT would see the default value. Now they happen completely behind the scenes.

workflow/template-build-and-run-hpxml-with-stochastic-occupancy.osw
shorowit: Maybe we can update OpenStudio so that it allows this?

BuildResidentialHPXML/measure.rb
joseph-robertson: Some, yes, are totally arbitrary. I can fix/refactor those.

Others, for example heat pump backup fuel, are not. We can report the default value in the argument description (there's already precedent for this) and set it in its respective `set_xxx` method.

shorowit: Why does heat pump backup fuel need a default?

HPXMLtoOpenStudio/resources/meta_measure.rb
joseph-robertson: I put it here because I found that ResStock needs to use it (with `delete_auto = false`).

shorowit:
Doesn't seem like we're going to proceed with this path, closing.

PR #1877, opened 2024-11-07
Address GSHP fan unit bug

afontani:
## Pull Request Description

This PR addresses some unit inconsistencies in the GSHP fan/pump equations. There seem to be some unit issues with these equations: https://github.com/NREL/OpenStudio-HPXML/blob/2f7c47554c5db31052781fbf29453023f805fbe3/HPXMLtoOpenStudio/resources/hvac.rb#L2023-L2028

### Unit issues:
1. The `power_f` and `power_p` have units of KW*h/BTU but is expected to be Wh/BTU in the EIR calculation.
2. IER is supposed to be unitless and the terms `-power_f` and `-power_p` are not unitless and need another unit conversion like in the first term of the equation.

## Checklist

Not all may apply:


afontani:
@jmaguire1: Some of the verification tests are failing. I think the units are now right, but the cooling COP calculated is 6.86 in the example model. I think this is a pretty big change. The test is here: https://github.com/NREL/OpenStudio-HPXML/blob/2f7c47554c5db31052781fbf29453023f805fbe3/HPXMLtoOpenStudio/tests/test_hvac.rb#L1054. 

Another failure is an additional 10 feet in the borefield depth per borehole, which I do not think is that big of a deal.

jmaguire1:
@afontani: Yeah, it is indeed a much bigger deal than I thought, and suggests we would have been really substantially underestimating the impact of the fan/pump and doing net to gross calculations:

image


For reference, this is the base-hvac-ground-to-air-heat-pump, with a nominal COP of 3.6 and EER of 16.6. Maybe it's worth talking to someone like Jeff Munk and being sure we're doing things as intended before making such a big change? I just worry we have some other issue here with such a large difference in results. I'd like to feel more confident we've definitely done this right and then not have to revisit this again.

afontani:
From a conversation with Jeff Munk, Scott, Jeff Maguire: There is another modification where the 0.375 is the fan power. From ANSI/RESNET/ICC 301, 0.2 is used for ductwork. the remaining is to overcome the static pressure.

**Actions:**
- Assume rated fan power use 50% to scale down for the ductwork.
- Create a new variable and comment what it is right above the power_f and power_p equations.

PR #1878, opened 2024-11-08
Allow Modeling Two Speed & Var Speed Geothermal Heat Pumps

yzhou601:
## Pull Request Description

Allow optional compressor type inputs to be specified. Allow modeling two and variable speed geothermal heat pumps.
- [x] Calculate two speed system performance curve coefficients from catalog data, using E+ CurveFitTool.
- [x] Add two speed ghp system model to OS-HPXML
- [x] Calculate variable speed system performance curve coefficients from catalog data, using E+ CurveFitTool.
- [x] Add variable speed ghp system model to OS-HPXML
- [x] ~Default number of speeds based on efficiency~ Require `CompressorType` input
- [x] ~Allow desuperheater to work with new GSHP coil object https://github.com/NREL/EnergyPlus/pull/10945 (Not in this PR, this PR is targeted to be merged before new OS/E+)~
- [x] Address pump flow rate

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -594,34 +594,92 @@ def self.apply_ground_to_air_heat_pump(model, runner, weather, heat_pump, hvac_s
...
+        eir_fwf_curve = Model.add_curve_quadratic(
+          model,
+          name: "Cool-EIR-fWF#{i + 1}",
+          coeff: [1, 0, 0],
+          min_x: 0, max_x: 2, min_y: 0, max_y: 2
+        )
+        # Recoverable heat modifier as a function of indoor wet-bulb and water entering temperatures.
+        waste_heat_ft = Model.add_curve_biquadratic(
+            model,
+            name: "WastHeat-FT#{i + 1}",
```
jmaguire1, line 673: "WasteHeat" (I think there's a typo, missing the "e")

yzhou601:
Update: Performance curves for two speed systems are figured out.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -576,56 +576,194 @@ def self.apply_ground_to_air_heat_pump(model, runner, weather, heat_pump, hvac_s
...
+    else
+      num_speeds = hp_ap.cool_capacity_ratios.size
+      # TODO: Curve placeholder
+      plf_fplr_curve = Model.add_curve_quadratic(
+        model,
+        name: 'Cool-PLF-fPLR',
+        coeff: [1, 0, 0],
+        min_x: 0, max_x: 1, min_y: 0.7, max_y: 1
+      )
+      clg_coil = OpenStudio::Model::CoilCoolingWaterToAirHeatPumpVariableSpeedEquationFit.new(model, plf_fplr_curve)
```
shorowit, line 634: Be sure to audit the code for references to e.g. `CoilCoolingWaterToAirHeatPumpEquationFit`. I'm seeing at least two place in our code (installation quality EMS program and desuperheaters) that presumably need to be updated.

yzhou601: We talked about desuperheaters, the variable speed coil type is not yet supported by E+ to be connected with desuperheater, so we may have to temporarily ignore that feature (and document that only single speed gshp is currently supported).

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2807,23 +2807,37 @@ def self.apply_hvac_equipment_adjustments(mj, runner, hvac_sizings, weather, hva
 
       gshp_coil_bf = 0.0806
```
yzhou601, line 2808: We can look at E+ bypass factor calculations.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: @shorowit @afontani The E+ IO freeze for March release is 2/12, is it still possible to add this capability to Coil:Cooling:WaterToAirHeatPump:VariableSpeedEquationFit for the upcoming release? There're 12 open PR for the E+ team to review before 2/12, so I'm not sure if it's possible to add this one and get someone's review. Also the date is overlapping with the two speed/var speed tasks (plan to be wrapped up before end of Feb). Is it fine if we prioritize the two/var speed system work, then if we still get some space we can add the capability to the E+ Oct release?

We can explicitly document that the 2/var speed is not currently supporting desuperheater, and if user still needs to model it, the workaround will be specifying a single speed system (instead of defaulting based on efficiency, or specifying it as two/var speed systems) with desuperheater, which will be handled in the same way as before.

shorowit: It seems like it should be easy to review and there are many possible reviewers, I wouldn't worry about that.

As far as I'm aware, you are allowed to just make the IDD change before the IO freeze and complete the implementation later. So that could give you another 4-6 weeks? You could confirm by asking Edwin or Mike.

Documenting that we are introducing a breaking change is not a great solution. For a downstream software tool (like REM), it could be a lot of work -- adding a new compressor type input in their user interface for GSHPs, adding logic when there's a desuperheater, etc. And their user would get different heating/cooling results based on whether there is a desuperheater or not? That's confusing.

In my opinion, either we make it a breaking change or we defer this for a subsequent OS-HPXML release.

shorowit: (Sorry, the last sentence should have said **non**-breaking change. Fixed.)

yzhou601: > It seems like it should be easy to review and there are many possible reviewers, I wouldn't worry about that.
> 
> As far as I'm aware, you are allowed to just make the IDD change before the IO freeze and complete the implementation later. So that could give you another 4-6 weeks? You could confirm by asking Edwin or Mike.
> 
> Documenting that we are introducing a breaking change is not a great solution. For a downstream software tool (like REM), it could be a lot of work -- adding a new compressor type input in their user interface for GSHPs, adding logic when there's a desuperheater, etc. And their user would get different heating/cooling results based on whether there is a desuperheater or not? That's confusing.
> 
> In my opinion, either we make it a non-breaking change or we defer this for a subsequent OS-HPXML release.

The IDD change before IO freeze and implement later sounds interesting, I'm fine with that, assuming we'll wrap up most of the 2/var speed system OS-HPXML changes before that. Do you mean that it's better to wait for the E+ release and this PR is non-breaking then we can merge it? @afontani Is it fine to do so? The E+ release is end of March, so probably the merge will be early April, but we can still have most functions being implemented so that ResStock can test the batch runs based on this branch.

yzhou601: Another thought to make it non-breaking to merge early. So what about if we exclude the efficiency-based defaulting for now, and always default the compressor type to be single speed, and when the E+ capability is added then we add the efficiency-based defaulting approach?

shorowit: That seems like a great idea/compromise to me!

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -41,7 +41,7 @@ def test_hvac_configurations
     args_hash = { 'hpxml_path' => File.absolute_path(@tmp_hpxml_path),
                   'skip_validation' => true }
     Dir["#{@sample_files_path}/base-hvac*.xml"].each do |hvac_hpxml|
-      next unless hvac_hpxml.include? 'ground-to-air'
```
yzhou601, line 44: This line only ran gshp sizing, felt like it's probably left by debugging something.

shorowit: Good catch, yes it seems like this line should be removed.

yzhou601: Opened another PR to remove this line and ran the whole test set.

BuildResidentialHPXML/measure.rb
```diff
@@ -5767,7 +5767,7 @@ def self.set_heat_pumps(hpxml_bldg, args)
       end
     end
 
-    if [HPXML::HVACTypeHeatPumpAirToAir, HPXML::HVACTypeHeatPumpMiniSplit].include? heat_pump_type
+    if [HPXML::HVACTypeHeatPumpAirToAir, HPXML::HVACTypeHeatPumpMiniSplit, HPXML::HVACTypeHeatPumpGroundToAir].include? heat_pump_type
       compressor_type = args[:heat_pump_cooling_compressor_type]
```
joseph-robertson, line 5771: Need to update the argument description for `heat_pump_cooling_compressor_type` to include ground-to-air systems.

BuildResidentialHPXML/measure.rb
```diff
@@ -1423,7 +1423,7 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
 
     arg = OpenStudio::Measure::OSArgument::makeChoiceArgument('heat_pump_cooling_compressor_type', compressor_type_choices, false)
```
yzhou601, line 1424: Now that it's required for ASHP, MSHP, and GSHP (the same applies to cooling systems), should we add error-checking if it's not provided for these system types?

shorowit: I think it's fine to just rely on the downstream schematron error-checking. But at a minimum, we should make sure the description is updated so the user knows which system types this argument is required for.

HPXMLtoOpenStudio/tests/test_hvac_sizing.rb
```diff
@@ -97,7 +97,7 @@ def test_hvac_configurations
 
           # Get more values for heat pump checks
           htg_load = autosized_bldg.hvac_plant.hdl_total
-          clg_load = autosized_bldg.hvac_plant.cdl_sens_total + autosized_bldg.hvac_plant.cdl_lat_total
```
yzhou601, line 100: From doc: 
- HERS: same as ACCA except autosized heat pumps have their nominal capacity sized equal to at least the larger of heating and sensible cooling design loads.

The test here is checking capacities are equal to or greater than the larger of heating and **total** cooling design loads. Fixed it to be aligned with expectations

shorowit (commented):
We need to reconcile this PR with https://github.com/NREL/OpenStudio-HPXML/pull/1879. For example, one thing we should do is update the `FanMotorType` defaulting here to be based on compressor type and not rated efficiency (as we do for ASHPs).

yzhou601:
> We need to reconcile this PR with #1879. For example, one thing we should do is update the `FanMotorType` defaulting here to be based on compressor type and not rated efficiency (as we do for ASHPs).

I scheduled a meeting tomorrow to talk about what needs to address in each PR for GHP related changes. There's also hvac sizing approach that I used the ashp methods that may have changed in the RESNET PR, we can walk through the changes and see what we can bring into this PR.

workflow/tests/base_results/results_simulations_energy.csv
```diff
@@ -278,12 +278,17 @@ base-hvac-furnace-oil-only.xml,54.479,54.479,30.812,30.812,0.0,23.667,0.0,0.0,0.
...
-base-hvac-ground-to-air-heat-pump-heating-only.xml,36.784,36.784,36.784,36.784,0.0,0.0,0.0,0.0,0.0,0.0,5.314,1.265,0.0,0.0,0.0,0.0,8.992,0.0,0.0,4.507,0.0,0.334,0.0,0.0,0.0,0.0,2.146,0.0,0.0,0.319,0.365,1.513,1.529,0.0,2.116,8.384,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
-base-hvac-ground-to-air-heat-pump.xml,40.055,40.055,40.055,40.055,0.0,0.0,0.0,0.0,0.0,0.0,5.348,1.273,0.0,0.0,2.352,0.932,9.013,0.0,0.0,4.507,0.0,0.334,0.0,0.0,0.0,0.0,2.072,0.0,0.0,0.319,0.365,1.513,1.529,0.0,2.116,8.384,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```
yzhou601, line 286: I compared the IDFs between this branch and master and found that the single speed ghp test results changes are caused by:
1. The rated air flow rate change, this is a bugfix that we used the design cfm as its rated before, and now it's fixed by using the rated cfm/ton * capacity(ton)
2. The hvac sizing changes that caused a small diff in cooling cfm, I think it's a combined impact of:  
     (1) introduced the heating performance curve adjustment to convert design to rated conditions, which was missing before 
    (2) In calculating the cooling performance curve, we precalculated the cooling air flow using sensible cooling loads and cfm/ton, it's now changed to use total cooling loads.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2598,6 +2599,130 @@ def self.apply_hvac_piping_load(hvac_loads, zone_loads, hvac_heating)
     zone_loads.Heat_Tot += heat_load
   end
 
+  # Iteratively calculate design SHR --> design sensible capacity --> use design sensible capacity to RECALCULATE design airflow using MJ approach
```
jmaguire1, line 2602: Is this part of the other PR? Doesn't seem specific to GHPs?

jmaguire1 (changes requested):
Mostly minor stuff, but see the individual comments I left. Some might just need some discussion rather than a change.

BuildResidentialHPXML/README.md
```diff
@@ -2048,7 +2061,7 @@ The rated efficiency value of the cooling system. Ignored for evaporative cooler
 
 **Cooling System: Cooling Compressor Type**
 
-The compressor type of the cooling system. Only applies to central air conditioner and mini-split. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#central-air-conditioner'>Central Air Conditioner</a>, <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#mini-split-air-conditioner'>Mini-Split Air Conditioner</a>) is used.
+The compressor type of the cooling system. Only applies to central air conditioner and mini-split.
```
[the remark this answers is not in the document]
yzhou601: This is for cooling system not heat pump, for the heat pump ground-to-air is added.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -563,8 +569,8 @@ def self.apply_ground_to_air_heat_pump(runner, model, weather, heat_pump, hvac_s
 
     htg_cfm = heat_pump.heating_airflow_cfm
     clg_cfm = heat_pump.cooling_airflow_cfm
-    htg_cfm_rated = heat_pump.airflow_defect_ratio.nil? ? htg_cfm : (htg_cfm / (1.0 + heat_pump.airflow_defect_ratio))
-    clg_cfm_rated = heat_pump.airflow_defect_ratio.nil? ? clg_cfm : (clg_cfm / (1.0 + heat_pump.airflow_defect_ratio))
+    htg_cfm_rated = UnitConversions.convert(heat_pump.heating_capacity, 'Btu/hr', 'ton') * hp_ap.heat_capacity_ratios[-1] * hp_ap.heat_rated_cfm_per_ton[-1]
```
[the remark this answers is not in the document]
yzhou601: The `htg_cfm` and `clg_cfm` are the operational cfms which should already account for the airflow defects, the rated airflows are used for coil rated inputs, they should be calculated with rated cfm/ton and capacities instead of using the operational, so this is actually a bugfix.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -576,56 +582,213 @@ def self.apply_ground_to_air_heat_pump(runner, model, weather, heat_pump, hvac_s
...
+          coeff: hp_ap.heat_eir_fflow_spec[i],
+          min_x: 0, max_x: 2, min_y: 0, max_y: 2
+        )
+        eir_fwf_curve = Model.add_curve_quadratic(
+          model,
+          name: "Heat-EIR-fWF#{i + 1}",
+          coeff: hp_ap.heat_eir_fwf_spec[i],
+          min_x: 0.45, max_x: 2, min_y: 0, max_y: 2
+        )
+        # Recoverable heat modifier as a function of indoor wet-bulb and water entering temperatures.
```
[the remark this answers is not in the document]
yzhou601: This curve is not for water heater desuperheaters, it is for hot gas reheat: https://bigladdersoftware.com/epx/docs/24-2/input-output-reference/group-heating-and-cooling-coils.html#field-speed-x-reference-unit-waste-heat-fraction-of-power-input. The desuperheater improvements are included in the #1978

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1985,45 +2196,184 @@ def self.get_heat_cfm_per_ton(compressor_type, use_cop_or_htg_sys = false)
...
-    # E+ equation fit coil coefficients following approach from Tang's thesis:
-    # See Appendix B Figure B.3 of  https://shareok.org/bitstream/handle/11244/10075/Tang_okstate_0664M_1318.pdf?sequence=1&isAllowed=y
-    # Coefficients generated by catalog data: https://www.climatemaster.com/download/18.274be999165850ccd5b5b73/1535543867815/lc377-climatemaster-commercial-tranquility-20-single-stage-ts-series-water-source-heat-pump-submittal-set.pdf
-    # Data point taken as rated condition:
-    # EWT: 60F EAT: 70F AFR: 1200 cfm, WFR: 4.5 gpm
-
-    # Heating Curves
-    hp_ap.heat_cap_curve_spec = [[-3.75031847962047, -2.18062040443483, 6.8363364819032, 0.188376814356582, 0.0869274802923634]]
-    hp_ap.heat_power_curve_spec = [[-8.4754723813072, 8.10952801956388, 1.38771494628738, -0.33766445915032, 0.0223085217874051]]
+    hp_ap.cool_capacity_ratios = get_cool_capacity_ratios_gshp(heat_pump)
```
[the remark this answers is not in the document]
yzhou601: Good catch, yes this should only be used for advanced model!

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1985,45 +2196,184 @@ def self.get_heat_cfm_per_ton(compressor_type, use_cop_or_htg_sys = false)
...
+      cool_cop_ratios = [1.0]
+      heat_cop_ratios = [1.0]
+
+    when HPXML::AdvancedResearchGeothermalModelTypeAdvanced
+      case heat_pump.compressor_type
+      when HPXML::HVACCompressorTypeSingleStage
+        # Cooling Curves
+        # E+ Capacity and EIR as function of temperature curves(bi-quadratic) generated using E+ HVACCurveFitTool
+        # See: https://bigladdersoftware.com/epx/docs/24-2/auxiliary-programs/hvac-performance-curve-fit-tool.html#hvac-performance-curve-fit-tool
+        # Catalog data from ClimateMaster residential tranquility 30 premier two-stage series Model SE036: https://files.climatemaster.com/RP3001-Residential-SE-Product-Catalog.pdf
```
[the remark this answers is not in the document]
yzhou601: Yes this is a copy/paste from dual speed, do you mind sharing me the link to the single-speed data?

HPXMLtoOpenStudio/resources/hvac_sizing.rb
yzhou601: This is the approach for ASHP , I refactored a bit so that it's a separate method and can be called by both ghp and ashp.

HPXMLtoOpenStudio/resources/hvac.rb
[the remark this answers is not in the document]
yzhou601: Assuming we're using the same catalog data as previous equation fit model, I updated the comment here. Let me know if it's not the case.

BuildResidentialHPXML/README.md
```diff
@@ -193,6 +193,19 @@ Research feature to select the type of defrost model. Use standard for default E
 
 <br/>
 
+**Simulation Control: Geothermal Model Type**
+
+Research feature to select the type of geothermal model. Use simple for simple E+ geothermal coil modeling. Use advanced for an improved model that better accounts for coil staging; using advanced may impact simulation runtime. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#hpxml-simulation-control'>HPXML Simulation Control</a>) is used.
```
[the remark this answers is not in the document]
yzhou601: I'm leaving this question to Scott, also this is the docs for buildresidentiaHPXML, if we want to add more details to docs, it could be in the OS-HPXML too.

yzhou601:
Stock level results:
image
The team is pretty satisfied with the results, we can proceed to request reviews and finalize this PR for merging. The E+ bug is ignored for now and it shouldn't require extra efforts after switching to the new OS/E+ except for the #1978 .

jmaguire1 (approved):

shorowit (commented):
There was a lot to review here. I didn't really look at or scrutinize the new model's performance maps but tried to look at pretty much everything else.

BuildResidentialHPXML/README.md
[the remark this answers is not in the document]
yzhou601: I think the requirement of `CompressorType` for air source hvac systems is already pulled into this PR with https://github.com/NREL/OpenStudio-HPXML/pull/1949, and I found this description is not updated so I modified it here.

BuildResidentialHPXML/README.md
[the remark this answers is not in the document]
shorowit: We don't need to include this language just because an EMS program is used (we use EMS in every simulation -- e.g., for infiltration). Only if the EMS program has a noticeable impact on runtime speed.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -576,56 +582,231 @@ def self.apply_ground_to_air_heat_pump(runner, model, weather, heat_pump, hvac_s
...
-    clg_coil.setRatedTotalCoolingCapacity(UnitConversions.convert(heat_pump.cooling_capacity, 'Btu/hr', 'W'))
-    clg_coil.setRatedSensibleCoolingCapacity(UnitConversions.convert(hp_ap.cooling_capacity_sensible, 'Btu/hr', 'W'))
+    if [HPXML::AdvancedResearchGeothermalModelTypeSimple].include? hpxml_header.geothermal_model_type
+      # Derived from: https://www.e3s-conferences.org/articles/e3sconf/pdf/2018/19/e3sconf_eko-dok2018_00139.pdf
+      plf_fplr_curve = Model.add_curve_cubic(
+        model,
+        name: 'Cool-PLF-fPLR',
+        coeff: [0.4603, 1.6416, -1.8588, 0.7605],
+        min_x: 0, max_x: 1, min_y: 0.7, max_y: 1
+      )
```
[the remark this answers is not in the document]
yzhou601: Yes, we want to use the same PLR curve here to separate the impact of differences between advanced and simple models

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -691,6 +872,7 @@ def self.apply_ground_to_air_heat_pump(runner, model, weather, heat_pump, hvac_s
     setpoint_mgr_follow_ground_temp.addToNode(plant_loop.supplyOutletNode)
 
     # Pump
+    control_type = heat_pump.compressor_type == HPXML::HVACCompressorTypeVariableSpeed ? EPlus::PumpControlTypeContinuous : EPlus::PumpControlTypeIntermittent
```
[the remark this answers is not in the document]
yzhou601: This is something confusing, I noticed this input when trying to fix the pump mass flow rate, but finally found it having no impact. I was imagining this input similar to the "Discrete"/"Continuous" on the Fan speed control type but actually not. I could change the pump control type to intermittent for all the system types and it should have little impact.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1982,48 +2211,194 @@ def self.get_heat_cfm_per_ton(compressor_type, use_cop_or_htg_sys = false)
...
+    power_f = rated_fan_watts_per_cfm * hp_ap.cool_rated_cfm_per_ton[-1] / UnitConversions.convert(1.0, 'ton', 'Btu/hr') # result is in W per Btu/hr of capacity
     power_p = rated_pump_watts_per_ton / UnitConversions.convert(1.0, 'ton', 'Btu/hr') # result is in W per Btu/hr of capacity
-
-    cool_eir = UnitConversions.convert(((1 - UnitConversions.convert(power_f, 'Wh', 'Btu')) / heat_pump.cooling_efficiency_eer - power_f - power_p), 'Wh', 'Btu')
-    heat_eir = (1 + UnitConversions.convert(power_f, 'Wh', 'Btu')) / heat_pump.heating_efficiency_cop - UnitConversions.convert(power_f + power_p, 'Wh', 'Btu')
-
-    hp_ap.cool_rated_cops = [1.0 / cool_eir]
-    hp_ap.heat_rated_cops = [1.0 / heat_eir]
+    cool_eir_rated = UnitConversions.convert(((hp_ap.cool_capacity_ratios[-1] - UnitConversions.convert(power_f, 'Wh', 'Btu')) / heat_pump.cooling_efficiency_eer - power_f - power_p), 'Wh', 'Btu')
+    heat_eir_rated = (hp_ap.heat_capacity_ratios[-1] + UnitConversions.convert(power_f, 'Wh', 'Btu')) / heat_pump.heating_efficiency_cop - UnitConversions.convert(power_f + power_p, 'Wh', 'Btu')
```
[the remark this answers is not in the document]
yzhou601: No we are not changing the equation here except that I was using the last capacity ratio instead of hardcoded 1, but they should be 1. I was proposing something else in the first place so it looks like a diff, I will revert.

HPXMLtoOpenStudio/resources/hvac_sizing.rb
```diff
@@ -2920,28 +2946,38 @@ def self.apply_hvac_equipment_adjustments(mj, runner, hvac_sizings, weather, hva
...
-        hvac_sizings.Heat_Capacity_Supp = hvac_sizings.Heat_Load_Supp
-
-        # For single stage compressor, when heating capacity is much larger than cooling capacity,
-        # in order to avoid frequent cycling in cooling mode, heating capacity is derated to 75%.
-        if hvac_sizings.Heat_Capacity >= 1.5 * hvac_sizings.Cool_Capacity
-          hvac_sizings.Heat_Capacity = hvac_sizings.Heat_Load * 0.75
+        if (hpxml_header.geothermal_model_type == HPXML::AdvancedResearchGeothermalModelTypeSimple) && (hvac_heating.compressor_type == HPXML::HVACCompressorTypeSingleStage)
+          # For single stage compressor, when heating capacity is much larger than cooling capacity,
+          # in order to avoid frequent cycling in cooling mode, heating capacity is derated to 75%.
+          # Currently only keep it for simple ghp models
```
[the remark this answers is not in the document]
yzhou601: Yes, the assumption here seems to be suitable for single-speed systems, while @jmaguire1 found that the single-speed equipment we looked into had a dehumidification mode which seems to be reasonable to remove this adjustment. We kept it for the old model so it's not changing the old results. I think @jonwinkler and @jmaguire1 had some discussions around this, I'll leave it to them for more details.

HPXMLtoOpenStudio/resources/hvac.rb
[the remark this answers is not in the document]
shorowit: I thought this was going to be a temporary change just to isolate the impact, not that we were actually going to propose this change for the simple model and get different results. I'm open to changing it for the simple model, but if it has a significant impact, I want to be confident it's an improvement.

workflow/tests/base_results/results_simulations_energy.csv
yzhou601: Yes I think this is something we should discuss, I'm pretty sure the ~20% increase is caused by the PLR curve we introduced, I was trying to justify the diff between "simple" and "advanced", so I made them as consistent as possible except for they're using two different E+ coils.

image

This is the results I got from simple model + reverting the PLR curve change, current simple model, and current advanced model.

HPXMLtoOpenStudio/resources/hvac.rb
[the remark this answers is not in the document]
jmaguire1: A little more info/context on what we did here: there isn't any sort of published cycling information for water to air heat pumps in the US analogous to c_d (according to Jon W and Jeff Munk). However, there's a European standard for this that defines a similar cycling coefficient C_c. Similar to c_d in the U.S., there's a minimum value that gets assumed if c_d isn't reported/tested. For C_c, that minimum value is 0.9 (for c_d, my understanding is it's 0.25). The curve we're using here is derived from using this assumed worst case scenario of 0.9, then fitting a curve to the PLF/PLR relationship (since there isn't an existing E+ curve object that exactly fits the relationship).

I think this probably better than just assuming the EnergyPlus default relationship, since it's at least based on some data. Since this is a European standard I'm not sure how easy it will be to come up with better estimates for what C_c might be, but if we were able to find some information like that we could potentially update the curve to reflect a different value.

image


This is all based on the paper in the comment: https://www.e3s-conferences.org/articles/e3sconf/pdf/2018/19/e3sconf_eko-dok2018_00139.pdf

jmaguire1:
@shorowit: I think this is now where we want to land in terms of PLR curves, no impact on the base case because it's still a little uncertain and impactful but we'll keep it for ResStock and document this as being a potential known issue and/or how exactly we're doing our PLR curves here.

If you want to make any other changes (like making this an undocumented feature or changing some of the names around so as not to mislead anyone), let me and @yzhou601 know.

Changelog.md
```diff
@@ -2,7 +2,8 @@
 
 __New Features__
 - HVAC modeling updates:
-  - **Breaking Change**: `CompressorType` required for central and mini-split air conditioners and heat pumps.
+  - **Breaking Change**: `CompressorType` required for central and mini-split air conditioners and heat pumps as well as ground-to-air heat pumps.
+  - Optional input `SimulationControl/AdvancedResearchFeatures/GroundToAirHeatPumpModelType` to choose "simple" (default) or "advanced"; "advanced" ground-to-air heat pump model better accounts for coil staging.
```
shorowit, line 6: How about if we change "advanced" to "experimental"? That would be a small change to the code, and I think it would be a sufficient warning to a user that they probably shouldn't choose this yet unless they have a good reason for it.

docs/source/workflow_inputs.rst
```diff
@@ -109,7 +109,7 @@ These features may require shorter timesteps, allow more sophisticated simulatio
   ``DefrostModelType``                    string             See [#]_          No        standard  Defrost model type for air source heat pumps [#]_
   ``OnOffThermostatDeadbandTemperature``  double    F        > 0 [#]_          No                  Temperature difference between cut-in and cut-out temperature for HVAC operation [#]_
   ``HeatPumpBackupCapacityIncrement``     double    Btu/hr   > 0 [#]_          No                  Capacity increment of multi-stage heat pump backup systems [#]_
-  ``GroundToAirHeatPumpModelType``        string             See [#]_          No        simple    Ground-to-air heat pump sytem model type [#]_
+  ``GroundToAirHeatPumpModelType``        string             See [#]_          No        standard  Ground-to-air heat pump system model type [#]_
```
yzhou601, line 112: Sorry for this 🤦

shorowit: That's what reviews are for 🙂

shorowit (approved):
It looks good to me. Thanks @yzhou601 and @jmaguire1! Will merge assuming the CI is happy with the latest changes.

PR #1879, opened 2024-11-09
Update HVAC models per RESNET HVAC addendum

### LINKED ISSUE #1871: Update HP models per RESNET HVAC addendum
See https://www.resnet.us/about/standards/minhers/draft-pds-01-minhers-addendum-82-dx-ac-hp-modeling/

For example:
- [x] @yzhou601 Updated rated cfm/ton, watts/cfm assumptions, expose FanModelType input/assumptions https://github.com/NREL/OpenStudio-HPXML/pull/1879
- [x] @yzhou601 Check/update sensitivity to indoor conditions
- [x] @yzhou601 Update defaulting for variable-speed systems; introduce rated/nominal speed inputs
  - [x] Address capacity retention inputs https://github.com/NREL/OpenStudio-HPXML/pull/1931
  - [x] Allow HeatingCapacity17F to be defaulted (only required for ERI)
  - [x] Require rated/nominal inputs (in addition to minimum/maximum speeds) for detailed performance data approach
  - [x] Documentation, etc.
- [x] @yzhou601 Updates to defrost modeling
  - [x] Add new EnergyPlus actuators for frost capacity/power multipliers
- [x] @shorowit Update SHR https://github.com/NREL/OpenStudio-HPXML/pull/1968
- [x] @shorowit Update 1-speed/2-speed models, allow detailed performance data https://github.com/NREL/OpenStudio-HPXML/pull/1936 
  - [x] Update airflow rates at different ODB temperatures
- [x] @shorowit Add measured airflow rate input https://github.com/NREL/OpenStudio-HPXML/pull/1979
- [x] @shorowit Reconsider hvac_sizing.rb default airflow rates https://github.com/NREL/OpenStudio-HPXML/pull/1964
- [x] @shorowit Add optional EER/EER2 inputs https://github.com/NREL/OpenStudio-HPXML/pull/1897
- [x] @shorowit Update crankcase heating power
- [x] @shorowit Data extrapolation (net capacity and input power instead of capacity and cop) #1904 
- [x] @shorowit Introduce pan heaters https://github.com/NREL/OpenStudio-HPXML/pull/1932
  - [ ] Default to "continuous" or "defrost mode" control?
- [ ] Update OS-ERI
  - [x] Pan heaters
  - [x] Lockout/switchover temperatures
  - [x] Ensure supplemental heating to prevent unmet load (but don't operate during defrost)
  - [ ] Updated HERS HVAC acceptance criteria

yzhou601:
## Pull Request Description

closes #1871 

RESNET spreadsheet: [NEEP-Statistical-Model.zip](https://github.com/user-attachments/files/19475925/NEEP-Statistical-Model.zip) [Updated 3/26/2025]

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -2105,20 +2105,70 @@ def self.apply_hvac(runner, hpxml_bldg, weather, convert_shared_systems, unit_nu
...
-    # Fan power
+    # Fan model type
+    hpxml_bldg.heating_systems.each do |heating_system|
+      next unless heating_system.heating_system_type == HPXML::HVACTypeFurnace
+      next unless heating_system.fan_model_type.nil?
+      next if (not heating_system.distribution_system.nil?) && (heating_system.distribution_system.air_type == HPXML::AirTypeGravity)
+
+      if (not heating_system.attached_cooling_system.nil?) && (not heating_system.attached_cooling_system.compressor_type.nil?)
+        # Based on RESNET DX Modeling Appendix
+        heating_system.fan_model_type = (heating_system.attached_cooling_system.compressor_type == HPXML::HVACCompressorTypeSingleStage) ? HPXML::HVACFanModelTypePSC : HPXML::HVACFanModelTypeBPM
```
yzhou601, line 2116: For attached heating and cooling systems, grab the cooling compressor type first to default, otherwise follow the previous HEScore assumption (based on efficiency)

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -2105,20 +2105,70 @@ def self.apply_hvac(runner, hpxml_bldg, weather, convert_shared_systems, unit_nu
...
+
+      if (not cooling_system.attached_heating_system.nil?) && (not cooling_system.attached_heating_system.fan_model_type.nil?)
+        cooling_system.fan_model_type = cooling_system.attached_heating_system.fan_model_type
+        cooling_system.fan_model_type_isdefaulted = true
+      elsif [HPXML::HVACTypeCentralAirConditioner].include? cooling_system.cooling_system_type
+        # Based on RESNET DX Modeling Appendix
+        cooling_system.fan_model_type = (cooling_system.compressor_type == HPXML::HVACCompressorTypeSingleStage) ? HPXML::HVACFanModelTypePSC : HPXML::HVACFanModelTypeBPM
+        cooling_system.fan_model_type_isdefaulted = true
+      elsif [HPXML::HVACTypeMiniSplitAirConditioner].include? cooling_system.cooling_system_type
+        cooling_system.fan_model_type = HPXML::HVACFanModelTypeBPM
```
yzhou601, line 2134: Always assume BPM for minisplit, as they're variable speed systems?

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -2105,20 +2105,70 @@ def self.apply_hvac(runner, hpxml_bldg, weather, convert_shared_systems, unit_nu
...
+    end
+    hpxml_bldg.heat_pumps.each do |heat_pump|
+      next unless heat_pump.fan_model_type.nil?
+
+      if [HPXML::HVACTypeHeatPumpAirToAir].include? heat_pump.heat_pump_type
+        # Based on RESNET DX Modeling Appendix
+        heat_pump.fan_model_type = (heat_pump.compressor_type == HPXML::HVACCompressorTypeSingleStage) ? HPXML::HVACFanModelTypePSC : HPXML::HVACFanModelTypeBPM
+        heat_pump.fan_model_type_isdefaulted = true
+      elsif [HPXML::HVACTypeHeatPumpGroundToAir].include? heat_pump.heat_pump_type
+        if heat_pump.heating_efficiency_cop > 8.75 / 3.2 # HEScore assumption
```
yzhou601, line 2148: This is kept the same as previous assumptions, need to discuss when we have two speed/var speed GSHP capabilities. Not sure for that case should it follow the RESNET assumption dependent on speed number or should it be something else (out of RESNET scope).

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2537,21 +2539,32 @@ def self.create_supply_fan(model, obj_name, fan_watts_per_cfm, fan_cfms)
...
+    if hvac_system.fan_model_type.nil?
+      # Cubic relationship fan power curve
+      fan_power = max_fan_power * (fan_ratio**3)
+    elsif hvac_system.fan_model_type == HPXML::HVACFanModelTypeBPM
+      # bpm fan
+      index = hvac_system.distribution_system_idref.nil? ? 3 : 2.75
+      fan_power = max_fan_power * (fan_ratio**index)
+    else
+      # psc fan
+      fan_power = max_fan_power * fan_ratio * (0.3 * fan_ratio + 0.7)
```
yzhou601, line 2565: Different fan power equations

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4489,21 +4511,19 @@ def self.calc_ceer_from_eer(cooling_system)
   def self.set_fan_power_rated(hvac_system, use_eer_cop)
     hvac_ap = hvac_system.additional_properties
 
+    # Based on RESNET DX Modeling Appendix
+    psc_ducted_watts_per_cfm = 0.414 # W/cfm, PSC fan
+    psc_ductless_watts_per_cfm = 0.414 # W/cfm, PSC fan
```
yzhou601, line 4516: So far as I can tell, psc_ductless_watts_per_cfm is not used for any case. Ductless minisplit is always var speed and defaulted to bpm fans. Room AC and PTHPs are outside the scope.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5076,7 +5096,7 @@ def self.calculate_heat_pump_defrost_load_power_watts(heat_pump, unit_multiplier
     end
     # cooling capacity and airflow are already with unit multiplier, calculate the capacity w/o multiplier
     nominal_cooling_capacity = heat_pump.cooling_capacity / unit_multiplier
-    defrost_power_fraction = (max_heating_airflow / design_airflow)**3
+    defrost_power_fraction = calculate_fan_power_from_curve(1.0, max_heating_airflow / design_airflow, heat_pump)
```
yzhou601, line 5099: This should call the calculate_fan_power_from_curve in the first place, so I captured it here.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1069,6 +1069,8 @@
       <sch:assert role='ERROR' test='count(h:FractionHeatLoadServed) = 1 or count(../h:HeatPump/h:BackupSystem) &gt;= 1'>Expected 1 element(s) for xpath: FractionHeatLoadServed</sch:assert> <!-- See [HeatingSystem=HeatPumpBackup] -->
       <sch:assert role='ERROR' test='count(h:extension/h:FanPowerWattsPerCFM) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/FanPowerWattsPerCFM</sch:assert>
       <sch:assert role='ERROR' test='number(h:extension/h:FanPowerWattsPerCFM) &gt;= 0 or not(h:extension/h:FanPowerWattsPerCFM)'>Expected extension/FanPowerWattsPerCFM to be greater than or equal to 0</sch:assert>
+      <sch:assert role='ERROR' test='count(h:extension/h:FanModelType) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/FanModelType</sch:assert>
+      <sch:assert role='ERROR' test='h:extension/h:FanModelType[text()="psc" or text()="bpm"] or not(h:extension/h:FanModelType)'>Expected extension/FanModelType to be 'psc' or 'bpm'</sch:assert>
```
shorowit, line 1073: @yzhou601 This should be FanMotorType, not FanModelType. Please update this as as well as the hpxml.rb variable names and documentation.

yzhou601: Oops, good catch!

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -2238,21 +2278,42 @@ def self.apply_hvac(runner, hpxml_bldg, weather, convert_shared_systems, unit_nu
...
+        # FIXME: Review compared to AHRI data
+        cop = 0.2692 * seer + 0.2706 # Regression based on inverse model
+      when HPXML::HVACCompressorTypeTwoStage
+        # FIXME: Review compared to AHRI data
+        cop = 0.2773 * seer - 0.0018 # Regression based on inverse model
+      when HPXML::HVACCompressorTypeVariableSpeed
+        # FIXME: Need to develop something here based on AHRI/NEEP data
+        cop = 4.5
+      end
+      hvac_system.cooling_efficiency_eer = UnitConversions.convert(cop, 'W', 'Btu/hr').round(2)
```
shorowit, line 2304: I will look at AHRI or ENERGY STAR database to derive new defaults for EER2 (as a function of SEER2 and # speeds). Also need to check whether these values are being used as net or gross values.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -2532,21 +2525,32 @@ def self.create_supply_fan(model, obj_name, fan_watts_per_cfm, fan_cfms)
...
-    # Cubic relationship fan power curve
-    return max_fan_power * (fan_ratio**3)
+  # @param max_fan_power [Double] Rated fan power consumption
+  # @param fan_ratio [Double] Fan cfm ratio to full speed
+  # @param hvac_system [HPXML::HeatingSystem or HPXML::CoolingSystem or HPXML::HeatPump] The HPXML HVAC system of interest
+  # @return [Double] Fan power at any speed or mode
+  def self.calculate_fan_power_from_curve(max_fan_power, fan_ratio, hvac_system)
+    if hvac_system.fan_motor_type.nil?
+      # Cubic relationship fan power curve
+      fan_power = max_fan_power * (fan_ratio**3)
```
shorowit, line 2544: What are the cases where the fan motor type is unknown and not defaulted? I guess I thought we'd always be defaulting upstream, in which case this code can go away because it's also PSC or BPM.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1336,11 +1343,14 @@
...
       <sch:assert role='ERROR' test='count(h:CoolingDetailedPerformanceData) &lt;= 1'>Expected 0 or 1 element(s) for xpath: CoolingDetailedPerformanceData</sch:assert> <!-- See [CoolingDetailedPerformanceData]-->
       <sch:assert role='ERROR' test='count(h:IntegratedHeatingSystemFuel) = 0'>Expected 0 element(s) for xpath: IntegratedHeatingSystemFuel</sch:assert>
       <sch:assert role='ERROR' test='count(h:extension/h:FanPowerWattsPerCFM) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/FanPowerWattsPerCFM</sch:assert>
       <sch:assert role='ERROR' test='number(h:extension/h:FanPowerWattsPerCFM) &gt;= 0 or not(h:extension/h:FanPowerWattsPerCFM)'>Expected extension/FanPowerWattsPerCFM to be greater than or equal to 0</sch:assert>
+      <sch:assert role='ERROR' test='count(h:extension/h:FanMotorType) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/FanMotorType</sch:assert>
+      <sch:assert role='ERROR' test='h:extension/h:FanMotorType[text()="BPM"] or not(h:extension/h:FanMotorType)'>Expected extension/FanMotorType to be 'BPM'</sch:assert>
```
shorowit, line 1353: Unless we are 100% sure there are no mini-splits w/ PCM motors, we should probably allow it as an option. We can still default to BPM, but it would allow user flexibility.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: This method is called by `create_supply_fan` where more system types are called, e.g. evap cooler, water loop to air, boiler, unit heater.

shorowit: All of those should have `fan_ratio==1`, no? In which case this equation is confusing because we aren't really using a cubic relationship.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
shorowit: The docs need to be updated too, not just EPvalidator.xml

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: That's right, good point. So that means actually these relationship equations are the same for `fan_ratio==1`, we can use either of them to yield the same results. These equations are to calculate the fan power at any fan speed, these system types are the cases when there's only one speed being modeled. Is the question that specifying 1 speed in E+ doesn't actually model cubic relationship while in the comment we says so? Should we add some comments to clarify?

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
yzhou601: Interesting that the docs do have both as options for mini-splits.

shorowit: Oh, true. We are good here then.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: In my opinion, the best thing would be to simply use `fan_power = max_fan_power` when the fan motor type is nil, as that is very explicit as to how we're handling it. I would also suggest adding a `fail` error message if fan_ratio is not 1, since we don't expect (nor handle) that situation.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3023,22 +3030,20 @@ def self.add_data_point_adaptive_step_size(data_array, mode, tol = 0.1)
...
   # @return [nil]
   def self.correct_ft_cap_eir(data_array, mode)
     # Add sensitivity to indoor conditions
     # single speed cutler curve coefficients
     if mode == :clg
-      cap_ft_spec_ss, eir_ft_spec_ss = get_cool_cap_eir_ft_spec(HPXML::HVACCompressorTypeSingleStage)
       rated_t_i = HVAC::AirSourceCoolRatedIWB
-      indoor_t = [50.0, rated_t_i, 80.0]
+      # Added two data points to be held constant outside the range 57F to 72F
+      indoor_t = [40.0, 57.0, rated_t_i, 72.0, 90.0]
```
shorowit, line 3041: Do we really need to manually hold things constant beyond 57 and 72? When we create the TableLookup object, we set the extrapolation method to "constant" (see [here](https://github.com/NREL/OpenStudio-HPXML/blob/5bdc5548e4ea54640e8c7d438af9a5a3bb4ec35b/HPXMLtoOpenStudio/resources/hvac.rb#L2781)), so wouldn't that already cover this?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -3062,19 +3067,24 @@ def self.correct_ft_cap_eir(data_array, mode)
           data_tmp << dp_new
           if mode == :clg
             dp_new.indoor_wetbulb = t_i
+            # Cooling variations  shall be held constant for Tiwb less than 57°F and greater than 72°F, and for Todb less than 75°F
+            curve_t_i = [[t_i, 57].max, 72].min
+            curve_t_o = [dp_new.outdoor_temperature, 75].max
```
shorowit, line 3072: Question above applies to this too.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1746,13 +1746,16 @@
     <sch:title>[HeatingDetailedPerformanceDataPointWithCapacity]</sch:title>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:HeatingDetailedPerformanceData/h:PerformanceDataPoint/h:Capacity'>
       <sch:assert role='ERROR' test='count(../../../h:HeatingCapacity) = 1'>Expected 1 element(s) for xpath: ../../../HeatingCapacity</sch:assert>
+      <sch:assert role='ERROR' test='number(../../../h:HeatingCapacity) = number(../../h:PerformanceDataPoint[h:OutdoorTemperature=47 and h:CapacityDescription="nominal"]/h:Capacity)'>Expected ../../../HeatingCapacity to be equal to ../../PerformanceDataPoint[OutdoorTemperature=47 and CapacityDescription="nominal"]/Capacity</sch:assert>
+      <sch:assert role='ERROR' test='(number(../../../h:HeatingCapacity17F) = number(../../h:PerformanceDataPoint[h:OutdoorTemperature=17 and h:CapacityDescription="nominal"]/h:Capacity)) or not (../../../h:HeatingCapacity17F)'>Expected ../../../HeatingCapacity17F to be equal to ../../PerformanceDataPoint[OutdoorTemperature=17 and CapacityDescription="nominal"]/Capacity</sch:assert>
```
shorowit, line 1750: Good thinking. Ideally we would now remove the requirement that `HeatingCapacity` and `CoolingCapacity` is required (and update the docs) since we get the same data from the detailed performance data. No point requiring the same value to be specified two different places. In defaults.rb, we can populate `HeatingCapacity` and `CoolingCapacity` from the detailed performance datapoints. And then all the asserts would be similar to the `HeatingCapacity17F` assert and handle the fact that the input is optional (e.g., ` or not (../../../h:HeatingCapacity17F)`).

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1688,34 +1710,50 @@
...
-      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature!=95 and h:CapacityDescription="maximum"]) &gt;= 1'>Expected 1 or more element(s) for xpath: PerformanceDataPoint[OutdoorTemperature!=95 and CapacityDescription="maximum"]</sch:assert> <!-- See [PerformanceDataPoint] -->
+      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature=82 and h:CapacityDescription="minimum"]) = 1'>Expected 1 element(s) for xpath: PerformanceDataPoint[OutdoorTemperature=82 and CapacityDescription="minimum"]</sch:assert> <!-- See [PerformanceDataPoint] -->
+      <sch:assert role='ERROR' test='count(h:PerformanceDataPoint[h:OutdoorTemperature=82 and h:CapacityDescription="nominal"]) = 1'>Expected 1 element(s) for xpath: PerformanceDataPoint[OutdoorTemperature=82 and CapacityDescription="nominal"]</sch:assert> <!-- See [PerformanceDataPoint] -->
```
shorowit, line 1739: There is no nominal/rated 82F datapoint, we can't require this. Let's plan to discuss. (If we update this, also need to update docs and changelog.)

BuildResidentialHPXML/measure.rb
```diff
@@ -1584,7 +1584,7 @@ def arguments(model) # rubocop:disable Lint/UnusedMethodArgument
 
     arg = OpenStudio::Measure::OSArgument.makeStringArgument('hvac_perf_data_heating_outdoor_temperatures', false)
     arg.setDisplayName('HVAC Detailed Performance Data: Heating Outdoor Temperatures')
-    arg.setDescription('Outdoor temperatures of heating detailed performance data if available. Applies only to variable-speed air-source HVAC systems (central air conditioners, mini-split air conditioners, air-to-air heat pumps, and mini-split heat pumps). One of the outdoor temperatures must be 47 F. At least two performance data points are required using a comma-separated list.')
+    arg.setDescription('Outdoor temperatures of heating detailed performance data if available. Applies only to variable-speed air-source HVAC systems (central air conditioners, mini-split air conditioners, air-to-air heat pumps, and mini-split heat pumps). Outdoor temperatures must be 47F, 17F, and 5F (and one optional temperature less than 5F). At least two performance data points are required using a comma-separated list.')
```
shorowit, line 1587: Note to self: need to update all of these out-of-date descriptions ("At least two performance data points are required using a comma-separated list").

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
yzhou601: I found that issue today as well. Currently made up some numbers there but we need to handle these datapoints as optional inputs

yzhou601:
CI is complaining about the BuildResidentialHPXML test:  test_workflows at the `extra-second-heating-system-fireplace-to-heat-pump.xml`. I checked the error and generated hpxml file, it's because the HeatingCapacity17F input is not consistent with 17F nominal datapoint in detailed performance data. But I can't find the place where the HeatingCapacity17F is set. @joseph-robertson @shorowit can you help point out the place?

HPXMLtoOpenStudio/tests/test_defaults.rb
```diff
@@ -2184,6 +2184,81 @@ def test_air_source_heat_pumps
...
+    assert_in_epsilon(7500.0, nom_dp_17f.capacity, 0.01)
+    assert_in_epsilon(3127.5, min_dp_17f.capacity, 0.01)
+    assert_in_epsilon(1.66, max_dp_17f.efficiency_cop, 0.01)
+    assert_in_epsilon(1.84, nom_dp_17f.efficiency_cop, 0.01)
+    assert_in_epsilon(2.08, min_dp_17f.efficiency_cop, 0.01)
+    # 5F
+    assert_in_epsilon(7950.8, max_dp_5f.capacity, 0.01)
+    assert_in_epsilon(7857.4, nom_dp_5f.capacity, 0.01)
+    assert_in_epsilon(2552.7, min_dp_5f.capacity, 0.01)
+    # The RESNET table seems to be incorrect in calculating the Net Power @ 5F,
```
yzhou601, line 2232: @shorowit @jonwinkler Probably we should let Neal know this error? It's the calculation in spreadsheet.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1696,34 +1718,50 @@
...
     </sch:rule>
   </sch:pattern>
 
   <sch:pattern>
     <sch:title>[CoolingDetailedPerformanceDataPointWithCapacity]</sch:title>
     <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:CoolingDetailedPerformanceData/h:PerformanceDataPoint/h:Capacity'>
-      <sch:assert role='ERROR' test='count(../../../h:CoolingCapacity) = 1'>Expected 1 element(s) for xpath: ../../../CoolingCapacity</sch:assert>
+      <sch:assert role='ERROR' test='(number(../../../h:CoolingCapacity) = number(../../h:PerformanceDataPoint[h:OutdoorTemperature=95 and h:CapacityDescription="nominal"]/h:Capacity)) or not (../../../h:CoolingCapacity)'>Expected ../../../CoolingCapacity to be equal to ../../PerformanceDataPoint[OutdoorTemperature=95 and CapacityDescription="nominal"]/Capacity</sch:assert>
     </sch:rule>
```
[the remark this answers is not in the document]
yzhou601: Good idea, working on this now.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
```diff
@@ -1752,16 +1752,19 @@
...
-    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:HeatingDetailedPerformanceData/h:PerformanceDataPoint/h:Capacity'>
-      <sch:assert role='ERROR' test='(number(../../../h:HeatingCapacity) = number(../../h:PerformanceDataPoint[h:OutdoorTemperature=47 and h:CapacityDescription="nominal"]/h:Capacity)) or not (../../../h:HeatingCapacity)'>Expected ../../../HeatingCapacity to be equal to ../../PerformanceDataPoint[OutdoorTemperature=47 and CapacityDescription="nominal"]/Capacity</sch:assert>
-      <sch:assert role='ERROR' test='(number(../../../h:HeatingCapacity17F) = number(../../h:PerformanceDataPoint[h:OutdoorTemperature=17 and h:CapacityDescription="nominal"]/h:Capacity)) or not (../../../h:HeatingCapacity17F)'>Expected ../../../HeatingCapacity17F to be equal to ../../PerformanceDataPoint[OutdoorTemperature=17 and CapacityDescription="nominal"]/Capacity</sch:assert>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:HVAC/h:HVACPlant/*/h:HeatingDetailedPerformanceData/h:PerformanceDataPoint[number(h:OutdoorTemperature)=47 and h:CapacityDescription="nominal"]/h:Capacity'>
```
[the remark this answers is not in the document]
shorowit: A couple suggestions. Remove the trailing `h:Capacity` from the context, then you can remove one level of navigating up the tree ("..") for all the elements. And use a separate `sch:rule` for the 17F datapoint, it doesn't make sense to look for errors related to the 17F datapoint just because the 47F datapoint is present; rather, look for errors related to the 17F datapoint when the 17F datapoint is present.

Thanks for adding a test!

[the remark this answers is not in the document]
yzhou601: The `h:capacity` is necessary because there's use case when capacity fraction of nominal is provided, then we don't want to perform the capacity check, or we need to check the fractions are 1.0 at nominal. The second suggestion is good, I will change it.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
shorowit: The 5F nominal/rated datapoint should also be optional.

docs/source/workflow_inputs.rst
```diff
@@ -2770,14 +2788,16 @@ Each air-to-air heat pump is entered as a ``/HPXML/Building/BuildingDetails/Syst
...
   ``AnnualCoolingEfficiency[Units="SEER" or Units="SEER2"]/Value``  double   Btu/Wh    > 0                       Yes                       Rated cooling efficiency [#]_
+  ``AnnualCoolingEfficiency[Units="EER" or Units="EER2"]/Value``    double   Btu/Wh    > 0                       No        See [#]_        Rated cooling efficiency [#]_
   ``AnnualHeatingEfficiency[Units="HSPF" or Units="HSPF2"]/Value``  double   Btu/Wh    > 0                       Yes                       Rated heating efficiency [#]_
   ``CoolingDetailedPerformanceData``                                element                                      No        <none>          Cooling detailed performance data [#]_
   ``HeatingDetailedPerformanceData``                                element                                      No        <none>          Heating detailed performance data [#]_
   ``extension/HeatingCapacityRetention[Fraction | Temperature]``    double   frac | F  >= 0, < 1 | <= 17         No        See [#]_        Heating output capacity retention at cold temperature [#]_
```
shorowit, line 2795: We will replace `extension/HeatingCapacityRetention[Fraction | Temperature]` with `extension/HeatingCapacityFraction17F` or some similar name. Update docs, changelog, etc.

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.xml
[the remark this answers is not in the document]
shorowit: Or you could include h:Capacity inside [] brackets, so the rule is only triggered if Capacity is present, but the rules are relative to the datapoint.

HPXMLtoOpenStudio/tests/test_defaults.rb
shorowit: Yueyue sent an email to Neal, who fixed the bug.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5213,6 +4646,103 @@ def self.apply_advanced_defrost(model, htg_coil, air_loop_unitary, conditioned_s
...
+      program.addLine("If (#{tout_db_sensor.name} <= #{UnitConversions.convert(32.0, 'F', 'C')})")
+    else # Don't run pan heater during heating unavailable period
+      program.addLine("If (#{tout_db_sensor.name} <= #{UnitConversions.convert(32.0, 'F', 'C')}) && (#{htg_avail_sensor.name} == 1)")
+    end
+    if heat_pump.pan_heater_control_type == HPXML::HVACPanHeaterControlTypeContinuous
+      program.addLine("  Set #{pan_heater_energy_oe_act.name} = #{heat_pump.pan_heater_watts}")
+    elsif heat_pump.pan_heater_control_type == HPXML::HVACPanHeaterControlTypeDefrost
+      # Set defrost fraction per RESNET MINHERS Addendum 82
+      # FIXME: Reuse this code in the defrost model
+      program.addLine("  Set F_defrost = 0.134 - (0.003 * ((#{tout_db_sensor.name} * 1.8) + 32))")
```
yzhou601, line 4736: @shorowit I think the RESNET equations are using degree F for outdoor air temperature, we should convert the tout_db_sensor to be degree F?

shorowit: @yzhou601 That's what `((#{tout_db_sensor.name} * 1.8) + 32)` does

yzhou601: Ahh that's right, sorry that I missed that.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4856,20 +4292,18 @@ def self.add_installation_quality_program(fault_program, tin_sensor, tout_sensor
...
       qgr_values, p_values, ff_chg_values = get_charge_fault_heating_coeff(f_chg)
       suffix = 'htg'
     end
-    fault_program.addLine("Set a1_AF_Qgr_#{suffix} = #{cap_fflow_spec[0][0]}")
-    fault_program.addLine("Set a2_AF_Qgr_#{suffix} = #{cap_fflow_spec[0][1]}")
-    fault_program.addLine("Set a3_AF_Qgr_#{suffix} = #{cap_fflow_spec[0][2]}")
-    fault_program.addLine("Set a1_AF_EIR_#{suffix} = #{eir_fflow_spec[0][0]}")
-    fault_program.addLine("Set a2_AF_EIR_#{suffix} = #{eir_fflow_spec[0][1]}")
-    fault_program.addLine("Set a3_AF_EIR_#{suffix} = #{eir_fflow_spec[0][2]}")
+    fault_program.addLine("Set a1_AF_Qgr_#{suffix} = #{hvac_ap.cool_cap_fflow_spec[0]}")
```
yzhou601, line 4301: Is it intended to use cooling air flow curve coefficients for both heating and cooling?

shorowit: Good catch, definitely a refactoring bug. Fixed in the last commit.

PR #1238, opened 2022-12-02
Issues around zero occupants, try 2

joseph-robertson:
## Pull Request Description

Closes #1079.

When number of occupants equals 0 (i.e., operational calculation), set a vacancy period identical to the simulation period.

Create sample files:
* `base-calctype-operational-residents-0.xml`
* `base-calctype-operational-residents-0-runperiod-1-month.xml`

This PR also extends vacancy periods to affect local ventilation (i.e., kitchen and bath fans).

Questions:
- [ ] Do we need a new sample file for 0 occupants when asset calculation?

## Checklist

PR Author: Check these when they're done. Not all may apply. ~~strikethrough~~ and check any that do not apply. 

PR Reviewer: Verify each has been completed.

- [ ] ~Schematron validator (`EPvalidator.xml`) has been updated~
- [x] Sample files have been added/updated (via `tasks.rb`)
- [ ] Unit tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests`)
- [x] Documentation has been updated
- [x] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [x] No unexpected changes to simulation results of sample files

PR #1899, opened 2024-12-17
Fixes whole MF building bug when zero occupants in one unit

shorowit:
## Pull Request Description

Fixes zero occupants specified for one unit in a whole MF building from being treated like zero occupants for every unit.

The [previous approach](https://github.com/NREL/OpenStudio-HPXML/pull/1238), in which we use a year-round vacancy period, was fine until we started modeling whole MF buildings. But it applies to the entire building, so that doesn't work for a MF building with units that have varying numbers of occupants. Now we explicitly zero out target end uses.

## Checklist

Not all may apply:


shorowit:
TODO: Add a unit test.

shorowit:
@joseph-robertson This gives identical results in ResStock, with one exception. As demonstrated in [this commit](https://github.com/NREL/resstock/pull/1333/commits/36dc782c6a54c1046765c8edd04085e19e24b26f), the previous implementation was correctly zeroing out ceiling fan electricity usage, but was not disabling the ceiling fan setpoint offset. So this PR will cause a small change in homes that have zero occupants, ceiling fans, and use a ceiling fan setpoint offset.

joseph-robertson (approved):

PR #1902, opened 2024-12-24
Add EER2 defaults

shorowit:
## Pull Request Description

Adds EER2 defaults based on regression of ENERGY STAR products, so that we can use this in our model going forward. Currently the model is still based on EER, so no results change (other than minor diffs due to unit conversions/rounding).

image

image

image

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1701,7 +1701,7 @@ def self.set_cool_curves_dx_air_source(cooling_system)
         clg_ap.cool_rated_airflow_rate = clg_ap.cool_rated_cfm_per_ton[0]
         clg_ap.cool_fan_speed_ratios = calc_fan_speed_ratios(clg_ap.cool_capacity_ratios, clg_ap.cool_rated_cfm_per_ton, clg_ap.cool_rated_airflow_rate)
         clg_ap.cool_cap_fflow_spec, clg_ap.cool_eir_fflow_spec = get_cool_cap_eir_fflow_spec(cooling_system.compressor_type)
-        clg_ap.cool_rated_cops = [UnitConversions.convert(cooling_system.cooling_efficiency_eer, 'Btu/hr', 'W')]
+        clg_ap.cool_rated_cops = [(0.2692 * cooling_system.cooling_efficiency_seer + 0.2706).round(2)] # Regression based on inverse model
```
yzhou601, line 1704: This is the previous seer -> cop conversion, I thought we're not going to keep it, but rather replacing it with the seer2->eer2 EnergyStar regression, and adding the net to gross conversion somewhere before setting E+ coil inputs. Any reason that we cannot use EER2 to convert to COP here?

shorowit: I thought it would be better to not have diffs here. Otherwise we'd have to update all the test cases for this one change, and then update them yet again when we implement the rest of the RESNET methodology. I think it's better to just update the code to use SEER2/EER2/HSPF2 all at once.

yzhou601: I see, so it's now using the previous conversion for COP, and the EER/EER2 is not yet used in the simulation, right?

shorowit: Yep, that's right. EER2 is now always generated (either from user input, converted from EER, or defaulted), but not yet used in the modeling.

yzhou601 (approved):
This looks good to me.

PR #1907, opened 2025-01-08
Stochastic schedule error-checking for `update_hpxmls` task

shorowit:
## Pull Request Description

Adds error-checking to the `update_hpxmls` task to ensure that when multiple HPXML sample files reference the same stochastic schedule CSV filename, the schedule each HPXML file generates is identical. This takes a little extra time, since we now regenerate schedule CSVs for all HPXMLs that reference stochastic schedules.

Previously we only generated the schedule CSV once for each unique filename, as this was faster. We assumed the generated schedule would always be the same for any HPXML file that references it. But that assumption isn't great, because, e.g., we could have generated the schedule for an HPXML file w/ 6 occupants and then applied it to an HPXML file w/ 3 occupants.

No errors were caught by this error-checking, but this is more robust and ensures we won't introduce errors like this in the future as we make changes to the BuildResidentialScheduleFile measure.

## Checklist

Not all may apply:


tasks.rb
```diff
@@ -72,14 +75,29 @@ def create_hpxmls
...
+        # call the BuildResidentialScheduleFile measure below, it will generate a stochastic schedule that matches
+        # the simulation timestep; so we'll avoid the call and just use the stochastic schedule generated from
+        # another HPXML file.
+        csv_path = nil
+      end
+      if not csv_path.nil?
         sch_args = { 'hpxml_path' => hpxml_path,
                      'output_csv_path' => csv_path,
                      'hpxml_output_path' => hpxml_path,
                      'building_id' => "MyBuilding#{suffix}" }
```
shorowit, line 92: If I add `'schedules_random_seed' => hpxml_i`, I verified that an error is indeed thrown because now the schedules generated for a given CSV filename are different.

HPXMLtoOpenStudio/resources/schedule_files/README.md
```diff
@@ -0,0 +1,4 @@
+All of the stochastic occupancy schedules are automatically regenerated by running:
```
rajeee, line 1:
```suggestion
All of the stochastic occupancy schedules (filenames that begin with 'occupancy-stochastic') are automatically regenerated by running:
```

tasks.rb
```diff
@@ -72,14 +75,29 @@ def create_hpxmls
...
+      csv_path = json_input['schedules_filepaths'].to_s.split(',').map(&:strip).find { |fp| fp.include? stochastic_sched_basename }
+      if hpxml_path.include?('base-schedules-detailed-occupancy-stochastic-10-mins.xml') ||
+         hpxml_path.include?('base-simcontrol-timestep-10-mins-occupancy-stochastic-60-mins.xml')
+        # These sample files test simulation timesteps that differ from the stochastic schedule timestep. If we
+        # call the BuildResidentialScheduleFile measure below, it will generate a stochastic schedule that matches
+        # the simulation timestep; so we'll avoid the call and just use the stochastic schedule generated from
+        # another HPXML file.
+        csv_path = nil
+      end
+      if not csv_path.nil?
```
rajeee, line 88: I think it might be better to define this list upfront.

```suggestion
      schedule_skip_list = [
          'base-schedules-detailed-occupancy-stochastic-10-mins.xml',
          'base-simcontrol-timestep-10-mins-occupancy-stochastic-60-mins.xml'
      ]  # This definition can be moved up along with the comment explaining which xml files should be here
      if csv_path and !schedule_skip_list.include?(File.basename(hpxml_path))
```

shorowit: Good suggestion, will do.

rajeee (approved):
Looks good to me. Thanks for accepting the suggestions!

PR #1928, opened 2025-02-11
Move BuildResidentialHPXML measure to option-based arguments

### LINKED ISSUE #1913: Move BuildResidentialHPXML measure to option-based arguments
The BuildResidentialHPXML measure is currently ~400 arguments and growing. That alone makes it difficult to use, but it's particularly challenging because there are many conditionally-required arguments (arguments that are only required based on the value of another argument). This limits its usefulness for some use cases.

This can be improved by largely moving from _property_-based arguments to _option_-based arguments. For example, the measure would replace ~10 clothes washer property arguments with a single dropdown of option names. A TSV file (e.g., `clothes washer.tsv`) will provide the mapping of an option name to its detailed properties.

(There is obviously some engineering judgment needed as to which properties are incorporated into the TSV. For the clothes washer example, Location and Usage Multiplier should probably remain separate arguments since they can apply to any clothes washer product. For HVAC systems, Capacity and Fraction Load Served would likely remain separate arguments too.)

There are several advantages to using option-based arguments:
- It will significantly reduce the number of arguments.
- It will largely remove the need for conditionally-required arguments. (This logic will be incorporated in the way options are described in the TSV.)
- It can help standardize options across different projects (e.g., ResStock, URBANopt, etc.).
- It removes any friction/resistance around introducing more advanced properties since they no longer directly show up in the measure (they are one level lower in a TSV).

If a user wants an additional option that's not offered, they can:
- Add it to their local copy of the TSV (and depending on how they use the measure, they may have to update the measure.xml prior to using the new option), or
- Open a PR or request a new option be formally added to OS-HPXML.

Sashadf1:
## Pull Request Description

closes #1913

## Checklist

Not all may apply:

### DIFF
--- a/BuildResidentialHPXML/README.md
+++ b/BuildResidentialHPXML/README.md
@@ -1150,29 +1150,16 @@ Assembly R-value for the ceiling (attic floor).
 
 <br/>
 
-**Roof: Material Type**
+**Enclosure: Roof Material**
 
-The material type of the roof. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#hpxml-roofs'>HPXML Roofs</a>) is used.
+The material type/color of the roof. If not provided, the OS-HPXML default (see <a href='https://openstudio-hpxml.readthedocs.io/en/v1.10.0/workflow_inputs.html#hpxml-roofs'>HPXML Roofs</a>) is used.
 
-- **Name:** ``roof_material_type``
+- **Name:** ``enclosure_roof_material``
 - **Type:** ``Choice``
 
 - **Required:** ``false``
 
-- **Choices:** `asphalt or fiberglass shingles`, `concrete`, `cool roof`, `slate or tile shingles`, `expanded polystyrene sheathing`, `metal surfacing`, `plastic/rubber/synthetic sheeting`, `shingles`, `wood shingles or shakes`
-
-<br/>
-
-**Roof: Color**
-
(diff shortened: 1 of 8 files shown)


BuildResidentialHPXML/resources/options/site_soil_type.tsv
```diff
@@ -0,0 +1,26 @@
...
+"Sand, Mixed"	"sand, mixed"		
+"Sand, Wet"	"sand, wet"		
+"Silt, Dry"	"silt, dry"		
+"Silt, Mixed"	"silt, mixed"		
+"Silt, Wet"	"silt, wet"		
+0.5 Conductivity		0.5	
+0.8 Conductivity		0.8	
+1.1 Conductivity		1.1	
+1.4 Conductivity		1.4	
+1.44 Conductivity			0.03
```
shorowit, line 22: @Sashadf1 Seems like something went wrong with the value here.

Longer term, I would hope we can remove this option since it's so close to the option right above. But it's fine for now.

Sashadf1: The 1.44 Btu/hr-ft-F Conductivity option should be correct. 

0.03 ft2/hr Diffusivity = 1.44 Btu/hr-ft-F Conductivity, since "If Conductivity not provided, defaults to Diffusivity / 0.0208 if Diffusivity provided, otherwise defaults based on SoilType and MoistureType per Table 1 of Ground Thermal Diffusivity Calculation by Direct Soil Temperature Measurement unless otherwise noted."

0.03 / 0.0208 = 1.4423 ~ 1.44

[source](https://openstudio-hpxml.readthedocs.io/en/latest/workflow_inputs.html#id91)

A 0.03 ft2/hr Diffusivity option is needed to support [this](https://github.com/NREL/OpenStudio-HPXML/blob/c0452c5dc6021c4f34c0b529060bb98f1a3223c6/workflow/hpxml_inputs.json#L2628) sample file.

PR #1932, opened 2025-02-17
Add heat pump pan heater inputs

shorowit:
## Pull Request Description

Branches off of #1931.

Allows optional pan heater inputs (`PanHeaterPowerWatts` and `PanHeaterControlType`) for central heat pumps and mini-splits; defaults to assuming a pan heater is present.

`PanHeaterControlType` can be 'continuous' or 'defrost mode'. If 'continuous', operates continuously when outdoor temperature is below 32F. If 'defrost mode', operates only during defrost mode when outdoor temperature is below 32F.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5423,6 +5429,102 @@ def self.apply_advanced_defrost(model, htg_coil, air_loop_unitary, conditioned_s
...
+    if htg_avail_sensor.nil?
+      program.addLine("If (#{tout_db_sensor.name} <= #{UnitConversions.convert(32.0, 'F', 'C')})")
+    else # Don't run pan heater during heating unavailable period
+      program.addLine("If (#{tout_db_sensor.name} <= #{UnitConversions.convert(32.0, 'F', 'C')}) && (#{htg_avail_sensor.name} == 1)")
+    end
+    if heat_pump.pan_heater_control_type == HPXML::HVACPanHeaterControlTypeContinuous
+      program.addLine("  Set #{pan_heater_energy_oe_act.name} = #{heat_pump.pan_heater_watts}")
+    elsif heat_pump.pan_heater_control_type == HPXML::HVACPanHeaterControlTypeDefrost
+      # Set defrost fraction per RESNET MINHERS Addendum 82
+      # FIXME: Reuse this code in the defrost model
```
shorowit, line 5509: A reminder that when we update the defrost model, we want to reference the same code both places.

ReportSimulationOutput/measure.rb
```diff
@@ -2041,11 +2042,26 @@ def get_report_meter_data_timeseries(meter_names, unit_conv, unit_adder, timeser
...
+    shift_values = {}
+    indexes.each_with_index do |_i, idx|
+      shift_values[idx] = false
+      if apply_ems_shift(timeseries_frequency)
+        if meter_names[idx].include? Constants::ObjectTypeWaterHeaterAdjustment
+          # Shift energy use adjustment to align with hot water energy use
+          shift_values[idx] = true
+        elsif meter_names[idx].include? Constants::ObjectTypePanHeater
+          # Shift energy use adjustment to align with HVAC operation and weather
+          shift_values[idx] = true
```
shorowit, line 2056: I verified the timeseries output is correct; pan heater energy use only shows up when temperatures are below 32F.

I also refactored this code a bit because I realized we were previously calling `apply_ems_shift(timeseries_frequency)` inside two nested loops, which is very inefficient. So now we do everything up front.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5347,10 +5351,11 @@ def self.apply_advanced_defrost(model, htg_coil, air_loop_unitary, conditioned_s
       comp_type_and_control: EPlus::EMSActuatorOtherEquipmentPower
     )
 
+    cnt = model.getOtherEquipments.count { |e| e.endUseSubcategory.start_with? Constants::ObjectTypeBackupSuppHeat } # Ensure unique meter for each heat pump
     defrost_supp_heat_energy_oe = Model.add_other_equipment(
       model,
       name: "#{air_loop_unitary.name} defrost supp heat energy",
-      end_use: Constants::ObjectTypeBackupSuppHeat,
+      end_use: "#{Constants::ObjectTypeBackupSuppHeat}#{cnt + 1}",
```
yzhou601, line 5358: Thanks for the improvements here! This makes me feel we should name it more explicitly if this is the only place to use the `ObjectTypeBackupSuppHeat`, something like `ObjectTypeDefrostBackupSuppHeat`?

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5411,6 +5417,102 @@ def self.apply_advanced_defrost(model, htg_coil, air_loop_unitary, conditioned_s
...
+        model,
+        name: "#{htg_coil.name} rtf s",
+        output_var_or_meter_name: 'Heating Coil Runtime Fraction',
+        key_name: htg_coil.name
+      )
+    end
+
+    # Create HVAC availability sensor
+    if not heating_unavailable_periods.empty?
+      htg_avail_sch = ScheduleConstant.new(model, 'heating availability schedule', 1.0, EPlus::ScheduleTypeLimitsFraction, unavailable_periods: heating_unavailable_periods)
```
yzhou601, line 5473: Not super familiar with unavailable periods, my understanding is that we need it here because we don't want to calculate the pan heater energy use in `continuous` mode if the hvac system is unavailable and outdoor temperature is below 32F. When `defrost mode`, the coil runtime fraction will be 0.0 during unavailable periods, so for advanced defrost EMS, we don't need to worry about the unavailable period, right? This approach looks good to me, just want to double check the advanced defrost is working well.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -5411,6 +5417,102 @@ def self.apply_advanced_defrost(model, htg_coil, air_loop_unitary, conditioned_s
...
+    end
+    program.addLine('Else')
+    program.addLine("  Set #{pan_heater_energy_oe_act.name} = 0.0")
+    program.addLine('EndIf')
+
+    # EMS calling manager
+    Model.add_ems_program_calling_manager(
+      model,
+      name: "#{program.name} calling manager",
+      calling_point: 'BeginTimestepBeforePredictor',
```
yzhou601, line 5511: Just a reminder that for this calling point, we'll be using the coil runtime fraction from last timestep. 1 time step lag is normal for EMS, so I'm not too worried about this. Though for advanced defrost EMS, I did use `InsideHVACSystemIterationLoop` for this reason, maybe it's not worth it.

yzhou601 (approved):
This PR looks good to me, good unit tests, good cleanups, and the results only increase in heating category because we introduced the pan heater heating energy, the `base-hvac-air-to-air-heat-pump-var-speed-pan-heater-none.xml` test file gave the same results as previous `base-hvac-air-to-air-heat-pump-var-speed.xml`. Good job, just a few comments that don't necessarily need to address/fix anything.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: Good idea. I will make this change.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: Yes, that exactly matches my thinking. I assume this is only needed for `continuous` mode option. But I will manually test a HPXML w/ the advanced defrost model and a power outage to double-check it's not needed there.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: Oh, yes, I should definitely change this. Thanks for pointing it out! I think I chose this calling point when I first implemented the `continuous` mode option and forgot that a different calling point would be better once I started using the coil RTF.

HPXMLtoOpenStudio/resources/hvac.rb
shorowit: Confirmed that it looks good, the defrost supplemental heating energy was zero during the power outage.

PR #1933, opened 2025-02-17
Fix stochastic schedule generation runtime

shorowit:
## Pull Request Description

Fixes a regression introduced [here](https://github.com/NREL/OpenStudio-HPXML/pull/1757) that resulted in very slow generation of stochastic schedules.

## Checklist

Not all may apply:


BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -194,7 +194,7 @@ def create_stochastic_schedules(args:,
     sinks = random_shift_and_normalize(sink_activity_sch, random_offset)
     baths = random_shift_and_normalize(bath_activity_sch, random_offset)
     fixtures = [showers, sinks, baths].transpose.map(&:sum)
-    @schedules[SchedulesFile::Columns[:HotWaterFixtures].name] = fixtures.map { |flow| flow / fixtures.max }
+    @schedules[SchedulesFile::Columns[:HotWaterFixtures].name] = normalize(fixtures)
```
shorowit, line 197: This is the primary bugfix. `fixtures.max` was being calculated within the iteration resulting in a huge performance hit.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -776,15 +797,15 @@ def get_ev_battery_schedule(away_schedule, hours_driven_per_year)
...
+        discharging_schedule.concat([0] * idle_time)           # Idle in the middle
+        discharging_schedule.concat([1] * second_half_driving) # End driving
+        charging_schedule.concat([0] * activity_minutes)
 
         driving_minutes_used += actual_driving_time
       else
-        charging_schedule += [1] * activity_minutes
-        discharging_schedule += [0] * activity_minutes
+        charging_schedule.concat([1] * activity_minutes)
+        discharging_schedule.concat([0] * activity_minutes)
```
shorowit, line 808: Specifically for EVs, this code was also resulting in some slowdown since `+=` creates new arrays and all of this is happening within a loop.

BuildResidentialScheduleFile/resources/schedules.rb
```diff
@@ -556,7 +558,8 @@ def get_ev_occupant_number(mkc_activity_schedules)
     end
 
     vehicle = @hpxml_bldg.vehicles[0]
-    hours_per_year = (vehicle.hours_per_week / 7) * UnitConversions.convert(1, 'yr', 'day')
+    hours_per_week = get_ev_hours_per_week(vehicle)
```
shorowit, line 561: Found a bug while working on all of this. `hours_per_week` is not a required input, so we need to default it when not provided.

I should see if we can apply defaults to the entire HPXML Building (or a copy of the object, if we need to prevent writing the defaults out to disk) to generically prevent issues like this in the future...

shorowit:
CI results:
- Before EV schedules: 96.6s
- After EV schedules: 230.2s
- This PR: 101.9s

I'm calling it good enough.

PR #1963, opened 2025-03-21
Update rubocop settings to match rake task

rajeee:
## Pull Request Description

There was a discrepancy between the settings in .rubocop.yml and the set of cops we were using [in the rake task](https://github.com/NREL/OpenStudio-HPXML/blob/12518133072823a3ce29e53505b8c5206864b145/tasks.rb#L2666). This PR updates the .rubocop.yml to use the same set of cops being used in rake task and removes hardcoded list.

## Checklist

Not all may apply:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,8 @@
 AllCops:
   DisabledByDefault: true
 
+Layout:
+  Enabled: True
 Layout/EndOfLine:
   Enabled: false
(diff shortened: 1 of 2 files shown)


shorowit (approved):
This might not be 100% identical to what we have before but it doesn't reformat the existing code and is clearly working catching things that I spot tested. I'll let the CI run through one more time just in case and then merge.

PR #1967, opened 2025-03-26
Make EV charging/discharging unavailable during vacancy periods

afontani:
## Pull Request Description

I was looking at the unavailable period table and what end-uses are unavailable during a vacancy period. I think the EV charging and discharging end-uses should be unavailable during the vacancy period.

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
+++ b/HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
@@ -36,5 +36,5 @@ bath_fan,1,1,0,0
 whole_house_fan,0,1,0,0
 battery_charging,0,1,0,0
 battery_discharging,0,1,0,0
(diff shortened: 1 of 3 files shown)


HPXMLtoOpenStudio/resources/data/unavailable_periods.csv
```diff
@@ -36,5 +36,5 @@ bath_fan,1,1,0,0
 whole_house_fan,0,1,0,0
```
shorowit, line 36: I agree with your change. Looking at the other end uses, it also seems like whole house fans should be unavailable during vacancies. Everything else looks reasonable to me.

```suggestion
whole_house_fan,1,1,0,0
```

shorowit (approved):
Ran quick tests for EVs and WHFs, results were impacted as expected. Thanks, @afontani.

PR #1674, opened 2024-04-04
Defrost Model Improvements

yzhou601:
## Pull Request Description

Allow an optional improved defrost model

**Breaking change**: Replaces `SimulationControl/TemperatureCapacitanceMultiplier` with `SimulationControl/AdvancedResearchFeatures/TemperatureCapacitanceMultiplier`.

## Checklist

PR Author: Check these when they're done. Not all may apply. ~~strikethrough~~ and check any that do not apply. 

PR Reviewer: Verify each has been completed.

- [x] Schematron validator (`EPvalidator.xml`) has been updated
- [x] Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)
- [x] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)
- [x] Documentation has been updated
- [x] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [ ] No unexpected changes to simulation results of sample files

PR #1987, opened 2025-04-18
RESNET Defrost Model Enhancement

yzhou601:
## Pull Request Description

Defrost models using RESNET approach, with frost multiplier EMS actuators from https://github.com/NREL/EnergyPlus/pull/10970. Require E+ 25.1 and OpenStudio 3.10. 
This PR may replace the advanced defrost model in #1674 .

## Checklist

Not all may apply:


yzhou601:
@shorowit  I spent some time looking into the E+ ZoneHVAC objects, I think the closest one to create an additional supplement heating system is [UnitHeater](https://bigladdersoftware.com/epx/docs/24-2/input-output-reference/group-zone-forced-air-units.html#zonehvacunitheater), while it still requires heating coil and fan objects to specified. I felt it's not a self-contained object like `OtherEquipment` objects, and harder to use EMS to override the controls and energy calculations. 

So I currently stopped there and kept our `OtherEquipment` approach, and fixed the reporting for delivered heating loads. But I'm open to further discussions around other options.

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -203,10 +204,20 @@ def self.apply_total_loads_ems_program(model, hpxml_osm_map, hpxml_header)
       end
 
       # Energy transferred in conditioned zone, used for determining heating (winter) vs cooling (summer)
+      # Add supp heat delivered energy to the meter, using Meter:Custom
+      htg_load_key_vars = [['', "Heating:EnergyTransfer:Zone:#{conditioned_zone_name.upcase}"]]
+      model.getOtherEquipments.sort.each do |o|
+        next unless o.endUseSubcategory.start_with? Constants::ObjectTypeHPDefrostSupplHeat
+        next unless o.space.get.thermalZone.get.name.to_s.upcase == conditioned_zone_name.upcase
+
+        htg_load_key_vars << [o.name.to_s.upcase, 'Other Equipment Total Heating Energy']
```
yzhou601, line 213: This doesn't seem working because the object level output variable doesn't incorporate thermal zone multiplier, need to revert this approach and try using sensors in the total_load EMS

shorowit: @yzhou601 There should be an energy meter that corresponds to each Equipment object (that would then incorporate the thermal zone multiplier). See [here](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportSimulationOutput/measure.rb#L3121-L3152) for example.

shorowit: In particular, [this part](https://github.com/NREL/OpenStudio-HPXML/blob/master/ReportSimulationOutput/measure.rb#L3145-L3151).

shorowit:
@yzhou601 will do three runs for standard E+ defrost model vs RESNET defrost model w/ backup vs RESNET defrost model w/o backup. If the results look good, we may not have to make any additional adjustments for the delivered heating load. (But perhaps we should document why somewhere.)

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -294,6 +306,9 @@ def self.apply_total_loads_ems_program(model, hpxml_osm_map, hpxml_header)
       if not dehumidifier_global_vars[unit].nil?
         program.addLine("  Set loads_htg_tot = loads_htg_tot - #{dehumidifier_global_vars[unit].name}")
       end
+      defrost_load_oe_sensors[unit].each do |defrost_ss|
+        program.addLine("  Set loads_htg_tot = loads_htg_tot + #{defrost_ss.name} * #{unit_multipliers[unit]}")
```
yzhou601, line 310: Try if this fixes the unit multiplier issue we had before, if CI passing I'm going to add the unit test (manually checked good).

HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -193,7 +193,8 @@ def self.apply_total_loads_ems_program(model, hpxml_osm_map, hpxml_header)
       duct_zone_names = unit_model.getThermalZones.select { |z| z.isPlenum }.map { |z| z.name.to_s }
       dehumidifier = unit_model.getZoneHVACDehumidifierDXs
       dehumidifier_name = dehumidifier[0].name.to_s unless dehumidifier.empty?
-      defrost_load_oes = unit_model.getOtherEquipments.select { |o| o.name.to_s.include? 'defrost heat load' }
+      # Spaces in other equipment names are replaced with '_' after merging models, include both names
+      defrost_load_oes = unit_model.getOtherEquipments.select { |o| (o.name.to_s.include? 'defrost heat load') || (o.name.to_s.include? 'defrost_heat_load') }
```
shorowit, line 197: @yzhou601 Using object names is not a robust way of determining object types (as evidence by the fact that you had to make a fix to the logic). Instead, you should set 'ObjectType' on the `additionalProperties` of the equipment object and check against that. We have lots of examples in the code of doing this, let me know if you want me to point you to one.

shorowit: Much cleaner, thanks!

docs/source/workflow_inputs.rst
```diff
@@ -117,7 +117,7 @@ These features may require shorter timesteps, allow more sophisticated simulatio
          Values greater than 1.0 have the effect of smoothing or damping the rate of change in the indoor air temperature from timestep to timestep.
          This heat capacitance effect is modeled on top of any other individual mass inputs (e.g., furniture mass, partition wall mass, interior drywall, etc.) in the HPXML.
   .. [#] DefrostModelType choices are "standard" and "advanced".
-  .. [#] Use "standard" for default E+ defrost setting.
+  .. [#] Use "standard" for defrost model per RESNET MINHERS Addendum 82, with integrated capacity and power approach.
```
yzhou601, line 120: Not sure if this is the best place to put the descriptions considering we're likely to remove the DefrostModelType input.

yzhou601:
This is overall in good shape, the last item is to add the unit test to verify supplemental systems don't impact heating load results. @shorowit Do you have suggestion which file is best for this? I can only think of util.rb where we parsed the simulation results/reports. But it seems to be general checks for all test files. Let me know if you have any thoughts.

shorowit:
> This is overall in good shape, the last item is to add the unit test to verify supplemental systems don't impact heating load results. @shorowit Do you have suggestion which file is best for this? I can only think of util.rb where we parsed the simulation results/reports. But it seems to be general checks for all test files. Let me know if you have any thoughts.

Is the idea that you want to run two simulations (one with backup during defrost and one without) and compare their results? If so, I guess I'd probably put it in `workflow/tests/test_other.rb`.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -4920,25 +4921,26 @@ def self.apply_advanced_defrost_ems_program(model, htg_coil, conditioned_space,
       calling_point: 'InsideHVACSystemIterationLoop',
       ems_programs: [program]
     )
+    return program
```
yzhou601, line 4924: Note: Update method @return description

shorowit (approved):

PR #1988, opened 2025-04-21
Refactor BuildResidentialScheduleFile measure to use the HPXML class

shorowit:
## Pull Request Description

It's always bothered me that we are doing direct XML manipulation rather than using the HPXML class. The latter is much cleaner/simpler. The original reason we went the direct XML manipulation path was in case we got an HPXML from a third-party software tool, we wanted to make sure that we don't lose any information not supported by the HPXML class when we write the updated HPXML. To address that, I now:
1. Create a backup of the original HPXML and give a warning message if the user has provided an output HPXML path that is the same as the input HPXML path.
2. Change the `run_simulation.rb` default so that it writes a new HPXML file rather than updating the original. (The template OSWs already did this. So unless someone uses their own OSWs where they deviate from our templates, they will not encounter the situation where the original HPXML is overwritten.)

## Checklist

Not all may apply:

### DIFF
--- a/BuildResidentialScheduleFile/measure.rb
+++ b/BuildResidentialScheduleFile/measure.rb
@@ -107,7 +107,8 @@ class BuildResidentialScheduleFile < OpenStudio::Measure::ModelMeasure
       hpxml_path = File.expand_path(hpxml_path)
     end
     unless File.exist?(hpxml_path) && hpxml_path.downcase.end_with?('.xml')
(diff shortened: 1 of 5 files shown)


shorowit:
FYI @joseph-robertson. I think this will result in an additional file `foo_bak.xml` for ResStock. I assume you can somehow delete that file during the ResStock workflow. But if this change really bothers you, I'm open to other ideas on how to handle this.

PR #1992, opened 2025-04-24
TankModelType=stratified bugfixes

shorowit:
## Pull Request Description

Fixes tank loss coefficient when TankModelType=stratified for a conventional storage water heater. See original bug [from the water heater GEB PR](https://github.com/NREL/OpenStudio-HPXML/blob/f19df1cafbc134713dd501e60295c27ae272b244/HPXMLtoOpenStudio/resources/waterheater.rb#L1663), where the U-factor was not converted to SI units.

Also adds error-checking to prevent TankModelType=stratified to be used for a non-electric water heater.

And some small code cleanup.

## Checklist

Not all may apply:


shorowit:
FYI @jmaguire1.

Mixed ERWH vs orig stratified ERWH vs new stratified ERWR:
image
image

Results between mixed and stratified results are now more similar.

And the new U-factor for the stratified tank (0.422 W/m2-K) is now more consistent with the UA for the mixed tank (0.7 W/K), given that the tank area is ~18 ft2 (~1.67 m2).

jmaguire1:
Results look reasonable to me, we'd expect slightly less energy use for the stratified tank because of the largely unheated volume below the lower element, so I think this is now working as intended. Thanks again for catching and fixing!

As far as why we can't model a stratified gas tank: we could, but it's not super exciting. The burner is always at the bottom of the tank, so since you're heating the entire volume the results aren't all that different than if you allowed for a mixed tank. There's maybe some differences in delivered energy with stochastic events (if say a 20 gal draw hits a 40 gal tank, the top is still hot in a stratified tank, not so much in mixed), but I didn't consider it a big enough priority . With electric resistance you have a dead volume at the bottom of the tank pretty consistently, and you have the upper element available for quicker recovery after larger draws, so there's a bit more difference in the results and it's a bit more interesting of a use case.

PR #1997, opened 2025-04-29
Refactor BuildResidentialScheduleFile measure to use the HPXML class, take 2

shorowit:
## Pull Request Description

In the BuildResidentialScheduleFile measure, we now smartly avoid creating a backup file when it's not needed (i.e., when the HPXML file we are updating was written by the HPXML class). This means that the ResStock workflow, for example, will not produce the backup file.

## Checklist

Not all may apply:


joseph-robertson (approved):
Nice. So a backup file is only written when the HPXML file's written contents (excluding the stochastic schedule line) won't match the original.

PR #2001, opened 2025-05-01
Pull master to resnet_heat_pump

yzhou601:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -6283,11 +6287,18 @@ def self.interpolate_seer2(seer2, eer2, seer2_array, seer2_eer2_ratio_array, cop
...
+      clg_ap.cool_cap_fflow_spec_iq = [0.718664047, 0.41797409, -0.136638137]
+      clg_ap.cool_eir_fflow_spec_iq = [1.143487507, -0.13943972, -0.004047787]
       clg_ap.cool_rated_shr_gross = 0.708
       clg_ap.cool_rated_cfm_per_ton = HVAC::RatedCFMPerTon
       return
     end
 
+    # Based on RESNET MINHERS Addendum 82
+    clg_ap.cool_cap_fflow_spec = [0.718664047, 0.41797409, -0.136638137]
+    clg_ap.cool_eir_fflow_spec = [1.143487507, -0.13943972, -0.004047787]
```
shorowit, line 6301: We shouldn't duplicate values in code, that's not good practice. If they ever change in the future, it'd be easy to change one set and not realize there's a second set. Also, it's pretty confusing that the IQ code in hvac.rb and hvac_sizing.rb only sometimes uses the `foo_iq` variables.

I suggest:
1. Set the `foo_iq` variable up front for all system types. Only use these variables in the IQ code in hvac.rb and hvac_sizing.rb.
2. Then set `foo = foo_iq` for all non-GHP system types. These variables will be used for the model.

shorowit (commented):
CI is passing and none of the existing results are changing. Looking good.

HPXMLtoOpenStudio/resources/hvac.rb
```diff
@@ -1571,46 +1748,209 @@ def self.get_cooling_setpoints(hpxml_bldg, hvac_control, has_ceiling_fan, year,
...
-    hp_ap = heat_pump.additional_properties
-
-    # E+ equation fit coil coefficients generated following approach in Tang's thesis:
-    # See Appendix B of  https://shareok.org/bitstream/handle/11244/10075/Tang_okstate_0664M_1318.pdf?sequence=1&isAllowed=y
-    # Coefficients generated by catalog data: https://files.climatemaster.com/Genesis-GS-Series-Product-Catalog.pdf, p180
-    # Data point taken as rated condition:
-    # EWT: 80F EAT:80/67F, AFR: 1200cfm, WFR: 4.5gpm
+  # @return [Array<Double, Double>] Cooling apacity ratios at each speed
+  def self.get_cool_capacity_ratios_gshp(heat_pump)
+    # For each speed, ratio of capacity to nominal capacity
```
shorowit, line 1756: I'm hoping we can remove the `get_cool_capacity_ratios_gshp`, `get_heat_capacity_ratios_gshp`, and `set_curves_gshp` methods here, and integrate all of that code into the `set_hvac_cooling_performance` and `set_hvac_heating_performance` methods in defaults.rb. Then air-source and ground-source systems will be pretty consistent.

It'd be nice to move the `set_gshp_assumptions` into defaults.rb too. And probably it should be renamed to `set_geothermal_loop_assumptions`, it seems to be specific to the geothermal loop and not the GHP.

HPXMLtoOpenStudio/resources/defaults.rb
yzhou601: This is good suggestion Scott! I was trying to resolve this and realized that it's not the best solution, wrote it to wish some quick fix to let it run. I'll change it so that it's not duplicated.

HPXMLtoOpenStudio/resources/hvac.rb
yzhou601: I got the point that you moved those methods for ashp to be in defaults.rb when resolving conflicts, but found the `set_curves_gshp` still there, not sure if it's intended because it's not using table lookup, or if it's something you want to resolve when pulling in the GHP work. I agree that it's better to be in defaults.rb, I'll give it a try.

yzhou601:
@shorowit I think it is ready for review, let me know if there's anything else to address.

shorowit (approved):

shorowit:
Thanks, @yzhou601!

PR #2026, opened 2025-06-18
Pull latest master branch to resnet heat pump branch

yzhou601:
## Pull Request Description

[description here]

## Checklist

Not all may apply:


shorowit:
@yzhou601 I found that the results are very good when you have either duct conduction losses by itself or duct leakage by itself, but once you have both, then the difference gets to be ~10%. I don't know why that is, but I assume it's due to the way we model duct losses. I highly doubt it's anything specific to our updated HVAC modeling. So I'm just going to bump up the tolerance for duct component loads and will add a note that we should see if we can improve it in the future.

Issue #1993, opened 2025-04-24
Suspicious HPWH vs ERWH tank losses


shorowit:
If we compare reasonably similar ERWH & HPWH products:
1. We see a ~1.7x difference in tank losses. Can we found ERWH measured data to support this difference?
2. We have a different assumption between EF and UEF models for ERWHs. Should probably make them consistent. Does the sensitivity to `f_low` make sense?

shorowit:
Waiting on Bethany experiment. Regardless of results, we should still consider `f_low` consistency between EF and UEF models.

jmaguire1:
I haven't tested it yet, but I think if we agree we want to use f_low with EF, it'd look like this:

what we're doing now to calculate UA with EF:
ua_old = q_load * (1.0 / water_heating_system.energy_factor - 1.0) / ((t - t_env) * 24.0) # Btu/hr-F

what we'd do using f_low when calculating UA with EF:
ua = q_load * (1.0 / water_heating_system.energy_factor - 1.0) / ((24.0 * (t - t_env)) * (f_high + f_low * ((t_in - t_env) / (t - t_env)))) # Btu/hr-F

@shorowit: Do we want to try this out, and see how it impacts UA calculations with EF? We'll get slightly different answers still for two systems with the same EF value as UEF because q_load is different, although you could do a test run where you change the UEF assumptions to match the EF test (or vice versa) to make sure you get an identical UA value if we want to. It's a simple enough change but want to make sure this is what we actually want to consider as the resolution here before I do testing.

PR #2042, opened 2025-07-15
Use `f_low` when calculating UA from EF for storage water heaters

jmaguire1:
## Pull Request Description

See #1993 for the full discussion. Use "f_low" to account for the fact that below the lower element isn't heated for electric resistance water heaters when calculating tank losses. This is already done if you use UEF as the efficiency metric, but not if you use EF. This uses f_low consistently regardless of whether EF or UEF is the efficiency metric.

## Checklist

Not all may apply:


jmaguire1:
@shorowit: I don't think we need any documentation change here? I updated the comment in the code about how we calculate UA, but we never had anything in our documentation about how we go from UEF and RE to UA and conversion efficiency since those are the direct inputs. We could add something if we think it's helpful though?

 LIkewise not sure about Schamtron validator or sample files/tests, we could add tanks with different EFs to test that the UA value calcs don't change if we wanted to. 

I might need a little help from you on what we want to do to finish this up, but I think the core change of the calculating UA from EF is done and behaving as expected.

shorowit (approved):
Looks great to me. Thanks, @jmaguire1!

Issue #2067, opened 2025-08-19
Relax storm window limits


shorowit:
Accommodate more double pane windows w/ storms.

PR #2082, opened 2025-09-16
Relax Window Ufactor Limits to Apply Storm Windows

yzhou601:
## Pull Request Description

Addresses #2067 

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/constructions.rb
+++ b/HPXMLtoOpenStudio/resources/constructions.rb
@@ -2758,16 +2758,17 @@ module Constructions
 
     # Ref: https://labhomes.pnnl.gov/documents/PNNL_24444_Thermal_and_Optical_Properties_Low-E_Storm_Windows_Panels.pdf
     # U-factor and SHGC adjustment based on the data obtained from the above reference
-    min_base_ufactor_for_storm = 0.45
+    # Fixme: Should we remove the limit? It doesn't seem to only applicable to single-pane windows
+    min_base_ufactor_for_storm = 0.35
     if base_ufactor < min_base_ufactor_for_storm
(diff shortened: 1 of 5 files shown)


HPXMLtoOpenStudio/resources/constructions.rb
```diff
@@ -2758,16 +2758,17 @@ def self.get_ufactor_shgc_adjusted_by_storms(storm_type, base_ufactor, base_shgc
 
     # Ref: https://labhomes.pnnl.gov/documents/PNNL_24444_Thermal_and_Optical_Properties_Low-E_Storm_Windows_Panels.pdf
     # U-factor and SHGC adjustment based on the data obtained from the above reference
-    min_base_ufactor_for_storm = 0.45
+    # Fixme: Should we remove the limit? It doesn't seem to only applicable to single-pane windows
+    min_base_ufactor_for_storm = 0.35
```
yzhou601, line 2762: We could potentially remove this, I added a floor of 0.0 (or we can throw errors) below so that it never goes negative if we want to remove this limit. 
Related PR and discussions: https://github.com/NREL/OpenStudio-HEScore/pull/319
We have this check for single-pane because it's an single-pane window upgrade in HEScore, and we used to be concerned if adjustment goes negative.

shorowit: I would suggest just using 0.3 as the cutoff, that's realistically the lower limit of how good of a double pane window you might add a storm to. You can also add some language to the documentation describing why the limit exists. And let's update the link in the documentation/code to https://www.pnnl.gov/main/publications/external/technical_reports/PNNL-24444.pdf

PR #2086, opened 2025-10-09
HPWH COP Adjustment in confined space

### LINKED ISSUE #2009: HPWH adjustment for confined space
See RESNET Addendum

yzhou601:
## Pull Request Description

Fixes #2009 

## Checklist

Not all may apply:

- [x] Schematron validator (`EPvalidator.sch`) has been updated

### DIFF
--- a/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.sch
+++ b/HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.sch
@@ -2571,7 +2571,8 @@
       <sch:assert role='ERROR' test='count(h:WaterHeaterInsulation/h:Jacket/h:JacketRValue) &lt;= 1'>Expected 0 or 1 element(s) for xpath: WaterHeaterInsulation/Jacket/JacketRValue</sch:assert>
       <sch:assert role='ERROR' test='count(h:HotWaterTemperature) &lt;= 1'>Expected 0 or 1 element(s) for xpath: HotWaterTemperature</sch:assert>
       <sch:assert role='ERROR' test='count(h:UsesDesuperheater) &lt;= 1'>Expected 0 or 1 element(s) for xpath: UsesDesuperheater</sch:assert> <!-- See [Desuperheater] -->
-      <sch:assert role='ERROR' test='count(h:extension/h:HPWHInConfinedSpaceWithoutMitigation) &lt;= 1'>Expected 0 or 1 element(s) for xpath: HPWHInConfinedSpaceWithoutMitigation</sch:assert> <!-- See [HPWHInConfinedSpaceWithoutMitigation] -->
+      <sch:assert role='ERROR' test='count(h:extension/h:HPWHInConfinedSpaceWithoutMitigation) &lt;= 1'>Expected 0 or 1 element(s) for xpath: extension/HPWHInConfinedSpaceWithoutMitigation</sch:assert> <!-- See [HPWHInConfinedSpaceWithoutMitigation] -->
+      <sch:assert role='ERROR' test='h:extension/h:HPWHInConfinedSpaceWithoutMitigation[text()="true" or text()="false"] or not(h:extension/h:HPWHInConfinedSpaceWithoutMitigation)'>Expected extension/HPWHInConfinedSpaceWithoutMitigation to be 'true' or 'false'</sch:assert>
       <!-- Moved/deprecated extension/OperatingMode input; see https://github.com/NREL/OpenStudio-HPXML/pull/1289 -->
       <sch:assert role='ERROR' test='count(h:extension/h:OperatingMode) = 0'>extension/OperatingMode has been replaced by HPWHOperatingMode</sch:assert>
       <!-- Warnings -->
--- a/HPXMLtoOpenStudio/measure.xml
+++ b/HPXMLtoOpenStudio/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.1</schema_version>
   <name>hpxm_lto_openstudio</name>
   <uid>b1543b30-9465-45ff-ba04-1d1f85e763bc</uid>
-  <version_id>443ede68-ce2f-4532-921c-ebf2521b68f8</version_id>
-  <version_modified>2025-10-31T18:42:49Z</version_modified>
+  <version_id>ebf347a8-a3b8-4096-a72c-078aa24e7d59</version_id>
+  <version_modified>2025-10-31T18:46:56Z</version_modified>
   <xml_checksum>D8922A73</xml_checksum>
   <class_name>HPXMLtoOpenStudio</class_name>
   <display_name>HPXML to OpenStudio Translator</display_name>
@@ -402,7 +402,7 @@
       <filename>hpxml_schematron/EPvalidator.sch</filename>
       <filetype>sch</filetype>
       <usage_type>resource</usage_type>
-      <checksum>5C6EBF7D</checksum>
+      <checksum>565E6550</checksum>
     </file>
     <file>
       <filename>hpxml_schematron/iso-schematron.xsd</filename>


docs/source/workflow_inputs.rst
```diff
@@ -4309,6 +4311,11 @@ Each heat pump water heater is entered as a ``/HPXML/Building/BuildingDetails/Sy
   .. [#] NumberofBedroomsServed only required if IsSharedSystem is true.
          Tank losses will be apportioned to the dwelling unit using its number of bedrooms divided by the total number of bedrooms served by the water heating system per `ANSI/RESNET/ICC 301-2022 <https://codes.iccsafe.org/content/RESNET3012022P1>`_.
          Each dwelling unit w/zero bedrooms should be counted as 1 bedroom -- e.g., a value of 3 should be entered for a shared system serving 3 studio (zero bedroom) apartments.
+  .. [#] Use HPWHInstalledProperly to specify whether to apply the HPWH COP adjustment for confined space, refer to `RESNET HERS Addendum 77 <https://www.resnet.us/about/standards/minhers/draft-pds-01-minhers-addendum-77-integrated-heat-pump-water-heaters/>`_ for conditions where it applies.
+  .. [#] HPWHContainmentVolume only required if HPWHInstalledProperly is false.
+  .. [#] HPWHContainmentVolume is used to adjust HPWH COP per `RESNET HERS Addendum 77 <https://www.resnet.us/about/standards/minhers/draft-pds-01-minhers-addendum-77-integrated-heat-pump-water-heaters/>`_.
```
yzhou601: It's now pointing to the draft addendum, will need to replace after it's officially released, better link to use? Or just explain the approach details here without reference?

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.sch
```diff
@@ -2571,13 +2571,21 @@
...
       <sch:assert role='ERROR' test='count(h:extension/h:OperatingMode) = 0'>extension/OperatingMode has been replaced by HPWHOperatingMode</sch:assert>
       <!-- Warnings -->
       <sch:report role='WARN' test='number(h:HotWaterTemperature) &lt; 110'>Hot water setpoint should typically be greater than or equal to 110 deg-F.</sch:report>
     </sch:rule>
   </sch:pattern>
 
+  <sch:pattern>
+    <sch:title>[HPWHInstalledProperly]</sch:title>
+    <sch:rule context='/h:HPXML/h:Building/h:BuildingDetails/h:Systems/h:WaterHeating/h:WaterHeatingSystem/h:extension[h:HPWHInstalledProperly="false"]'>
+      <sch:assert role='ERROR' test='count(h:HPWHContainmentVolume) = 1'>Expected 1 element(s) for xpath: HPWHContainmentVolume</sch:assert>
```
yzhou601: Expected behaviors:

Installed Properly | Containment Volume | Reaction
-- | -- | --
TRUE | Provided | Ignore with warning
FALSE | Provided | Apply adjustment
TRUE | Missing | Pass without adjustment
FALSE | Missing | Error

HPXMLtoOpenStudio/tests/test_validation.rb
```diff
@@ -2091,6 +2098,16 @@ def test_ruby_warning_messages
       when 'floor-or-ceiling2'
         hpxml, hpxml_bldg = _create_hpxml('base-foundation-unvented-crawlspace.xml')
         hpxml_bldg.floors[0].floor_or_ceiling = HPXML::FloorOrCeilingCeiling
+      when 'hpwh-installed-properly'
+        hpxml, hpxml_bldg = _create_hpxml('base-dhw-tank-heat-pump-containment-volume.xml')
+        hpxml_bldg.water_heating_systems[0].hpwh_installed_properly = true
+      when 'hpwh-small-containment-volume-without-backup-element'
+        hpxml, hpxml_bldg = _create_hpxml('base-dhw-tank-heat-pump-containment-volume.xml')
+        hpxml_bldg.water_heating_systems[0].backup_heating_capacity = 0.0
+        hpxml_bldg.water_heating_systems[0].hpwh_containment_volume = 250.0
```
yzhou601: Unit test to verify that we throw warning that the adjustment may not be accurate when there's no backup heating

HPXMLtoOpenStudio/tests/test_water_heater.rb
```diff
@@ -1048,6 +1050,40 @@ def test_tank_heat_pump_capacities
     assert_in_epsilon(cap, coil.ratedHeatingCapacity, 0.01)
   end
 
+  def test_tank_heat_pump_containment_volume_adjustment
+    # volumes and their corresponding expected cops are based on RESNET spreadsheet: ConstrainedHPWH_04xlsx.xlsx
+    containment_volumes = [3000, 1500, 960, 707, 453, 200, 83.5]
+    expected_cop = [3.720, 3.720, 3.647, 3.518, 3.191, 2.367, 1.642]
```
yzhou601: I'm using the RESNET spreadsheet to calculate the adjusted COPs while changed the original COP to be 3.73135, which is our test file configurations. All the other changes in this file are made so that we can configure test files on the fly and verify EnergyPlus inputs.

docs/source/workflow_inputs.rst
yzhou601: @jmaguire1 Please review the doc changes

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -976,6 +976,19 @@ def self.apply_hpwh_dxcoil(runner, model, water_heating_system, elevation, obj_n
 
     cop = water_heating_system.additional_properties.cop
 
+    # Adjust COP based on RESNET HERS Addendum 77
+    if not water_heating_system.hpwh_containment_volume.nil?
+      if water_heating_system.hpwh_installed_properly.nil? || (water_heating_system.hpwh_installed_properly == true)
+        runner.registerWarning("HPWHContainmentVolume provided in #{water_heating_system.id} is ignored because it is only used when HPWHInstalledProperly is 'false'.")
+      else
+        if water_heating_system.hpwh_containment_volume < 450.0 && (water_heating_system.backup_heating_capacity == 0.0)
```
jmaguire1: Can we put a "FUTURE: apply for 120V HPWH and other system types correction may not be accurate for"? Just to track this better outside of that PR.

jmaguire1:
@shorowit: Reviewing this reminded me of a long wish listed improvement for HPWH interaction factor, because if it's in a confined space and this penalty applies we know there's a lot of localized cooling and the interaction factor should be < 1. Last time I talked to Ben Larson about what HPWHSim does he thought 0.75 is a reasonable confined space guess, but HPWHSim still uses 1.0 everywhere.

Maybe we should just start an issue if this is too much for this PR or a follow up PR, but I double checked our implementation of it in the past and it's really simple: we just multiply these two lines in our HPWH internal gains EMS by the interaction factor: https://github.com/NREL/OpenStudio-HPXML/blob/5141bcd2f2a2c607b72dde834e2aaf79cc617d54/HPXMLtoOpenStudio/resources/waterheater.rb#L1290

HPXMLtoOpenStudio/resources/hpxml_schematron/EPvalidator.sch
yzhou601: We decided to document the limitation instead of throwing warnings in this case.

Changelog.md
```diff
@@ -10,6 +10,7 @@ __New Features__
 - Water heater improvements:
   - Improves electric water heater tank losses when using `EnergyFactor` as the metric; now consistent with how `UniformEnergyFactor` is handled.
   - Improves HPWH tank volume defaulting, particularly when `NumberofResidents` is provided.
+  - Account for HPWH COP adjustment for confined space per RESNET HERS Addendum 77. `extension/HPWHContainmentVolume` is used to calculate the adjustment when `extension/HPWHInstalledProperly` is set to "false".
```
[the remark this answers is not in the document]
jmaguire1: To document this we should probably also link to the report on this topic (I put a link in one of my comments on documentation already).

HPXMLtoOpenStudio/tests/test_water_heater.rb
shorowit: For the link to the standard itself, just link to the final draft available at time of implementation (e.g., [this](https://www.resnet.us/about/standards/hers/draft-pds-03-hers-addendum-77-integrated-heat-pump-water-heaters-ihpwh/)) and we can update it down the road when it's officially published.

yzhou601:
> @shorowit: Reviewing this reminded me of a long wish listed improvement for HPWH interaction factor, because if it's in a confined space and this penalty applies we know there's a lot of localized cooling and the interaction factor should be < 1. Last time I talked to Ben Larson about what HPWHSim does he thought is a reasonable confined guess if you have 0 information about the installation location, but HPWHSim still uses 1.0 everywhere.
> 
> Maybe we should just start an issue if this is too much for this PR or a follow up PR, but I double checked our implementation of it in the past and it's really simple: we just multiply these two lines in our HPWH internal gains EMS by the interaction factor:
> 
> https://github.com/NREL/OpenStudio-HPXML/blob/5141bcd2f2a2c607b72dde834e2aaf79cc617d54/HPXMLtoOpenStudio/resources/waterheater.rb#L1290
> 
> Here's what the PNNL lab home testing had:
> 
> image
> If I were to propose something, we could default to 0.75 (or 1, to preserve our current behavior) and use 0.5 if you're in a confined space based on what PNNL saw in their lab home. It depends on a lot of things we still don't have information on, but we do at least expect if you're in a confined space and this correction applies it's not going to be 1.0 (or the cooling wouldn't be localized enough that we need this adjustment).

I've addressed other comments but this, I felt that we probably need more discussions around this (e.g. what volume to consider what factor to use, definition of confined space?). The equation to calculate relative volume uses 1500cbft as base volume but the standard specifies 1000cbft enclosed space as one of the mitigation, which volume is considered as the definition of confined space? Should we consider implementing it in this PR or with a follow-up issue?

docs/source/workflow_inputs.rst
```diff
@@ -4259,26 +4259,28 @@ Heat Pump
...
+  ``UsageBin`` or ``FirstHourRating``                  string or double  str or gal/hr  See [#]_ or > 0         No        See [#]_        EnergyGuide label usage bin/first hour rating
+  ``WaterHeaterInsulation/Jacket/JacketRValue``        double            F-ft2-hr/Btu   >= 0                    No        0               R-value of additional tank insulation wrap
+  ``HotWaterTemperature``                              double            F              > 0                     No        125             Water heater setpoint [#]_
+  ``UsesDesuperheater``                                boolean                                                  No        false           Presence of desuperheater? [#]_
+  ``extension/NumberofBedroomsServed``                 integer                          > NumberofBedrooms      See [#]_                  Number of bedrooms served directly or indirectly
+  ``extension/HPWHInConfinedSpaceWithoutMitigation``   boolean                                                  No                        Whether hpwh is installed in confined space without mitigation [#]_
```
shorowit: Whenever Required=no, the default value should be listed.

tasks.rb
```diff
@@ -2092,6 +2092,9 @@ def apply_hpxml_modification_sample_files(hpxml_path, hpxml)
     elsif ['base-dhw-tank-gas-fhr.xml'].include? hpxml_file
       hpxml_bldg.water_heating_systems[0].first_hour_rating = 56.0
       hpxml_bldg.water_heating_systems[0].usage_bin = nil
+    elsif ['base-dhw-tank-heat-pump-containment-volume.xml'].include? hpxml_file
```
shorowit: Can we rename the sample file to `base-dhw-tank-heat-pump-confined-space.xml`? Every water heater has a surrounding volume, but the point here is that the surrounding volume in this test file is small (i.e., it's a confined space).

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -976,6 +976,20 @@ def self.apply_hpwh_dxcoil(runner, model, water_heating_system, elevation, obj_n
 
     cop = water_heating_system.additional_properties.cop
 
+    # Adjust COP based on RESNET HERS Addendum 77
+    if not water_heating_system.hpwh_containment_volume.nil?
+      if water_heating_system.hpwh_confined_space_without_mitigation.nil? || (water_heating_system.hpwh_confined_space_without_mitigation == false)
```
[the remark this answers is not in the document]
yzhou601: I thought about this before, I felt the input name implies that this is something only applicable to HPWH and especially when it's in confined space, so I was trying to allow no input (and no default inputs being added) when it doesn't make sense to apply. But I think it's easy to sort out the cases, so that it is only defaulted to hpwh system type, and with better naming now it may not be that confusing, I'm fine with doing that.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -976,6 +976,20 @@ def self.apply_hpwh_dxcoil(runner, model, water_heating_system, elevation, obj_n
...
+        runner.registerWarning("HPWHContainmentVolume provided in #{water_heating_system.id} is ignored because it is only used when HPWHInConfinedSpaceWithoutMitigation is 'true'.")
```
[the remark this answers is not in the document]
yzhou601: Actually, this is a tricky question because based on the equation, even for volume greater than 1500cbft (the relative volume is calculated as a ratio to 1500), the COP will still be slightly adjusted. However, if the volume is greater than 1000cbft, the standard says it satisfies the requirement of a mitigation:
image
It means if the volume is for example 1200cbft, and user provides this input, then there'll be a small penalty on COP, while if the user omits this input, it's legit and there's no penalty. This is quite confusing.
And otherwise it sounds good to change the warning message.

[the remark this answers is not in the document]
shorowit: The way I think about it is that there's no need to give a warning if the adjustment is _very small_. And anything above 1000 cuft seems like a very small adjustment to me.

PR #2105, opened 2025-12-01
Whole building MF inter-unit heat transfer -- simplification

shorowit:
## Pull Request Description

Branches off of #1850

## Checklist

Not all may apply:

### DIFF
--- a/HPXMLtoOpenStudio/resources/hpxml.rb
+++ b/HPXMLtoOpenStudio/resources/hpxml.rb
@@ -12516,6 +12516,9 @@ class HPXML < Object
             if not adjacent_ap.respond_to? :adjacent_hpxml_id
               adjacent_ap.adjacent_hpxml_id = sameas_object.id
               adjacent_ap.adjacent_unit_number = sameas_object.parent_object.parent_object.buildings.index(sameas_object.parent_object)
+              # Note: sameas surface is assumed to have the same interior_adjacent_to as the adjacent surface.
+              # If that's not the case, we would have to allow InteriorAdjacentTo to be provided for the sameas
+              # surface. See https://github.com/NREL/OpenStudio-HPXML/pull/2105#discussion_r2583146171.
               adjacent_ap.adjacent_space_type = adjacent_obj.interior_adjacent_to
             elsif adjacent_ap.adjacent_hpxml_id != sameas_object.id
               fail "'#{adjacent_obj.id}' is referenced by multiple objects."
--- a/HPXMLtoOpenStudio/resources/model.rb
+++ b/HPXMLtoOpenStudio/resources/model.rb
@@ -1173,7 +1173,8 @@ module Model
         fail "Could not find adjacent space type ('#{adjacent_space_type}') in unit #{adjacent_unit_number}."
       end
 
-      adjacent_space = merged_model_objects.find { |s| s.to_Space.is_initialized && s.name.to_s == unit_adjacent_zone.spaces[0].name.to_s }.to_Space.get
+      unit_adjacent_space = unit_adjacent_zone.spaces[0]
+      adjacent_space = merged_model_objects.find { |s| s.to_Space.is_initialized && s.name.to_s == unit_adjacent_space.name.to_s }.to_Space.get
 
       # Create/assign adjacent surface
       adjacent_surface = surface.createAdjacentSurface(adjacent_space)
--- a/HPXMLtoOpenStudio/measure.xml
+++ b/HPXMLtoOpenStudio/measure.xml
@@ -3,8 +3,8 @@
   <schema_version>3.1</schema_version>
   <name>hpxm_lto_openstudio</name>
   <uid>b1543b30-9465-45ff-ba04-1d1f85e763bc</uid>
-  <version_id>d942f727-39ff-4006-bfdb-d8d8d1e85ed4</version_id>
-  <version_modified>2025-12-03T02:02:44Z</version_modified>
+  <version_id>095f8403-82c3-4bee-a208-7961172a65c7</version_id>
+  <version_modified>2025-12-03T02:16:57Z</version_modified>
   <xml_checksum>D8922A73</xml_checksum>
   <class_name>HPXMLtoOpenStudio</class_name>
   <display_name>HPXML to OpenStudio Translator</display_name>
@@ -384,7 +384,7 @@
       <filename>hpxml.rb</filename>
       <filetype>rb</filetype>
       <usage_type>resource</usage_type>
-      <checksum>9BD32C5F</checksum>
+      <checksum>615D33CF</checksum>
     </file>
     <file>
       <filename>hpxml_schema/HPXML.xsd</filename>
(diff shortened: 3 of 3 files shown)


HPXMLtoOpenStudio/resources/model.rb
```diff
@@ -1155,19 +1155,51 @@ def self.merge_unit_models(model, hpxml_osm_map)
...
+      unit_surface = OpenStudio::Model::Surface.new(adjacent_surface.vertices, unit_model)
+      surface_type = surface.additionalProperties.getFeatureAsString('SurfaceType').get
+      if surface_type == 'Floor'
+        surface_type = 'Ceiling'
+      elsif surface_type == 'Ceiling'
+        surface_type = 'Floor'
+      end
+      unit_surface.additionalProperties.setFeature('SurfaceType', surface_type)
+      unit_surface.setSpace(unit_adjacent_space)
+      unit_surface.setName(adjacent_surface.name.to_s)
```
shorowit: This was the final change I had to make to get the CI passing. Because we no longer create the adjacent surface during the unit model generation, it was being excluded from the component loads program (and so the component loads were not correctly summing to the total loads). I guess that is a disadvantage to this approach. But I still think this approach is a big improvement overall.

yzhou601: I see, so this code is to add a corresponding surface to unit model with the same name, so that the sensor could be correctly added to the combined model. The component load program is fixed and CI is passing. It doesn't sound like there's impact on the component load program accuracy with the fix. 

It's a bit tricky that the component load program is still reading the unit model objects after model combination, I guess the reason is that we're storing a lot of information in the `additionalProperties` and it's only available at unit models? I think this is probably the best we can do and I'm fine with this workaround.

workflow/tests/base_results/results_simulations_misc.csv
```diff
@@ -65,9 +65,9 @@ base-bldgtype-mf-unit-shared-water-heater-recirc-scheduled.xml,0.0,0.0,0.0,1323.
 base-bldgtype-mf-unit-shared-water-heater-recirc.xml,0.0,0.0,0.0,1323.0,910.6,12046.2,4107.6,873.2,1706.3,1706.3,873.2,1706.3,1706.3,4.161,8.333,0.0
 base-bldgtype-mf-unit-shared-water-heater.xml,0.0,0.0,0.0,1323.0,910.6,12046.2,4107.6,836.5,1669.6,1669.6,836.5,1669.6,1669.6,4.161,8.333,0.0
 base-bldgtype-mf-unit.xml,0.0,0.0,0.0,1323.0,910.7,12046.2,4107.6,1704.3,2124.6,2124.6,1704.3,2124.6,2124.6,3.934,8.329,0.0
-base-bldgtype-mf-whole-building-common-spaces.xml,0.0,1.0,0.0,7938.3,5463.9,72300.7,24653.5,18291.0,14055.6,18291.0,18291.0,14055.6,18291.0,34.281,23.945,0.0
+base-bldgtype-mf-whole-building-common-spaces.xml,0.0,1.0,0.0,7938.3,5463.9,72300.7,24653.5,18292.3,14055.5,18292.3,18292.3,14055.5,18292.3,34.285,23.944,0.0
```
yzhou601: small diffs, great job! 👍

HPXMLtoOpenStudio/resources/geometry.rb
```diff
@@ -134,8 +134,7 @@ def self.apply_walls(runner, model, spaces, hpxml_bldg, hpxml_header)
     foundation_top = hpxml_bldg.building_construction.additional_properties.foundation_height_above_grade
 
     hpxml_bldg.walls.each do |wall|
-      orig_wall = wall # Only sameas walls have different "wall" and "orig_wall"
-      wall = wall.sameas if orig_wall.sameas_id
+      next if wall.sameas_id
```
yzhou601: Nice cleanups!

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -4889,11 +4881,7 @@ def exposed_fraction
     #
     # @return [Boolean] True if an exterior surface
     def is_exterior
-      if @sameas_id || (@additional_properties.respond_to? :referenced_by_sameas)
-        return false
-      else
-        return @exterior_adjacent_to == LocationGround
-      end
+      return @exterior_adjacent_to == LocationGround
```
yzhou601: Glad to see the complication here is gone!

yzhou601 (approved):
It's a very great improvement to simplify codes! The `adjacentSpaceType` and `adjacentUnitNumber` additional properties are serving as a space ID reference that we can find the correct space to assign to the surface at combining. The `adjacentHpxmlID` is stored to create a more explicit adjacent surface name, I don't know if it's worth to store but I'm fine with keeping it. 
Overall I think the approach is robust enough. Only one small question (should be a simple fix) and all others look good to me.

HPXMLtoOpenStudio/resources/hpxml.rb
```diff
@@ -12530,11 +12511,13 @@ def self.get_sameas_obj(parent_building, object)
...
-            # Assign referenced_by_sameas
+            # Assign adjacent_hpxml_id
             ap = obj.additional_properties
-            if not ap.respond_to? :referenced_by_sameas
-              ap.referenced_by_sameas = object.id
-            elsif ap.referenced_by_sameas != object.id
+            if not ap.respond_to? :adjacent_hpxml_id
+              ap.adjacent_hpxml_id = object.id
+              ap.adjacent_unit_number = object.parent_object.parent_object.buildings.index(object.parent_object)
+              ap.adjacent_space_type = obj.interior_adjacent_to
```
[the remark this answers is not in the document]
yzhou601: Oh I think it comes to the same situation as the previous approach, the sameas object doesn't contain interior adjacent to, so we need to borrow it from the fully-described object, and for most cases the interior adjacent to is supposed to be the same - "conditioned space". I think that's probably why you use the `obj` here.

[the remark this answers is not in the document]
yzhou601: So there's actually some limitations for use cases like base-bldgtype-mf-whole-building-inter-unit-heat-transfer.xml, e.g. the surfaces at attic can only refer to another surface at attic, the same applies to unconditioned basement.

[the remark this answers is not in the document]
shorowit: Weird, somehow your additional comments didn't appear when I made my comment above. Sorry. In any case, it seems that we came to the same conclusion. The only way I can see to "fix" this would be to allow `InteriorAdjacentTo` for the sameas surface, but I don't think it's worth it right now.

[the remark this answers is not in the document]
yzhou601: Yes the only thought related to this approach is that if there's only very small amount of elements in e.g. the attic, and the space is not created after skipping all the sameas surfaces(is it possible?), then the space approach will likely to have some issues.

PR #2137, opened 2026-01-06
Fix for batteries and vehicles when using unit multipliers

joseph-robertson:
## Pull Request Description

This PR fixes several issues:
* We previously weren't checking unit multipliers for batteries and vehicles
* We were assuming the Resilience output type should change based on unit multiplier
* Some kind of floating point issue was causing weird resilience hour values for batteries
* We were adding extra battery losses (and without incorporating unit multipliers) to critical load
* Vehicles were missing multiplying `eff_discharge_power` by `unit_multiplier`

## Checklist

Not all may apply:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,8 @@ __Bugfixes__
 - Fixes possible incorrect unmet hours outputs for unavailable periods with no space heating only (or no space cooling only).
 - Fixes PanHeaterControlType="heat pump mode" incorrectly disallowed for mini-split heat pumps.
 - Fixes the design cooling temperature calculations for some vented attic roof types (shingles, plastic/rubber/synthetic sheeting, concrete, cool roof, expanded polystyrene sheathing)
+- Fixes unit multiplier ignored for EV charging using `Vehicles`.
+- Fixes handling of battery losses in the battery resilience output.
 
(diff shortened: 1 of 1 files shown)


ReportSimulationOutput/measure.rb
```diff
@@ -2036,7 +2039,7 @@ def get_resilience_timestep_value(init_time_step, batt_kwh, batt_kw, batt_soc_kw
       load_kw = crit_load[t]
 
       # even if load_kw is negative, we return if batt_soc_kwh isn't charged at all
-      return i / Float(ts_per_hr) if batt_soc_kwh <= 0
+      return i / Float(ts_per_hr) if batt_soc_kwh.abs <= Constants::Small
```
shorowit: I don't understand this change. If it's a large negative number, it no longer returns? Why is it not `batt_soc_kwh <= Constants::Small` if it's just a floating point issue?

shorowit: I also don't see any diffs for CI results, why is that?

shorowit (commented):
This looks good. Ran several simulations and verified the results look better now. One minor suggestion below.

shorowit (approved):

PR #2189, opened 2026-04-07
Refactor sensors

shorowit:
## Pull Request Description

While reviewing @joseph-robertson's #2151, I was reminded that we have many EMS programs that request identical sensors, which is duplicative and ineffecient. Now I create a number of global sensors (for site/outdoor conditions as well as the conditioned zone) up front _once_ and then reuse them as needed.

## Checklist

Not all may apply:

- [ ] Schematron validator (`EPvalidator.sch`) has been updated
- [ ] Sample files have been added/updated (`openstudio tasks.rb update_hpxmls`)
- [ ] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests/test*.rb` and/or `workflow/tests/test*.rb`)
- [ ] Documentation has been updated
- [ ] Changelog has been updated
- [x] `openstudio tasks.rb update_measures` has been run
- [x] No unexpected changes to simulation results of sample files

PR #2191, opened 2026-04-09
Fix unmet hours calculation when using on/off thermostat deadband model

shorowit:
## Pull Request Description

Noticed while working on #2189. Also adds the reporting tolerance for consistency with non-deadband simulations.

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/output.rb
```diff
@@ -151,7 +154,7 @@ def self.apply_unmet_hours_ems_program(model, hpxml_osm_map, hpxml_header)
         line += " && (#{htg_avail_sensor.name} == 1)" if not htg_avail_sensor.nil?
         program.addLine(line)
         if zone_air_temp_sensors.keys.include? unit # on off deadband
-          program.addLine("  If #{zone_air_temp_sensors[unit].name} < (#{htg_spt_sensors[unit].name} - #{UnitConversions.convert(onoff_deadbands, 'deltaF', 'deltaC')})")
+          program.addLine("  If #{zone_air_temp_sensors[unit].name} < (#{htg_spt_sensors[unit].name} - #{htg_tol})")
```
shorowit, line 157: While inspecting the results for `base-hvac-air-to-air-heat-pump-1-speed-research-features.xml`, I found that the indoor temperature was being compared to 61F. The heating setpoint is 64F and the deadband is 2F, so it should be compared to 63F. (Or, really, 63F minus the reporting tolerance of 0.2 deg-C.)

shorowit: Here's an example from another EMS program that shows  the lower end of the heating setpoint is the schedule value:
https://github.com/NatLabRockies/OpenStudio-HPXML/blob/209c72c378bf556232aec09aaf84d72d9a751727/HPXMLtoOpenStudio/resources/hvac.rb#L3450-L3451

workflow/tests/base_results/results_simulations_misc.csv
```diff
@@ -215,11 +215,11 @@ base-hvac-air-to-air-heat-pump-1-speed-detailed-performance.xml,0.0,0.0,0.0,1286
 base-hvac-air-to-air-heat-pump-1-speed-heating-capacity-17f.xml,0.0,0.0,0.0,1286.4,890.5,11468.5,3942.3,9035.1,4233.1,9035.1,9035.1,4233.1,9035.1,31072.0,24785.0,0.0
 base-hvac-air-to-air-heat-pump-1-speed-heating-only.xml,0.0,0.0,0.0,1286.4,890.5,11468.6,3942.3,9046.3,1887.6,9046.3,9046.3,1887.6,9046.3,31055.0,0.0,0.0
 base-hvac-air-to-air-heat-pump-1-speed-lockout-temperatures.xml,148.0,0.0,0.0,1286.4,890.5,11468.6,3942.3,15409.2,4217.6,15409.2,15409.2,4217.6,15409.2,51750.0,24621.0,0.0
-base-hvac-air-to-air-heat-pump-1-speed-research-features.xml,250.567,0.0,0.0,1286.4,890.5,12241.8,4208.1,20808.8,9769.5,20808.8,20808.8,9769.5,20808.8,63178.0,33162.0,0.0
+base-hvac-air-to-air-heat-pump-1-speed-research-features.xml,412.433,0.0,0.0,1286.4,890.5,12241.8,4208.1,20808.8,9769.5,20808.8,20808.8,9769.5,20808.8,63178.0,33162.0,0.0
```
shorowit, line 218: Regardless of the change, I wonder why there are so many unmet hours in this file? And the 2-speed one below?

jmaguire1: For two speed equipment with a deadband thermostat, we have a logic that it starts at speed 1 for 5 minutes (assuming 1 minute timestep), then determines if it has to go into speed 2 or not. That's a real logic of some units, but I'd expect it to lead to some unmet load since you can't jump ahead to speed 2 until 5 minutes have passed. More advanced control logics can do something else but it's very hard to actually pick a control logic out of field data.

shorowit: There are also significant unmet hours for the 1 speed case.

But looking closer at the test files, it appears to be caused by having a large heating setback (6F), so we get unmet hours during setback recovery. Okay, I'm not concerned then.

HPXMLtoOpenStudio/resources/output.rb
yzhou601: Interesting, good catch, the previous approach is wrong because we already shifted the schedule by 1/2 deadband so that the `htg_spt_sensors` will give 63F (cut-in) instead of cut-out, so applying full deadband here is causing issues.

yzhou601 (approved):
Great bugfix, thanks @shorowit

PR #2192, opened 2026-04-13
Fix timeseries heating/cooling setpoints when using on/off thermostat deadband model

shorowit:
## Pull Request Description

We now use the heating/cooling setpoint schedules ("Schedule Value") rather than the zone output variables (e.g., "Zone Thermostat Heating Setpoint Temperature"), since E+ will automatically adjust the latter as part of the model.

Results for `base-hvac-air-to-air-heat-pump-1-speed-research-features.xml`:

image

The blue/orange data is before this change and the red/black is after. The latter no longer shows the E+ adjustments that are being made. However, they may still be confusing to a user because the HPXML inputs are 78F for cooling and 64-70F for heating, with a 2F deadband, while the timeseries results show the _edges_ of the deadband. Maybe we just need to document this?

## Checklist

Not all may apply:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -26,6 +26,9 @@ __Bugfixes__
   - Fixes lighting and plug/fuel load energy use to not be zeroed out when a kWh/year or therm/year value is provided.
   - Fixes pool/spa energy use to be zeroed out when a kWh/year or therm/year value is not provided, or when there is a "Vacancy" unavailable period.
 - Fixes incorrect hot water (gallons) output for solar thermal systems using `SolarFraction`.
+- Small bugfixes when using the on/off thermostat deadband advanced research feature:
+  - Fixes unmet hours outputs, which could be missing some periods of unmet hours.
+  - Fixes timeseries outputs for heating/cooling setpoints.
 
 ## OpenStudio-HPXML v1.11.1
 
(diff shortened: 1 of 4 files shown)


shorowit:
> However, they may still be confusing to a user because the HPXML inputs are 78F for cooling and 64-70F for heating, with a 2F deadband, while the timeseries results show the _edges_ of the deadband. Maybe we just need to document this?

@yzhou601 @jmaguire1 Let me know if you have an opinion on this.

jmaguire1:
I think documentation would be fine here unless we start getting complaints from users. Thanks for highlighting this one and fixing the bug!

yzhou601:
> > However, they may still be confusing to a user because the HPXML inputs are 78F for cooling and 64-70F for heating, with a 2F deadband, while the timeseries results show the _edges_ of the deadband. Maybe we just need to document this?
> 
> @yzhou601 @jmaguire1 Let me know if you have an opinion on this.

Yep we need to document it or we can post processing if it's just for reporting (it would be a larger lift that I'm not sure if it's worthy). The ideal is improving E+ to adjust the way they model the on-off deadband thermostat, currently it takes setpoint schedule as cut-in temperature and applies the temperature difference above/below setpoint, it would be nice if they can take setpoint schedule as the middle point and apply deadband around it. That will be a smaller lift on E+ side but can save so much processing on our end!

PR #1534, opened 2023-10-31
120 v hpwh

jmaguire1:
## Pull Request Description

Add 120V HPWH parameters into the model. We'll eventually want a "shared" and "dedicated" circuit HPWH model, so two different sets of parameters when it's a 120 V HPWH.

Based on [unmet_wh_loads PR](https://github.com/NREL/OpenStudio-HPXML/pull/1502)

## Checklist

PR Author: Check these when they're done. Not all may apply. ~~strikethrough~~ and check any that do not apply. 

PR Reviewer: Verify each has been completed.

- [ ] Schematron validator (`EPvalidator.xml`) has been updated
- [ ] Sample files have been added/updated (via `tasks.rb`)
- [ ] Tests have been added/updated (e.g., `HPXMLtoOpenStudio/tests` and/or `workflow/tests/hpxml_translator_test.rb`)
- [ ] Documentation has been updated
- [ ] Changelog has been updated
- [ ] `openstudio tasks.rb update_measures` has been run
- [ ] No unexpected changes to simulation results of sample files

PR #2195, opened 2026-04-20
120V HPWHs

### LINKED ISSUE #1427: 120 V HPWH
A piece of #1170. We'll need this capability for a project led by @lixiliu.

- [x] Generate performance map for 120 V products
- [x] Validate control logic used
- [ ] Add unmet fixture energy EMS program (use aggregate of SSB)
- [x] Allow optional `BackupHeatingCapacity` element https://github.com/NREL/OpenStudio-HPXML/pull/1857

If I have time I'll try to address the other pieces of #1170.

shorowit:
## Pull Request Description

Closes #1427.

Re-implements #1534 off the master branch. Uses the HPXML `HPWHVoltage`/`HasMixingValue`/`MixingValveSetpoint` elements (added [here](https://github.com/hpxmlwg/hpxml/pull/431)).

## Checklist

Not all may apply:

- [x] Schematron validator (`EPvalidator.sch`) has been updated


HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -6268,22 +6285,28 @@ def self.get_water_heater_tank_volume(fuel, is_hpwh, nbeds, nbaths, n_occ)
...
+        when HPXML::WaterHeaterUsageBinMedium
+          cop = 1.0909 * uef - 0.0868
+        when HPXML::WaterHeaterUsageBinHigh
+          cop = 1.1022 * uef - 0.0877
+        end
+      end
+    elsif water_heating_system.hpwh_voltage == HPXML::HPWHVoltage120Dedicated
+      cop = 3.6
+    else # 120V shared
+      cop = 4.2
```
shorowit: Need to replace with equations based on UEF. @jmaguire1

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -1063,6 +1087,7 @@ def self.apply_hpwh_stratified_tank(model, water_heating_system, obj_name, solar
     # https://neea.org/img/documents/hpwh-lab-report_ao-smith_hptu_12-09-2015.pdf.
     # More recent products do not show much change to UA values, see 2021 report:
     # https://neea.org/img/documents/Laboratory-Assessment-of-Rheem-Generation-5-Series-HPWH.pdf.
+    # FIXME: Is this close enough to the right value for 120V, or do we need new UA values?
```
shorowit: @jmaguire1 will check the data.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -963,17 +977,27 @@ def self.apply_hpwh_wrapped_condenser(model, obj_name, coil, tank, fan, airflow_
...
     # Curves
+    if water_heating_system.hpwh_voltage == HPXML::HPWHVoltage240
+      cap_coeff = [0.563, 0.0437, 0.000039, 0.0055, -0.000148, -0.000145]
+      cop_coeff = [1.1332, 0.063, -0.0000979, -0.00972, -0.0000214, -0.000686]
+    elsif water_heating_system.hpwh_voltage == HPXML::HPWHVoltage120Dedicated
+      cap_coeff = [0.636, 0.0227, 0.000406, -0.000437, 0.0, 0.0]
+      cop_coeff = [1.1798, 0.030, 0.0002063, -0.01935, 0.0001341, -0.000303]
+    else # 120V shared
+      cap_coeff = [0.813, 0.0160, 0.000537, 0.002032, -0.000086, -0.000069]
+      cop_coeff = [1.0132, 0.044, 0.0000117, -0.01113, 0.0000369, -0.000498]
```
shorowit: Need to double-check that I implemented this correctly. Was seeing an E+ warning about COP curve not 1 at nominal conditions.

docs/source/workflow_inputs.rst
```diff
@@ -4247,10 +4247,11 @@ Each heat pump water heater is entered as a ``/HPXML/Building/BuildingDetails/Sy
...
+  ``HPWHOperatingMode``                                string                           See [#]_                No        hybrid/auto     HPWH operating mode [#]_
+  ``HPWHVoltage``                                      string                           See [#]_                No        240V            HPWH voltage
   ``HPWHDucting/ExhaustAirTermination``                string                           See [#]_                No        <none>          The location where HPWH exhaust air is ducted to
-  ``HPWHOperatingMode``                                string                           See [#]_                No        hybrid/auto     Operating mode [#]_
   ``UsageBin`` or ``FirstHourRating``                  string or double  str or gal/hr  See [#]_ or > 0         No        See [#]_        EnergyGuide label usage bin/first hour rating
   ``WaterHeaterInsulation/Jacket/JacketRValue``        double            F-ft2-hr/Btu   >= 0                    No        0               R-value of additional tank insulation wrap
   ``HotWaterTemperature``                              double            F              > 0                     No        125             Water heater setpoint [#]_
```
shorowit: Require HotWaterTemperature >= 105F.

HPXMLtoOpenStudio/resources/hotwater_appliances.rb
```diff
@@ -352,6 +352,15 @@ def self.apply(runner, model, weather, spaces, hpxml_bldg, hpxml_header, schedul
         limits: EPlus::ScheduleTypeLimitsTemperature
       )
 
+      # FIXME: Use HPXML HasMixingValve and MixingValueSetpoint elements?
+      # FIXME: Should this be min(125.0, t_set)?
```
shorowit: - [x] Assuming mixing valve if 120V HPWH (@jmaguire1 will double check).
- [x] Default mixing valve setpoint at min(125.0, t_set).
- [x] MixingValueSetpoint must be < WHsetpoint and >= 105F
- [x] Warning that target temp > hot water temp is okay to ignore.

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: @jmaguire1 I'm pretty sure I implemented the curves correctly. See [120V dedicated here](https://github.com/NatLabRockies/OpenStudio-HPXML/blob/c50bbc5c2a95a8226186354af28bf398fc89491f/HPXMLtoOpenStudio/resources/waterheater.rb#L717-L722) and [120V shared here](https://github.com/NatLabRockies/OpenStudio-HPXML/blob/c50bbc5c2a95a8226186354af28bf398fc89491f/HPXMLtoOpenStudio/resources/waterheater.rb#L744-L749). Both give E+ warnings that the curves evaluate to ~0.8 at rated conditions.

HPXMLtoOpenStudio/resources/waterheater.rb
jmaguire1: Interestingly, it's not that great, so we might want new UA values:

<head>

<link rel=File-List
<style>

</style>
</head>

Model | Tank Volume | HPWHSim UA   value (kJ/hr-C) | HPWHSim UA   Value (Btu/hr-F) | OS-HPXML UA   (Btu/hr-F) | Hand   Calculation (Btu/hr-F)
-- | -- | -- | -- | -- | --
Rheem, Shared | 40 | 2.64 | 1.39 | 3.6 | -
Rheem, Shared | 50 | 2.38 | 1.25 | 3.6 | 1.95
Rheem, Shared | 65 | 2.96 | 1.56 | 4 | -
Rheem, Shared | 80 | 3.01 | 1.58 | 4.7 | -
Rheem,   Dedicated | 40 | 1.53 | 0.80 | 3.6 | -
Rheem,   Dedicated | 50 | 1.76 | 0.93 | 3.6 | 1.95



</body>

</html>


I was able to [pull UA from HPWHSim](https://github.com/bigladder/HPWHsim/tree/60ef3256630273318590b23069d439b3d23ac07d/test/models_json), and the [default units are kJ/hr-C](https://github.com/bigladder/HPWHsim/blob/60ef3256630273318590b23069d439b3d23ac07d/src/HPWH.hh#L787). So it appears that UA is a bit less for 120V units? I'm a little surprised how much less, but the UA hand calculation I did confirms that HPWHSim is at least in the ballpark of what I calculated for the units I had raw data for that I used to create and validate this model. I'd suggest we use HPWHSim's values, since it's citable, for 120V units.

HPXMLtoOpenStudio/resources/waterheater.rb
jmaguire1: This is in fact implemented correctly, the way I calculated these curves I didn't explicitly enforce 1.0 at nominal conditions. I'll adjust nominal values and recalculate curves to adjust (I think this is worth the effort, so our nominal COP and capacity have the expected meaning), I think the same biquadric tool for HVAC can be used here with minimal reconfiguring.

Note we had to [altitude adjust the rated wet bulb](https://github.com/NatLabRockies/OpenStudio-HPXML/blob/0932903b9c8ca9fd22c37b1b182295ae7dfb8236/HPXMLtoOpenStudio/resources/waterheater.rb#L1003) at one point, I'm fairly sure because we hit psychrometric issues if we didn't much like what you saw [here](https://github.com/NatLabRockies/OpenStudio-HPXML/issues/2227). I'll make sure we get 1.0 at sea level conditions, but you could still hit warnings at elevation because of this adjustment. We did this a long time ago, but I think it's still necessary.

jmaguire1: Shared Circuit, difference between previous COP and new curve centered at 1.0:

image

Shared Circuit, difference between previous capacity and new curve centered at 1.0:
image

Dedicated Circuit, difference between previous COP and new curve centered at 1.0:
image

For the dedicated circuit unit, the capacity at nominal conditions is already 0.997, so I'm going to leave that one as is.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -3372,7 +3372,7 @@ def self.apply_water_heaters(hpxml_bldg, eri_version, schedules_file)
           elsif water_heating_system.hpwh_voltage == HPXML::HPWHVoltage120Dedicated
             water_heating_system.heating_capacity = (UnitConversions.convert(0.422, 'kW', 'Btu/hr') * water_heating_system.additional_properties.cop).round
           else # 120V shared
-            water_heating_system.heating_capacity = (UnitConversions.convert(0.357, 'kW', 'Btu/hr') * water_heating_system.additional_properties.cop).round
+            water_heating_system.heating_capacity = (UnitConversions.convert(0.424, 'kW', 'Btu/hr') * water_heating_system.additional_properties.cop).round
```
shorowit: Thanks @jmaguire1. The two values (0.422 and 0.424) are so similar that I'm just going to use 0.423 for all 120V HPWHs.

HPXMLtoOpenStudio/resources/hotwater_appliances.rb
jmaguire1: So Rheem offers it both ways, [with and without a mixing valve](https://www.rheem.com/heatpumpwaterheaters/#) (they call it "hydroboost)

AO Smith [does not](https://assets.hotwater.com/damroot/Original/10017/100379653.pdf) include one (if you go through the manual, they recommend installing mixing valves at each point of use, rather than at the water heater outlet, and for the controls there's no way to set a separate mixing valve setpoint.

GE does include one, but their product is quite different and this performance map wouldn't hold (because it's a variable speed compressor along with other design decisions).

So bottom line, we should not assume a mixing valve is included, although certainly you can either buy a model with one or install your own after installation.

shorowit: Rather than assume a mixing valve by WH type, we will assume a mixing valve based on WH setpoint temperature. @jmaguire1 suggests using 140F as a threshold. (Jeff, feel free to add any additional information/thoughts if you want.)

jmaguire1: We're going to assume 140 F because most water heaters have that as the maximum setpoint, so anything above that (which at least 120V product allows) would be a special case where the scalding risk goes up dramatically. For posterity:

image

shorowit: Implemented, marking as resolved.

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: The updated curves no longer give a warning, marking this as resolved.

shorowit:
- [x] Add a warning if a 120V HPWH has backup elements with too high power (900W for shared, 1800W for dedicated).
- [x] UEF->COP correlation
- [x] Jeff will created README and upload zip file with UEF correlation input files
- [x] Jeff will double-check control algorithm for 120V systems w/ backup

HPXMLtoOpenStudio/resources/waterheater.rb
jmaguire1: OK, so I figured out the biggest issue here: the units in HPWHSim are W/C, not kJ/hr-C (the schema is in a separate repo, that's how I got a little turned around). Here's what the UA value comparison actually looks like (I pulled what HPWHSim says for Rheem 240V HPWHs for reference too):

<head>

<link rel=File-List
<style>

</style>
</head>

Model | Nominal Tank   Volume (gal) | HPWHSim UA   Value (Btu/hr-F) | OS-HPXML UA   (Btu/hr-F)
-- | -- | -- | --
Rheem, Shared | 40 | 5.0 | 3.6
Rheem, Shared | 50 | 4.5 | 3.6
Rheem, Shared | 65 | 5.6 | 4.0
Rheem, Shared | 80 | 5.7 | 4.7
Rheem, Dedicated | 40 | 2.9 | 3.6
Rheem,   Dedicated | 50 | 3.3 | 3.6
Rheem, 240 V | 50 | 3.2 | 3.6
Rheem, 240 V | 65 | 4.2 | 4.0
Rheem, 240 V | 80 | 4.7 | 4.7



</body>

</html>


So, looks like the values in HPWHSim are based on real data, and UA is a bit higher for 120V units. It seems like the shared units in particular DO have higher tank losses than their 240V counterparts. I have a hard time understanding why the this wouldn't trend with volume and the 50 gal units would have a lower UA than 40 gal units. I proposed a change using the HPWHSim values for shared and dedicated units, but assuming that UA for 50 gal tanks is the same as for 40 gal tanks.

HPXMLtoOpenStudio/resources/defaults.rb
jmaguire1: @shorowit: I spent a while today going back through how I did this the first time around, I think I'm largely set up to redo the same simulations to derive the COP/UEF correlations we need. Before I do that, we should make sure we're on the same page about what to use for UA, since that will have a small effect on the correlations I come up with. Once we have that sorted, it's not too hard to rerun all the simulations and compile the results into the correlations we need.

HPXMLtoOpenStudio/resources/waterheater.rb
shorowit: @jmaguire1 I'm good with your proposed changes. Feel free to proceed with developing the UEF/COP regressions.

jmaguire1:
Here are all the simulations performed to develop correlations between the nominal COP of the heat pump (at 125 F water temperature and air temperature/humidity/pressure corresponding to what you'd see during a UEF test) and the rated UEF:
- [main.zip](https://github.com/user-attachments/files/31976520/main.zip)
- [240v.zip](https://github.com/user-attachments/files/31976523/240v.zip)
- [120v_dedicated.zip](https://github.com/user-attachments/files/31976537/120v_dedicated.zip)
- [120v_shared.zip](https://github.com/user-attachments/files/31976539/120v_shared.zip)

(Unzip everything into the same top directory.)

There's simulation inputs and outputs for 240V, 120V shared circuit, and 120V dedicated circuit products, along with the associated helper files (mostly the draw profiles), a readme, and some spreadsheets used for data processing.

workflow/tests/base_results/results_simulations_energy.csv
```diff
@@ -521,9 +524,9 @@ house019.xml,128.278,128.278,50.034,50.034,78.244,0.0,0.0,0.0,0.0,0.0,0.0,2.316,
...
-house023.xml,142.494,142.494,62.006,62.006,0.0,80.489,0.0,0.0,0.0,0.0,0.0,2.224,0.0,0.0,5.711,0.867,20.068,0.0,0.0,9.214,0.0,0.692,0.045,0.0,0.0,0.0,4.027,0.0,0.0,0.489,0.409,3.04,1.945,0.0,3.151,10.124,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,80.489,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
+house023.xml,140.184,140.184,59.696,59.696,0.0,80.489,0.0,0.0,0.0,0.0,0.0,2.224,0.0,0.0,5.711,0.867,17.758,0.0,0.0,9.214,0.0,0.692,0.045,0.0,0.0,0.0,4.027,0.0,0.0,0.489,0.409,3.04,1.945,0.0,3.151,10.124,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,80.489,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```
shorowit: This house has a 200F (?!) water heater setpoint. With this PR, we now assume there is a mixing value, and so the target temperature for the dishwasher and clothes washer is 125F, resulting in ~10% lower hot water electricity use compared to before.

workflow/tests/util.rb
```diff
@@ -375,14 +375,15 @@ def _verify_outputs(rundir, hpxml_path, results, hpxml, unit_multiplier)
     if hpxml.buildings.any? { |hpxml_bldg| hpxml_bldg.water_heating_systems.count { |wh| wh.water_heater_type == HPXML::WaterHeaterTypeHeatPump && wh.location == HPXML::LocationOtherExterior } > 0 }
       next if message.include? 'Water heater tank set point temperature is greater than or equal to the cut-in temperature of the heat pump water heater.'
     end
+    # WHs w/ mixing valves
+    if hpxml.buildings.any? { |hpxml_bldg| hpxml_bldg.water_heating_systems.count { |wh| wh.has_mixing_valve } > 0 }
+      next if message.include? 'Target water temperature is greater than the hot water temperature'
+      next if message.include? 'Target water temperature should be less than or equal to the hot water temperature'
+    end
```
shorowit: I should double-check whether this still occurs.

workflow/tests/base_results/results_simulations_energy.csv
shorowit, line 528: This home has a standard electric WH with a 200F (!!) setpoint. Previously we modeled the dishwasher and clothes washer as using a certain amount of hot water at that temperature. Because it has a high setpoint, we now assume there's a mixing valve with a 125F mixing valve setpoint, and that becomes the target temperature for the dishwasher and clothes washer. So the total hot water electricity use for the home is ~10% lower.

jmaguire1:
@shorowit: sorry I didn't see you marked this ready for review last night, I did get the 120V backup controls working as intended and added a test file for that!

Here's the HPWH power vs. outdoor air temperature for the new test file, you can see that we start using the 900 W elements only when the outdoor air temp is < 37 F, which is exactly what is intended:
image


We probably want to also never use a HPWH operating mode schedule for a 120V product, because it's basically always in HP_only mode as we have that currently defined. I'm totally fine with just pushing that one to a future issue, it's a super niche scenario but we probably ought to throw an error if we have a 120V HPWH and there's an operating mode schedule defined.

shorowit:
Thanks @jmaguire1! I can address those two things.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -6816,8 +6902,17 @@ def self.get_branch_circuit_voltage_default_values(branch_circuit)
...
-            component.is_a?(HPXML::ClothesDryer) ||
+      elsif component.is_a?(HPXML::WaterHeatingSystem)
+        if component.fuel_type == HPXML::FuelTypeElectricity
+          if component.hpwh_voltage.nil? # Not HPWH
+            voltages << HPXML::ElectricPanelVoltage240
+          elsif component.hpwh_voltage == HPXML::HPWHVoltage240
+            voltages << HPXML::ElectricPanelVoltage240
+          else # 120V HPWH
+            voltages << HPXML::ElectricPanelVoltage120
+          end
```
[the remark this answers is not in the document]
shorowit: If we really are worried about someone using `HPWHVoltage` for a non-HP water heater, then we should prevent it in the schematron. But it doesn't feel necessary.

jmaguire1 (approved):
Nothing jumps out at me here, and the results have seemed reasonable from a few test runs. As far as I'm concerned this looks ready to merge!

PR #2207, opened 2026-04-30
Update default energy rates

### LINKED ISSUE #2194: Update default energy rates
I see that 2024 data is available in SEDS.

I also found that EIA has [an API](https://www.eia.gov/opendata/) that we should consider using.

For example, [here are electricity prices from SEDS](https://www.eia.gov/opendata/browser/seds?frequency=annual&data=value;&facets=seriesId;&seriesId=ESRCD;&sortColumn=period;&sortDirection=desc;)

That said, I think one of the reasons we switched to SEDS is because the data is easier to work with. But via the API, maybe we should reconsider? If we used other data sources, we could get more recent data faster. For example:
- [Electricity prices (2025)](https://www.eia.gov/opendata/browser/electricity/retail-sales?frequency=annual&data=price;&facets=sectorid;&sectorid=RES;&sortColumn=period;&sortDirection=desc;)
- [Natural gas prices (2025 for most states)](https://www.eia.gov/opendata/browser/natural-gas/pri/sum?frequency=annual&data=value;&facets=process;&process=PRS;&sortColumn=period;&sortDirection=desc;)

bpark1327:
## Pull Request Description

Closes #2194 

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/utility_bills.rb
```diff
@@ -104,31 +104,52 @@ def self.marginal_rate_to_average_rate(marginal_rate, fixed_charge, household_co
...
-      begin
-        seds_rate = Float(seds_rate)
-      rescue ArgumentError, TypeError
-        seds_rate = 0.0
-        runner.registerWarning("No EIA SEDS rate for #{fuel_type} was found for the state of #{state_code}.") if not runner.nil?
-      end
+    csv_path = File.join(File.dirname(__FILE__), '../../ReportUtilityBills/resources/simple_rates/eia_fuel_rates_by_state.csv')
+
+    # Collect all matching rows for this state + fuel, keyed by year (desc)
+    matched_rows = []
```
shorowit: @bpark1327 You can simplify this code a lot now that we just have a single row for each state/fuel combination, right? No longer a `matched_rows` array, no longer need `latest`/`prior` below, etc.

bpark1327:
@shorowit I wanted to follow up on your comment about ensuring OS-HPXML skips utility bill calculations entirely (instead of proceeding with a zero coal rate). Looking at the test case below, it seems like that behavior might already be in place. Am I interpreting that correctly?
https://github.com/NatLabRockies/OpenStudio-HPXML/blob/0e00e6ed9cddc81abc256fb457829ddbe5dc89ec/ReportUtilityBills/tests/test_report_utility_bills.rb#L357

shorowit:
> @shorowit I wanted to follow up on your comment about ensuring OS-HPXML skips utility bill calculations entirely (instead of proceeding with a zero coal rate). Looking at the test case below, it seems like that behavior might already be in place. Am I interpreting that correctly?

Yes, I agree with you. Great!

shorowit (approved):
Looks great to me, thanks @bpark1327! Just made a few small cleanup changes. Merging.

PR #2229, opened 2026-06-23
Remove duplicated inferred ceiling/floor

yzhou601:
## Pull Request Description

We used to create two adiabatic surfaces: one for ceiling one for floor, for all the middle-story surfaces, but they are actually the same surface. It is probably a legacy from previous approach of modeling multiple conditioned thermal zones, but we no longer need two surfaces.
This was captured by our investigation of thermal comfort metrics (MRT and zone operative temperatures), the duplicated surfaces does have an impact on these surface temperature based calculations.
In this PR, I removed the adiabatic ceiling, it will impact simulation results. 

## Checklist

Not all may apply:


shorowit (approved):
Looks good. Pulled in master again, will merge when the CI completes.

PR #2231, opened 2026-06-29
Change HPWH `HeatingCapacity` to be input capacity

shorowit:
## Pull Request Description

For heat pump water heaters, ``HeatingCapacity`` is now *input* rather than *output* capacity, similar to other water heater types. This information is also more readily available on product listings ([example](https://files.rheem.com/blobazrheem/wp-content/uploads/sites/2/Plug-in-Heat-Pump-Spec-Sheet.pdf)).

## Checklist

Not all may apply:


HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -7102,7 +7102,7 @@ def self.get_service_feeder_power_default_values(runner, hpxml_bldg, service_fee
         if water_heating_system.water_heater_type == HPXML::WaterHeaterTypeStorage
           watts += UnitConversions.convert(water_heating_system.heating_capacity, 'btu/hr', 'w')
         elsif water_heating_system.water_heater_type == HPXML::WaterHeaterTypeHeatPump
-          watts += [UnitConversions.convert(Waterheater.get_heating_input_capacity(water_heating_system.heating_capacity, water_heating_system.additional_properties.cop), 'btu/hr', 'w'),
+          watts += [UnitConversions.convert(water_heating_system.heating_capacity, 'btu/hr', 'w'),
                     UnitConversions.convert(water_heating_system.backup_heating_capacity, 'btu/hr', 'w')].max
```
shorowit, line 7106: FYI @joseph-robertson, this changes the electric panel defaulting code for a HPWH (and removes the `Waterheater.get_heating_input_capacity` method). I assume that's not a problem but just wanted to point it out.

workflow/tests/base_results/results_simulations_bills.csv
```diff
@@ -125,12 +125,12 @@ base-dhw-tank-gas-ef.xml,1815.05,144.0,1102.19,0.0,1246.19,144.0,424.86,568.86,0
...
 base-dhw-tank-gas.xml,1809.49,144.0,1100.83,0.0,1244.83,144.0,420.66,564.66,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
-base-dhw-tank-heat-pump-capacities.xml,1807.85,144.0,1185.24,0.0,1329.24,144.0,334.61,478.61,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+base-dhw-tank-heat-pump-capacities.xml,1807.17,144.0,1188.62,0.0,1332.62,144.0,330.55,474.55,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```
shorowit, line 128: This is the only file that really changed. Now the 3000 Btu/hr (879 W) heating capacity in the HPXML file is (correctly) treated as an input capacity rather than an output capacity.

Some other files show very minor changes due to rounding.

HPXMLtoOpenStudio/resources/defaults.rb
```diff
@@ -3354,7 +3354,7 @@ def self.apply_water_heaters(hpxml_bldg, eri_version, schedules_file)
         water_heating_system.additional_properties.cop = get_water_heater_heat_pump_cop(water_heating_system)
 
         if water_heating_system.heating_capacity.nil?
-          water_heating_system.heating_capacity = (UnitConversions.convert(0.5, 'kW', 'Btu/hr') * water_heating_system.additional_properties.cop).round
+          water_heating_system.heating_capacity = UnitConversions.convert(0.5, 'kW', 'Btu/hr').round
```
shorowit, line 3357: No longer include the COP in the defaulting. Docs updated accordingly.

HPXMLtoOpenStudio/resources/waterheater.rb
```diff
@@ -973,7 +973,8 @@ def self.apply_hpwh_dxcoil(runner, model, water_heating_system, elevation, obj_n
     )
 
     # Assumptions and values
-    cap = UnitConversions.convert(water_heating_system.heating_capacity, 'Btu/hr', 'W') * unit_multiplier # kW
+    cop = water_heating_system.additional_properties.cop
+    cap = UnitConversions.convert(water_heating_system.heating_capacity, 'Btu/hr', 'W') * cop * unit_multiplier # W, output capacity
     shr = 0.88 # unitless
```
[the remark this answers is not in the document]
shorowit: The code is correct, the heating capacity should be based on the unadjusted COP.

jmaguire1 (commented):
I don't know when this got introduced, but absolutely a bug, and not an insubstantial one! This goes a long way to explaining why the backup elements were never triggering, the heating capacity of the heat pump was larger than the element capacity. 

The last thing I wanted to check was if 500W was a reasonable default power level while we had this open. Looking at the report we used for the tank losses ([here](https://github.com/NatLabRockies/OpenStudio-HPXML/blob/7c247a9fd4d3270ae59649e405574dea02e95e8e/HPXMLtoOpenStudio/resources/waterheater.rb#L1075) of a recent Rheem product, their power at the end of a heating cycle, which is as close to nominal conditions as we can get, is almost exactly 500W, so that seems fine as well

shorowit:
Just to be clear, I don't really think it was a bug. The documentation correctly said that the HPXML input was the HPWH output capacity and we were using it in our code like that.
 
The only bug was that our one HPXML sample file that exercises this optional input had incorrectly specified an input capacity. So unless you're running HPXML files where this value was likewise specified as an input capacity, you shouldn't see any change.

PR #2272, opened 2026-09-18
Limit stochastic appliance events to eight hours

joseph-robertson:
### Overview

Limits stochastic cooking range, dishwasher, clothes washer, and clothes dryer events to eight hours.

The empirical duration distributions contain extreme outliers, including a 92-interval cooking event representing 23 hours. These events can produce implausibly high indoor temperatures, particularly in small homes without cooling.

### Changes

Add a shared 480-minute appliance event limit.
Apply the limit before and after monthly duration multipliers.
Add a deterministic regression test for the 23-hour cooking sample.
Document the limit and affected end uses.

### Validation

Seed 85650 selects the 92-interval cooking sample. The regression test verifies it is limited from 23 hours to 8 hours.

Related to NatLabRockies/resstock#1572.

### Demonstration

I created a 322 sqft multifamily unit with no cooling. Modifying schedule_files/occupancy-stochastic.csv, I created two versions:
* long (full power) event Jun 30 - Jul 1 (blue = schedule, orange = operative temperature)
* same long event but limited to 8 hrs (red = schedule, black = operative temperature)

Natural ventilation brings operative temperature way down, so I intentionally placed the long event between natural ventilation events.

image

Resulting IDF files have design levels (W) of 1398.1 vs 1438.5 -- annual cooking range energy use is identical, but the limited duration version has slightly higher operative temperatures throughout the year.

## Checklist

Not all may apply:


joseph-robertson:
**Why not just scale down magnitudes during long events?**

Scaling only the magnitudes of long events would be a worse fix because it would leave the unrealistic long duration in place while silently reducing the appliance load. That changes the modeled energy profile rather than correcting the underlying outlier event length, whereas capping the event duration preserves the sampled power distribution and removes only the implausible tail.

jmaguire1:
@joseph-robertson: Should we also have a reasonable minimum event duration? We had seen that cause unrealistic peak events in an [earlier issue](https://github.com/NatLabRockies/OpenStudio-HPXML/issues/1024) and Lixi did some work with Yingli as part of panels to address some of that. 

@lixiliu: Do you think there's anything else at the OS-HPXML level we might want to do to limit minimum event duration? Granted, this only became particularly problematic with 1 minute timesteps, since the minimum event duration is always 1 timestep, but if we think we need a minimum as well as a maximum now might be the time to think about what we should do.

